PR #227, opened 2026-01-26
Fix: Improve CoreBox shrink behavior when no recommendation

TalexDreamSoul:
## Problem

CoreBox does not properly shrink to minimum height when:
1. Recommendation is disabled or times out
2. Input is cleared but window stays expanded  
3. No results are available but window doesn't collapse

This creates a poor user experience where the window remains unnecessarily large.

### Issue 1: Recommendation Timeout Not Triggering Collapse
**Location**: `useSearch.ts` - recommendation timeout handler

The timeout handler would call `resetSearchState()` but not explicitly trigger a collapse, leaving the window in an expanded state.

### Issue 2: resetSearchState Not Triggering Collapse
**Location**: `useSearch.ts` - `resetSearchState()` function

When resetting search state, the function would dispatch a layout refresh event but not explicitly trigger collapse when input is empty.

### Issue 3: Height Calculation Not Forcing Minimum
**Location**: `useResize.ts` - `sendLayoutUpdate()` function

The height calculation would use `calculateDesiredHeight()` even when there are no results and no pending operations, instead of forcing minimum height.

### Issue 4: Waiting State Not Preserving Collapsed State
**Location**: `core-box/index.ts` - `applyLayoutUpdate()` function

When waiting for data (loading/recommendation), the function would keep current size but not check if already collapsed, potentially expanding unnecessarily.

### Fix 1: Explicit Collapse on Recommendation Timeout
```typescript
recommendationTimeoutId = setTimeout(() => {
  if (recommendationPending.value && searchResults.value.length === 0) {
    recommendationPending.value = false
    loading.value = false
    resetSearchState()
    // ✅ Explicitly trigger collapse when recommendation times out
    transport.send(CoreBoxEvents.ui.expand, { mode: 'collapse' }).catch(() => {})
  }
  recommendationTimeoutId = null
}, RECOMMENDATION_TIMEOUT_MS)
```

### Fix 2: Trigger Collapse in resetSearchState
```typescript
const resetSearchState = (): void => {
  // ... existing reset logic
  
  // ✅ Trigger collapse if no input and no results
    transport.send(CoreBoxEvents.ui.expand, { mode: 'collapse' }).catch(() => {})
  }
}
```

### Fix 3: Force Minimum Height When Appropriate
```typescript
// ✅ When no results and not waiting for data, force minimum height
let height: number
  height = MIN_HEIGHT
} else {
  height = calculateDesiredHeight(resultCount)
}
```

### Fix 4: Preserve Collapsed State When Waiting
```typescript
if (resultCount === 0 && (loading || recommendationPending)) {
  // ✅ If already collapsed, keep it collapsed
  if (coreBoxManager.isCollapsed) {
    return
  }
  // Not collapsed yet, keep current size to avoid jitter
  return
}
```

### Modified Files
1. **useSearch.ts**
   - Added explicit collapse trigger in recommendation timeout handler
   - Added collapse trigger in `resetSearchState()` when input is empty

2. **useResize.ts**  
   - Force minimum height when no results and not waiting for data
   - Improved height calculation logic

3. **core-box/index.ts**
   - Preserve collapsed state when waiting for data
   - Added comment clarifying strict collapse condition

## Testing

The fix should be tested with:
1. **Recommendation disabled**: Clear input → window should shrink
2. **Recommendation timeout**: Wait 400ms with no results → window should shrink
3. **Clear input**: Type then clear → window should shrink immediately
4. **No results**: Search with no matches → window should shrink

## Impact

- ✅ Improves user experience with proper window sizing
- ✅ Reduces visual clutter when no content is displayed
- ✅ Makes CoreBox behavior more predictable and consistent
- ✅ No breaking changes to existing functionality

## Related Issues

This PR addresses the CoreBox shrink issue reported by users.

## Bug Fixes

- Improved empty result state handling to prevent layout jitter when results are loading or pending.
- Fixed height calculation for empty searches to use consistent minimum sizing instead of content-based sizing.
- Enhanced collapse behavior during search reset and recommendation timeout scenarios.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/apps/core-app/src/main/modules/box-tool/core-box/index.ts
+++ b/apps/core-app/src/main/modules/box-tool/core-box/index.ts
@@ -190,6 +190,7 @@ export class CoreBoxModule extends BaseModule {
     const loading = payload.loading === true
     const recommendationPending = payload.recommendationPending === true
 
+    // Strict collapse condition: no results, not loading, not waiting for recommendation
     const shouldCollapse = resultCount === 0 && !loading && !recommendationPending
     if (shouldCollapse) {
       coreBoxManager.shrink()
@@ -198,7 +199,13 @@ export class CoreBoxModule extends BaseModule {
 
     // When waiting for data (loading/recommendation), keep current window size stable.
     // This avoids collapse-then-expand jitter while results are still pending.
+    // However, if already collapsed, stay collapsed.
     if (resultCount === 0 && (loading || recommendationPending)) {
+      if (coreBoxManager.isCollapsed) {
+        // Already collapsed, keep it collapsed
+        return
+      }
+      // Not collapsed yet, keep current size to avoid jitter
       return
     }
 
--- a/apps/core-app/src/renderer/src/modules/box/adapter/hooks/useResize.ts
+++ b/apps/core-app/src/renderer/src/modules/box/adapter/hooks/useResize.ts
@@ -73,7 +73,13 @@ export function useResize(options: UseResizeOptions): void {
       ? (recommendationPending?.value ?? false)
       : false
 
-    const height = calculateDesiredHeight(resultCount)
+    // When no results and not waiting for data, force minimum height
+    let height: number
+    if (resultCount === 0 && !isLoading && !isRecommendationPending) {
+      height = MIN_HEIGHT
+    } else {
+      height = calculateDesiredHeight(resultCount)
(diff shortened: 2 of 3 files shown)


TalexDreamSoul:
This problem has been fixed.

Issue #330, opened 2026-07-28
Test: repair the packages/test integration baseline and fixtures


TalexDreamSoul:
## Priority
P1 — repair the cross-package integration/compatibility suite after the CoreApp and Nexus release blockers are triaged.

## Problem
The workspace integration package currently reports 16 passing / 10 failing files and 301 passing / 26 failing tests. Failures include:

- Electron import/mock setup (`electron` missing `app`, `app.getLocale` undefined);
- missing historical helper and OCR fixture paths;
- stale plugin permission reason expectations;
- official plugin SDK marker/manifest surface drift;
- missing generated/bundled `touch-translation` files;
- window/system action platform-reason drift.

Several package-local plugin suites pass while this integration suite fails, indicating its fixtures and generated-runtime preconditions are no longer aligned with the canonical source/package pipeline.

## Proposed solution
- Define whether each test consumes canonical source, generated bundle, or packaged runtime; make that precondition explicit.
- Generate required plugin bundles in test setup or test the canonical source only—never depend on undeclared workspace residue.
- Replace deleted OCR image paths with a committed synthetic fixture that contains no user data.
- Centralize the Electron mock and permission-result factories so integration expectations follow shared contracts.
- Remove imports of deleted helper modules or restore an owned replacement.

## Acceptance criteria
- [ ] `corepack pnpm -C packages/test exec vitest run` passes from a clean checkout.
- [ ] No test depends on ignored/generated plugin output unless the command generates it first.
- [ ] Electron mocks provide the exact API surface consumed by the tested modules.
- [ ] Permission/platform assertions use canonical shared reason codes.
- [ ] Manifest-boundary expectations match the current supported SDK and remain fail-closed.
- [ ] OCR uses a repository-owned synthetic fixture and performs a real contract check when the addon is available.
- [ ] Two consecutive runs pass without relying on execution order.

## Verification
```bash
corepack pnpm -C packages/test exec vitest run
corepack pnpm plugins:validate
corepack pnpm test:targeted
```

## Non-goals
- Copying generated plugin artifacts into Git solely to make tests pass.

TalexDreamSoul:
## Measured baseline, and a warning about the green checkmark

`Integration suite (packages/test)` shows **SUCCESS** on every PR. It is not green — the job runs with `continue-on-error: true`, so it reports success whether the suite passes or not. Worth stating plainly, because it is easy to read that tick as this issue being resolved.

From the JUnit artifact rather than the log:

| | |
[...]

PR #1189, opened 2026-08-08
test(browser-bookmarks): drive the suite through the contracts the plugin actually has (#330)

TalexDreamSoul:
Refs #330. `packages/test` browser-bookmarks: **15 failing → 0 of 16**.
No test removed, none downgraded to `it.todo`.

> Rebuilt on current `master`. The branch previously carried unrelated history —
> my earlier PRs were squash-merged, so the originals still sat on this branch and
> the diff showed dozens of unrelated files. It's now two files.

---

Four separate pieces of drift, all pointing the same way: the suite was written
against a plugin that `e37c92c8c` (isolated-runtime migration) replaced.

### 1. `__test` is gone — deliberately

The migration removed the `__test` export, and the plugin's own suite **asserts its
absence** (`"exports lifecycle only"`, `plugins/touch-browser-bookmarks/index.test.cjs`).
Re-adding it to the shipped file would break that contract to serve a test.

Instead the internals are re-exported *at load time*, into this suite's copy of the
module only, through the source-transform loader that `intelligence.test.ts` already
uses for exactly this. The shipped plugin is untouched.

### 2. `FakeBuilder` had drifted from the plugin's own

- `createAndAddAction` **missing entirely** — only the older `addAction` existed, so
  every item build threw and the handler's generic failure item came back instead.
- `setMeta` overwrote rather than merged, while the plugin calls it twice per item.

### 3. `onItemAction` reads `item.actions[0]`, not `item.meta`

The tests passed `actionId` and `payload` inside `meta`, so `action` came back null,
the handler fell through every branch, and returned `undefined`. Eight call sites.

### 4. The permission model moved host-side

`e37c92c8c` deleted `ensurePermission` and **both** `permission.request` calls. The
plugin doesn't prompt any more — it calls the capability and reads the thrown error:
`PLUGIN_HOST_CAPABILITY_PERMISSION_DENIED` means denied, anything else means the call
failed. `permission.check` survives only to render diagnostics on the item.

The eight tests asserting the prompting contract now drive the capability the way the
host does. Intent unchanged — **denied / unavailable / failed / granted** — and two are
now specifically about *not* conflating a transport failure with a denial, which would
send a user to a permission screen that has nothing to fix.

### Assertion targets

`sourceType`, the payload and `capability` are written to the **action** by
`buildActionItem` (`index.js:448-465`); `item.meta` carries only `pluginName` /
`featureId` / `defaultAction`. The assertions read `item.meta`. `sourceKind` is
**dropped rather than relocated** — it doesn't exist anywhere in the plugin.

### The one plugin change

The `catch` in `onFeatureTriggered` logged the message and dropped the error. That
catch converts any fault into a generic `浏览器收藏加载失败` item, so the cause went
nowhere — which is why several unrelated faults in here all presented as the same
fallback item and stayed unexplained. It now carries the error.

### Verification

- `packages/test` browser-bookmarks: **16/16 passing** (was 1/16).
- The plugin's own `index.test.cjs`: still **5/5**, including the `__test`-absence assertion.
- ESLint clean on both changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error logging when browser bookmark actions fail, while preserving fallback behavior.
  * Enhanced handling of permission outcomes for opening URLs and copying to the clipboard.
  * Added support for retaining recently opened URLs.
  * Updated quicklink actions to use more reliable action payloads.

### DIFF
--- a/packages/test/src/plugins/browser-bookmarks.test.ts
+++ b/packages/test/src/plugins/browser-bookmarks.test.ts
@@ -1,44 +1,89 @@
 import { readFileSync } from 'node:fs'
 import { describe, expect, it, vi } from 'vitest'
-import { createPluginGlobals, loadPluginModule, withoutGlobal } from './plugin-loader'
+import { createPluginGlobals, loadPluginModule, loadPluginModuleWithSourceTransform, withoutGlobal } from './plugin-loader'
 
 const browserBookmarksUrl = new URL('../../../../plugins/touch-browser-bookmarks/index.js', import.meta.url)
-const browserBookmarksPlugin = loadPluginModule(browserBookmarksUrl)
-const { __test: browserBookmarksTest } = browserBookmarksPlugin
+// The plugin exports its lifecycle and nothing else. That is not an oversight to
+// route around: e37c92c8c removed the __test export in the isolated-runtime
+// migration, and the plugin's own suite asserts its absence ("exports lifecycle
+// only", plugins/touch-browser-bookmarks/index.test.cjs). Re-adding it to the
+// shipped file would break that contract to serve a test.
+//
+// So the internals are re-exported at load time instead, into this test's copy of
+// the module only -- the same approach intelligence.test.ts already takes.
+const TEST_EXPORT_NAMES = [
+  'normalizeUrlInput',
+  'upsertBookmark',
+  'cleanupRecent',
+  'mergeRecentForDisplay',
+] as const
 
+const browserBookmarksTest = loadPluginModuleWithSourceTransform<{
+  __test: Record<(typeof TEST_EXPORT_NAMES)[number], (...args: any[]) => any>
+}>(
+  browserBookmarksUrl,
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
**The red CI on this PR is not from this PR.** Every workflow whose first step is
`pnpm install` has been failing on Linux since 2026-08-05, on `master` too:

```
```

`ensure-node-pty-helper.cjs` asserts a path that only exists when node-pty downloads
a prebuild; on Linux it falls back to a source build, which writes
`build/Release/pty.node` and nothing under `prebuilds/`. The rebuild succeeds and the
guard throws anyway. Filed as #1194 with the log trail.

I checked this specifically rather than assuming — the failing `master` run
([30965683079](https://github.com/talex-touch/tuff/actions/runs/30965683079))
predates this branch and fails identically.

TalexDreamSoul:

Verified on that branch before closing: `packages/test/src/plugins/browser-bookmarks.test.ts` no longer destructures `__test`.

PR #1192, opened 2026-08-08
fix(test): make packages/test independent of the electron binary, and recover two dead suites (#330)

TalexDreamSoul:
Refs #330.

> **Corrected.** This PR originally claimed it took `notification-service` from 14
> failing to 0 by fixing a resolution mismatch that stopped `vi.mock` binding. That
> was wrong. Those 14 were failing because **my local machine had no electron
> binary** — the error said so and I talked myself out of believing it. Once electron
> installed, `notification-service` fails **2**, not 14, on plain `master` with no
> alias. CI installs electron normally, so those 12 never failed there.
>
> The rest of the PR stands and is re-verified against `origin/master` with electron
> present. Full correction on [#330](https://github.com/talex-touch/tuff/issues/330#issuecomment-5224981506).

### What this actually fixes

**1. Two suites failing with *zero* failed tests**

The summary line counts tests, so a file that dies during collection shows as a failed
*file* with no failed *test* — invisible next to the numbers beside it.

- **`common/index.test.ts`** imported `help/tree-generator`, pruned in `63c4f5022`
  (2026-07-14); this was its only remaining caller. The dead import took down the whole
  file, including the `str-num` test that had nothing to do with it. The file-tree test
  called `genFileTree(p)` and **asserted nothing**, so the suite is removed rather than
  rewritten — no coverage to reinstate.
- **`aisdk/provider-test.test.ts`** reached the real `@sentry/electron`. This one fails
  on CI too, independent of the binary: `@sentry/electron` cannot resolve `electron`
  under pnpm's strict layout, and with the binary present it fails differently —
  `The requested module 'electron' does not provide an export named 'app'`.

Both now run: **1 and 5 tests** that previously didn't exist in any report.

**2. `notification-service` API drift — 2 real failures**

`showUpdateDownloadCompleteNotification(version, taskId)` is now
`showUpdateReadyNotification({version, platform, onClick})` and **returns** whether it
notified. The tests asserted `.not.toThrow()`; they now assert the config gate through
that return value, which is what they were reaching for.

**3. Phantom dependency in the published `@talex-touch/utils`**

`packages/utils/plugin/node/logger.ts` imports `chalk`; `packages/utils` never declared
it. `package.json` lists `"plugin"` in `files` with no `exports` map, so
`@talex-touch/utils/plugin/node/logger` is a **public subpath of the published
package** — an external consumer gets that file with no `chalk` to resolve. In-repo it
was masked by hoisting.

Lockfile delta is additive; package key set unchanged at 5931.

### Why keep the aliases

Not to fix a mock that never bound — that framing was wrong. They make `packages/test`
**independent of a ~100MB binary download**, so the suite is hermetic and doesn't fail
differently depending on whether electron's postinstall ran.

| package | treatment | why |
|---|---|---|
| `electron` | aliased to a stub | Every export throws *by name*, so a test that forgot to mock a surface is told which one. |
| `@sentry/electron/main` | aliased | No-op, not throwing: fire-and-forget telemetry no test asserts. |
| `talex-mica-electron` | aliased | Inlining isn't enough — it's CJS, and vite's interop routes its `require()` through Node, which doesn't consult the alias map. |
| `@electron-toolkit/utils` | inlined | ESM, so the alias applies once it stops being external. |

The electron stub mirrors every name `apps/core-app/src/main` imports from `'electron'`.
That list is load-bearing: a namespace has no fallback for a missing export, so an
unlisted name fails the whole importing suite at collection.

### Verification

Baseline re-measured on `origin/master`, clean tree, **electron installed**: 61 failing
across 12 files. The three files this PR targets pass: 20/20. ESLint clean.

### DIFF
--- a/packages/test/src/aisdk/provider-test.test.ts
+++ b/packages/test/src/aisdk/provider-test.test.ts
@@ -1,7 +1,62 @@
 import type { IntelligenceProviderConfig } from '@talex-touch/tuff-intelligence'
 import { IntelligenceProviderType } from '@talex-touch/tuff-intelligence'
-import { describe, expect, it } from 'vitest'
+import { describe, expect, it, vi } from 'vitest'
 import { setIntelligenceProviderManager, tuffIntelligence } from '../../../../apps/core-app/src/main/modules/ai/intelligence-sdk'
+
+// This file's subject is the intelligence SDK, but importing it pulls in the
+// main-process bootstrap, which touches a wide slice of electron at module scope.
+// The stub that `electron` is aliased to throws by name on any unmocked surface,
+// which is the right default for a suite that means to assert electron behaviour
+// and the wrong one here: the boot path would have to be enumerated call by call,
+// and every future addition to it would break this file for no reason related to
+// what it tests.
+//
+// So the app-boot surfaces get a permissive auto-stub -- unknown members are
+// no-op functions -- while the handful of values the bootstrap actually reads are
+// given real answers. importOriginal is spread underneath so the surfaces nothing
+// here touches stay throwers rather than becoming undefined, which vitest rejects
+// outright when a factory omits an export.
+function autoStub(overrides: Record<string, unknown> = {}): any {
+  return new Proxy(overrides, {
+    get: (target, key) => key in target ? (target as any)[key] : () => undefined,
+  })
+}
+
+vi.mock('electron', async importOriginal => ({
+  ...(await importOriginal<typeof import('electron')>()),
+  app: autoStub({
+    isPackaged: false,
+    getPath: (name: string) => `/tmp/tuff-test/${name}`,
+    getLocale: () => 'en-US',
+    getVersion: () => '0.0.0-test',
+    getName: () => 'tuff-test',
+    requestSingleInstanceLock: () => true,
+    whenReady: async () => {},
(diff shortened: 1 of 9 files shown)


TalexDreamSoul:
**The red CI on this PR is not from this PR.** Every workflow whose first step is
`pnpm install` has been failing on Linux since 2026-08-05, on `master` too:

```
```

`ensure-node-pty-helper.cjs` asserts a path that only exists when node-pty downloads
a prebuild; on Linux it falls back to a source build, which writes
`build/Release/pty.node` and nothing under `prebuilds/`. The rebuild succeeds and the
guard throws anyway. Filed as #1194 with the log trail.

I checked this specifically rather than assuming — the failing `master` run
([30965683079](https://github.com/talex-touch/tuff/actions/runs/30965683079))
predates this branch and fails identically.

TalexDreamSoul:
Corrected the description: the 14 `notification-service` failures were my local machine missing the electron binary, not a mock-binding problem. Real count there is **2**. The two dead suites, the API drift and the `chalk` phantom dependency all still stand and are re-verified with electron present. Detail on #330.

TalexDreamSoul:
Superseded by #1224, which targets `TalexDreamSoul/app-shell-v2` — the branch this repo actually merges into. This one was opened against `master` by mistake, where it could never land. The replacement is verified on that branch.

Issue #1194, opened 2026-08-08
CI: every pnpm install on Linux fails — ensure-node-pty-helper checks a path only prebuilds create


TalexDreamSoul:
Every workflow that runs `pnpm install` on Linux has failed since **2026-08-05**.
It is not any one PR — it reproduces on `master`, and it currently shows up as red
CI on unrelated PRs.

```
apps/core-app postinstall: Error: node-pty spawn-helper is missing for linux-x64
 ELIFECYCLE  Command failed with exit code 1.
```

### Cause

`64bc4b514` (*feat(core-app): add local AI CLI quick invoke*) added
`apps/core-app/scripts/ensure-node-pty-helper.cjs` and wired it into `postinstall`.
It asserts a file exists at:

```
<node-pty>/prebuilds/<platform>-<arch>/spawn-helper
```

That path is populated when node-pty **downloads a prebuild**. On Linux CI there is
no prebuild, so node-pty falls back to a source build — and a source build writes
somewhere else entirely. From the install log of a failing `master` run
([30965683079](https://github.com/talex-touch/tuff/actions/runs/30965683079)):

```
> Rebuilding because directory .../node-pty/prebuilds/linux-x64 does not exist
  CXX(target) Release/obj.target/pty/src/unix/pty.o
  SOLINK_MODULE(target) Release/obj.target/pty.node
  COPY Release/pty.node
gyp info ok
```

The rebuild **succeeds**. It produces `build/Release/pty.node` and nothing else —
no `spawn-helper`, and nothing under `prebuilds/`. So the guard fires on a working
install, and because it `throw`s from `postinstall`, it takes the whole install
down with it.

Note the log line also shows node-pty itself reporting `prebuilds/linux-x64` absent
*before* rebuilding: the guard is checking the very path node-pty had just found
missing and worked around.

### Two things worth separating

1. **The path is wrong for source builds.** It only ever exists on the prebuild path.
2. **`spawn-helper` was not built at all here.** The gyp output above compiles only
   `pty.o` / `pty.node`. Whether Linux needs a `spawn-helper` is worth confirming
   against node-pty's `binding.gyp` before deciding what the guard should assert —
   if it is macOS-only, then the `process.platform !== 'win32'` condition is too
   broad and Linux should be excluded rather than have the check relocated.

### Blast radius

Confirmed failing on `master` and on PR branches: `CI - tuffex`, `CI - utils`,
[...]

PR #1195, opened 2026-08-08
test(snippets): rebuild the suite on the plugin's surviving surface (#330)

TalexDreamSoul:
Refs #330. `packages/test` snippets: **9 failing → 0**. Seven tests, from nine.

All nine failed for three different reasons, all introduced by `e37c92c8c`
(isolated-runtime migration).

### Deleted subjects — 3 tests

| gone | where the behaviour lives now |
|---|---|
| `applyPlaceholders` | `resolveClipboardPlaceholders`, reached through the **copy action**. Only `{{clipboard}}` survives. |
| `matchSnippet` | inlined into `showSearch` as a filter over `[title, content, ...tags]`, reached through the **search feature**. Both old assertions still hold under it. |

Rather than re-export deleted functions, these are driven through the surface the
plugin does expose — closer to what a user hits, and not dependent on internals at all.

### Deleted feature — 3 tests removed

`e37c92c8c` deleted the `pack-export` action outright. `onItemAction` has no branch
for it and returns `undefined`; there is nothing left to re-target those tests at.
The cloud-share flow that took over that ground is covered by the plugin's own suite
(`plugins/touch-snippets/index.test.cjs`), denial path included.

### Replaced model — 3 tests

The plugin no longer prompts. The host enforces and throws, and `actionFailure`
(`index.js:133-147`) maps `PLUGIN_HOST_CAPABILITY_PERMISSION_DENIED` →
`permission-denied` and `PLUGIN_HOST_CAPABILITY_UNAVAILABLE` → `host-unavailable`.
The three copy tests now drive that. One of them specifically separates a **transport
failure** from a **denial** — conflating them sends a user to a permission screen
that has nothing to fix.

### Same fixture drift as browser-bookmarks

`FakeBuilder` had no `createAndAddAction`, and `setMeta` overwrote instead of merging.
`buildItem` puts the payload in the **action** and `onItemAction` reads it back from
`item.actions`, so without it the payload never survived the round trip the test was
trying to exercise. The call sites also passed the payload in `meta`, where nothing
reads it.

### One thing reported rather than asserted

`applyPlaceholders` also substituted `{{date}}`, `{{time}}` and `{{uuid}}` — only
`{{clipboard}}` survived. I did **not** pin their absence. The manifest documents
`{{clipboard}}` alone and always did (checked at `e37c92c8c^`), so the three were
undocumented; the suite asserts what the plugin promises rather than freezing a loss
someone may want to undo. Flagged on #330 for a maintainer call.

### Verification

- `packages/test` snippets: **7/7 passing**.
- The plugin's own `index.test.cjs`: still **7/7**.
- ESLint clean.

> CI red on this branch is #1194 (`node-pty` postinstall), which fails on `master` too.

### DIFF
--- a/packages/test/src/plugins/snippets.test.ts
+++ b/packages/test/src/plugins/snippets.test.ts
@@ -1,18 +1,22 @@
 import { describe, expect, it, vi } from 'vitest'
-import { createPluginGlobals, loadPluginModule, withoutGlobal } from './plugin-loader'
+import { createPluginGlobals, loadPluginModule } from './plugin-loader'
 
 const snippetsUrl = new URL('../../../../plugins/touch-snippets/index.js', import.meta.url)
-const snippetsPlugin = loadPluginModule(snippetsUrl)
-const { __test: snippetsTest } = snippetsPlugin
 
+// Mirrors the builder in plugins/touch-snippets/index.test.cjs. This copy had
+// drifted: createAndAddAction was missing entirely, and setMeta overwrote instead
+// of merging. buildItem (index.js:116-131) puts the payload in the ACTION, and
+// onItemAction reads it back from item.actions -- so without the action the
+// payload never survives the round trip a test is trying to exercise.
 class FakeBuilder {
   item: Record<string, unknown>
 
   constructor(id: string) {
-    this.item = { id }
+    this.item = { id, meta: {}, actions: [] }
   }
 
-  setSource() {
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:

Verified on that branch before closing: `snippets.test.ts` there carries a comment naming the same `e37c92c8c` cause I derived independently.

PR #1197, opened 2026-08-08
fix(core-app): stop the node-pty postinstall check from failing every install (#1194)

### LINKED ISSUE #1194: CI: every pnpm install on Linux fails — ensure-node-pty-helper checks a path only prebuilds create

TalexDreamSoul:
Fixes #1194. **Unblocks CI repo-wide** — not just my branches.

Since 2026-08-05 every workflow whose first step is `pnpm install` has failed on
Linux, on `master` as well as on PRs:

```
 ELIFECYCLE  Command failed with exit code 1.
```

### What was wrong

The check asserted a file at `prebuilds/<platform>-<arch>/spawn-helper`. That path is
populated when node-pty **downloads a prebuild**. When it finds none it falls back to
`node-gyp rebuild`, which writes `build/Release/` and leaves **no `prebuilds/`
directory at all** — so the check fired on a build that had just succeeded.

The install log shows both halves plainly
([run 30965683079](https://github.com/talex-touch/tuff/actions/runs/30965683079)):

```
> Rebuilding because directory .../node-pty/prebuilds/linux-x64 does not exist
  CXX(target) Release/obj.target/pty/src/unix/pty.o
  SOLINK_MODULE(target) Release/obj.target/pty.node
  COPY Release/pty.node
gyp info ok
```

node-pty reports the prebuild directory missing, works around it, succeeds — and the
guard then checks the very path node-pty had just told it wasn't there.

### The fix — two changes, one per failure mode

| | |
|---|---|
| **layout** | Look in `build/Release/` as well as `prebuilds/`, so a source build is recognised rather than mistaken for a broken one. |
| **severity** | **Warn instead of throwing.** This runs from `postinstall`: throwing doesn't degrade the terminal, it stops the whole repo from installing. If a platform genuinely needs the helper and it's genuinely absent, node-pty says so when a PTY is spawned — a much better place to find out. |

The `chmod 0o755`, which is the actual point of the script, is unchanged and now
reaches the source-build layout it previously couldn't see.

Also guards `require.resolve('node-pty')` — a filtered or partial install may not have
node-pty at all, which was a second way for this to throw out of `postinstall`.

### Verified

Ran the script against all three layouts:

- `prebuilds/<platform>-<arch>/spawn-helper` present → chmod 644 → **755**, exit 0
- `build/Release/spawn-helper` present → chmod 644 → **755**, exit 0 *(the case that used to throw)*
- neither present → warns naming both paths, **exit 0**

### One thing I could not verify here

I claim in the commit message that no `spawn-helper` target runs on Linux. That is
read off the gyp output above — it compiles `pty.o` and `pty.node` and nothing else —
**not** from node-pty's `binding.gyp`, which I couldn't fetch (no registry access in
this environment). Worth a second pair of eyes if you want that stated as fact.

It does not affect the fix: the new behaviour is to warn rather than throw regardless
of platform, so the fix is correct either way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of native helper permissions during setup.
  * Supports both prebuilt and source-built helper locations.
  * Setup now continues gracefully when the helper is unavailable or permissions cannot be restored.
  * Windows environments are handled without unnecessary permission checks.
  * When found, the helper’s executable permissions are automatically restored.

### DIFF
--- a/apps/core-app/scripts/ensure-node-pty-helper.cjs
+++ b/apps/core-app/scripts/ensure-node-pty-helper.cjs
@@ -1,17 +1,60 @@
+/**
+ * node-pty ships `spawn-helper` as a separate executable and can lose its
+ * executable bit through some install paths, which then fails at runtime when a
+ * PTY is spawned rather than at install time. This restores it.
+ *
+ * It must not be able to fail an install. This runs from core-app's postinstall,
+ * so throwing here does not degrade the terminal -- it stops the repo from
+ * installing at all, on every workflow whose first step is `pnpm install`.
+ */
+
 const fs = require('node:fs')
 const path = require('node:path')
 
-if (process.platform !== 'win32') {
-  const entry = require.resolve('node-pty')
-  const helper = path.resolve(
-    path.dirname(entry),
-    '..',
-    'prebuilds',
-    `${process.platform}-${process.arch}`,
-    'spawn-helper'
-  )
-  if (!fs.existsSync(helper)) {
-    throw new Error(`node-pty spawn-helper is missing for ${process.platform}-${process.arch}`)
+function ensureHelperExecutable() {
+  if (process.platform === 'win32') {
+    return
   }
-  fs.chmodSync(helper, 0o755)
+
+  let packageRoot
+  try {
+    packageRoot = path.dirname(path.dirname(require.resolve('node-pty')))
+  }
+  catch {
+    // A filtered or partial install may not have node-pty at all. Nothing to fix.
+    return
+  }
+
+  // Two layouts, depending on how node-pty installed itself. `prebuilds/` is
+  // populated when it downloads a prebuild; when it finds none it falls back to
+  // `node-gyp rebuild`, which writes to `build/Release/` and leaves no `prebuilds/`
+  // directory behind. The original check knew only the first, so a perfectly good
+  // source build looked like a broken install.
+  const candidates = [
+    path.join(packageRoot, 'prebuilds', `${process.platform}-${process.arch}`, 'spawn-helper'),
+    path.join(packageRoot, 'build', 'Release', 'spawn-helper')
+  ]
+
+  const helper = candidates.find(candidate => fs.existsSync(candidate))
+
+  if (!helper) {
+    // Not necessarily a fault: on Linux the gyp build produces pty.node alone and no
+    // spawn-helper target runs at all. Say so and carry on. If a platform does need
+    // it and it is genuinely missing, node-pty reports that when a PTY is spawned,
+    // which is a better place to find out than a failed install of the whole repo.
+    console.warn(
+      `[ensure-node-pty-helper] no spawn-helper found for ${process.platform}-${process.arch}; `
+      + `skipping chmod. Looked in:\n  ${candidates.join('\n  ')}`
+    )
+    return
+  }
+
+  try {
+    fs.chmodSync(helper, 0o755)
+  }
+  catch (error) {
+    console.warn(`[ensure-node-pty-helper] could not chmod ${helper}:`, error)
+  }
 }
+
+ensureHelperExecutable()


TalexDreamSoul:
**Confirmed on Linux CI, not just locally.** From this PR's install step:

```
apps/core-app postinstall$ node scripts/ensure-node-pty-helper.cjs && echo 'Skipping ...'
apps/core-app postinstall: [ensure-node-pty-helper] no spawn-helper found for linux-x64; skipping chmod. Looked in:
apps/core-app postinstall:   .../node-pty/prebuilds/linux-x64/spawn-helper
apps/core-app postinstall:   .../node-pty/build/Release/spawn-helper
apps/core-app postinstall: Skipping electron-builder install-app-deps - will run during build
apps/core-app postinstall: Done
```

`Done` instead of `ELIFECYCLE`. `PR Quality` went FAILURE → SUCCESS on the same diff.

That output also settles the open question from the description: on `linux-x64`
**neither** layout exists — not the prebuild one and not the gyp output one. So
`spawn-helper` really isn't produced there, and warning rather than throwing is the
right severity, not just the safe one.

The remaining `Documentation Quality` failure is unrelated to this PR — it reports
`DOC-LINK-UNTRACKED=4, DOC-TASK-DUPLICATE-ID=9, DOC-TASK-META=39` (trellis task
metadata) and fails identically on #1189 and #1195, neither of which touches docs.

TalexDreamSoul:
Closing — **this problem does not exist on the branch that matters.**

`apps/core-app/scripts/ensure-node-pty-helper.cjs` is not present on
`TalexDreamSoul/app-shell-v2`, and that branch's `postinstall` is just
`echo 'Skipping electron-builder install-app-deps - will run during build'` — no node-pty
guard at all.

The guard arrives with `64bc4b514` (*feat(core-app): add local AI CLI quick invoke*), which
is one of only **four** commits `master` has that `app-shell-v2` doesn't. So #1194 is a
`master`-only artefact, and since every merged PR in this repo targets `app-shell-v2`, it
was never blocking the work I thought it was blocking.

The fix itself still stands if `master` is ever merged forward — the check asserted a path
only prebuilds create, and threw from `postinstall`, which stops the repo installing rather
than degrading the terminal. Reopen this if that merge is planned.

PR #1199, opened 2026-08-08
test(dev-toolbox): drive the suite through the current action contract (#330)

TalexDreamSoul:
Refs #330. `packages/test` dev-toolbox: **11 failing → 0 of 11**.

Eleven of twelve failed, in the two shapes this issue keeps producing.

### `__test` is gone — 3 tests

`parseToolboxConfig` and `normalizeExternalUrl` were reached through a `__test` export
that `e37c92c8c` removed. **Both functions still exist** — they're just not exported.
Re-exported at load time into this suite's copy of the module rather than adding a
test-only export back to the shipped plugin, using the source-transform loader
`intelligence.test.ts` already uses.

### The action shape moved — 8 tests

They passed `actionId` and `payload` inside `meta`. `onItemAction` resolves both from
`item.actions[0]` (`index.js:197-199`), so `action` came back null, every branch was
skipped, and the handler returned `undefined`.

That's why all eight reported the same `expected undefined to match object` — one
symptom, and it made eight distinct behaviours look like a single fault.

### The permission SDK isn't used any more

Those eight drove `permission.check` / `permission.request`. **This plugin doesn't
touch the permission SDK at all** — the only trace left is reading
`PLUGIN_HOST_CAPABILITY_PERMISSION_DENIED` off a thrown error (`index.js:233`). It calls
`openUrl` and interprets what comes back.

They now drive that, keeping the same intent: **invalid / denied / failed /
unavailable / granted**.

### Two tests are new, not remapped

Covering ground the old set implied but never asserted:

- a **transport failure** is reported as `open-url-failed`, not `permission-denied` —
  conflating them sends a user to a permission screen with nothing to fix
- an item whose `defaultAction` belongs to another plugin is **ignored**, not acted on

### Verification

- `packages/test` dev-toolbox: **11/11 passing**.
- The plugin's own `index.test.cjs`: still **3/3**.
- ESLint clean.

> CI red on this branch is #1194 (`node-pty` postinstall), fixed in #1197.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for developer toolbox link-opening behavior, including invalid links, denied access, host failures, unavailable browser support, and successfully normalized links.
  * Added validation to ensure unrelated actions are ignored.
  * Updated test setup to reflect the current development-tool behavior.

### DIFF
--- a/packages/test/src/plugins/dev-toolbox.test.ts
+++ b/packages/test/src/plugins/dev-toolbox.test.ts
@@ -1,10 +1,23 @@
 import { readFileSync } from 'node:fs'
 import { describe, expect, it, vi } from 'vitest'
-import { createPluginGlobals, loadPluginModule, withoutGlobal } from './plugin-loader'
+import { createPluginGlobals, loadPluginModule, loadPluginModuleWithSourceTransform, withoutGlobal } from './plugin-loader'
 
 const devToolboxUrl = new URL('../../../../plugins/touch-dev-toolbox/index.js', import.meta.url)
-const toolboxPlugin = loadPluginModule(devToolboxUrl)
-const { __test: toolboxTest } = toolboxPlugin
+
+// The plugin exports its lifecycle and nothing else: e37c92c8c removed the __test
+// export in the isolated-runtime migration. Both functions below still exist, they
+// are simply not exported, so they are re-exported at load time into this test's
+// copy of the module rather than adding a test-only export to the shipped file --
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Superseded by #1226, which targets `TalexDreamSoul/app-shell-v2` — the branch this repo actually merges into. This one was opened against `master` by mistake, where it could never land. The replacement is verified on that branch.

PR #1202, opened 2026-08-08
test(plugins): realign batch-rename, dev-utils and text-tools with the current plugins (#330)

TalexDreamSoul:
Refs #330. **18 failures → 0** across three files.

| file | before | after |
|---|---|---|
| `dev-utils` | 7 failing | **0** (11 tests, from 9) |
| `text-tools` | 4 failing | **0** (5 tests, from 4) |
| `batch-rename` | 7 failing | **0** (8 tests, from 9) |

---

### `text-tools` — four failures, one cause

All four failed for a single reason that presented as four different assertion errors.

The MD5 and SHA-1 items call `globalThis.crypto.digest(algorithm, bytes)` — a **host
capability**, not Web Crypto (whose `digest` lives on `crypto.subtle` and has no MD5
anyway). Nothing provided it, so `digestHex` threw `CRYPTO_DIGEST_UNAVAILABLE`; the hash
items aren't individually guarded, so the throw escaped `onFeatureTriggered` and every
test received the plugin's generic `加载失败` item instead of any real output.

**Worth knowing for other tests in this suite:** it's installed in `beforeAll`, not
through `createPluginGlobals`. The loader restores overridden globals as soon as the
module finishes *compiling*. Globals a plugin destructures at module scope survive
that; `globalThis.crypto` is read inside `digestHex` **at call time**, long after the
restore. Passing it through the loader silently does nothing.

### `dev-utils` — 10 helpers + the permission model

Ten helpers reached through the `__test` export `e37c92c8c` removed. All ten still
exist and are just unexported, so they're re-exported at load time into this suite's
copy of the module. Two more tests drove `permission.request`; the plugin now calls
`clipboard.writeText` and reads the thrown error.

### `batch-rename` — four separate pieces of drift

- `parseRules` / `buildRenamePlan` via the same source transform
- the plan field is `targetName`; `nextName` is gone
- items were located by `item.meta.actionId`, but `buildItem` puts the action id **on
  the action** — every lookup missed
- `FakeBuilder` had no `createAndAddAction`, so the items had no actions to find either
- `hasPermission` calls `permission.check` **only** — no `request` path survives, so
  assertions that one was called with a reason string described a removed gate

### Two behaviours now pinned rather than left implicit

- **A transport failure is not a permission denial.** Three plugins here distinguish
  them and nothing was checking it. Conflating them sends a user to a permission screen
  that has nothing to fix.
- **`hasPermission` collapses three cases into one.** Denied, SDK-absent and
  check-threw all become `false`, so all three surface as `permission-denied`. Pinned as
  current behaviour and flagged on #330 — a user whose permission host is missing or
  broken is currently told they *lack a permission*, which is misleading. Your call
  whether that's worth separating.

### One test was passing for the wrong reason

`keeps apply inert when no preview plan exists` asserted `undefined` — which it got only
because the action id was passed where nothing reads it, so the handler bailed before
reaching any logic. Against the real shape the plugin returns `preview-missing` **without
touching permissions**, which is the behaviour actually worth having: no permission
prompt before there's anything to apply. Rewritten to assert that.

### Verification

Each file run individually: **11/11, 5/5, 8/8**. ESLint clean on all three.

`manifest-boundary` (4 failures) is deliberately **not** in this PR — it's a trust-boundary
tripwire that is firing on what look like real gaps, not stale fixtures, and it needs its
own treatment. Detail on #330.

### DIFF
--- a/packages/test/src/plugins/batch-rename.test.ts
+++ b/packages/test/src/plugins/batch-rename.test.ts
@@ -1,15 +1,26 @@
 import { describe, expect, it, vi } from 'vitest'
-import { createPluginGlobals, loadPluginModule, withoutGlobal } from './plugin-loader'
+import { createPluginGlobals, loadPluginModule, loadPluginModuleWithSourceTransform, withoutGlobal } from './plugin-loader'
 
 const batchRenameUrl = new URL('../../../../plugins/touch-batch-rename/index.js', import.meta.url)
-const renamePlugin = loadPluginModule(batchRenameUrl)
-const { __test: renameTest } = renamePlugin
+
+// e37c92c8c removed the __test export. Both helpers still exist in the plugin and are
+// simply not exported, so they are re-exported at load time into this suite's copy of
+// the module rather than adding a test-only export back to the shipped file.
+const TEST_EXPORT_NAMES = ['parseRules', 'buildRenamePlan'] as const
+
+const renameTest = loadPluginModuleWithSourceTransform<{
+  __test: Record<(typeof TEST_EXPORT_NAMES)[number], (...args: any[]) => any>
+}>(
+  batchRenameUrl,
+  source => `${source}\nmodule.exports.__test={${TEST_EXPORT_NAMES.join(',')}}`,
+  createPluginGlobals(),
+).__test
 
 class FakeBuilder {
   item: Record<string, unknown>
 
   constructor(id: string) {
-    this.item = { id }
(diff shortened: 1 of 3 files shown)


TalexDreamSoul:
Superseded by #1226, which targets `TalexDreamSoul/app-shell-v2` — the branch this repo actually merges into. This one was opened against `master` by mistake, where it could never land. The replacement is verified on that branch.

PR #1205, opened 2026-08-08
test(native): give the OCR contract test a fixture it owns (#330)

TalexDreamSoul:
Refs #330. `packages/test` native: **2 failing → 0 of 11**.

### The OCR test was pointing at a deleted file

It read `shots/LogoBanner.png`, deleted in `bd4a98bf1` (a docs media migration) and
never replaced — so it failed on its own **first** assertion,
`expect(existsSync(fixturePath)).toBe(true)`.

`shots/` is **not** gitignored, so this failed on CI too. Not a local-environment case.

Replaced with a fixture that lives next to the test: **3.6KB** of generated block
letters instead of a **7.8MB** marketing asset that unrelated housekeeping can remove.
Apple Vision reads it as exactly `TUFF`, which is what the assertion wants.

`scripts/make-ocr-fixture.cjs` regenerates it with no font or image dependency, so it
stays reproducible anywhere Node runs rather than being an opaque binary nobody can
reconstruct.

### The other native failure needed no change

`tuff-native-screenshot` fails only when `tuff_native_screenshot.node` hasn't been
built — it's a **separate cargo build**, not part of `binding.gyp`, which is why a
plain `node-gyp rebuild` leaves it missing. `native-protocol.yml` runs
`build:screenshot` before the native tests.

I built it locally and the suite passes, so I've left it alone: environment, not defect.
Checking rather than assuming, after getting exactly this call wrong earlier on this
issue with the electron binary.

### Verification

`packages/test/src/native`: **11/11 passing**, with the screenshot addon built.
ESLint clean on both changed files. Fixture regenerated from the script to confirm it
reproduces.


TalexDreamSoul:
Superseded by #1221, which targets `TalexDreamSoul/app-shell-v2` — the branch this repo actually merges into. This one was opened against `master` by mistake, where it could never land. The replacement is verified on that branch.

Issue #1125, opened 2026-08-07
native-protocol.yml is red on every branch, and #543 is about to make it visible on master


TalexDreamSoul:
## What

`native-protocol.yml` fails on every branch it runs on. This is the workflow that gates `packages/tuff-native/**` and `apps/core-app/src/main/modules/native-capabilities/**` — the Rust addons compiled into the shipped Electron app.

## Evidence

```
$ gh run list --workflow=native-protocol.yml --limit 8
failure  chore/325-app-suite-baseline    2026-08-07T16:27
failure  audit-fix/579-drop-hoist-2      2026-08-07T15:37
failure  audit-fix/579-drop-hoist-2      2026-08-07T15:20
failure  audit-fix/579-drop-hoist-2      2026-08-07T15:09
failure  audit-fix/579-drop-hoist-2      2026-08-07T15:00
failure  audit-fix/579-drop-hoist-2      2026-08-07T14:51
failure  audit-fix/579-drop-hoist-2      2026-08-07T10:55
```

Failing steps on run `31197648002`, all three matrix legs (macos-latest, windows-2022, ubuntu-latest):

- `Build production screenshot addon and verify exports`
- `Test Rust workspace`
- `Type-check Electron main process`

These are independent of whatever each branch changed — `audit-fix/579-drop-hoist-2` is a pnpm hoist change and `chore/325-app-suite-baseline` is a test baseline.

## Why it surfaced now

#543 fixed this workflow's `push` trigger, which had targeted a `main` branch that does not exist. That trigger had **never fired**, so the only signal came from pull requests — and only from PRs touching the filtered paths, which is rare. The gate was quiet because it was half-disconnected, not because it was passing.

With #543 merged (`ac80d34eb`), pushes to `master` touching the native paths will now report these failures. That is correct behaviour, but it means a red gate is about to become visible on the default branch.

## Second red gate, same shape

`Documentation Quality` → step `Verify documentation` also fails on unrelated PRs. It failed on #1122, whose entire diff is an added `cargo` entry in `.github/dependabot.yml` — a change that touches no documentation. Whatever it is asserting, it is not asserting it about the PR's contents.

## What is needed

1. Triage the three native-protocol failures — they may share one root cause (the screenshot addon build) or be three separate ones. The `Type-check Electron main process` leg is suspicious, since a CoreApp typecheck failing inside the *native* workflow suggests it is failing everywhere and simply not gated elsewhere.
[...]

PR #1207, opened 2026-08-08
fix(tuff-native): expect the frozen-compose feature the capability always adds (#1125)

TalexDreamSoul:
Fixes the **Windows leg** of #1125.

`verify:screenshot-production` failed on every branch:

```
AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected
  [
    'display',
    'region',
+   'frozen-compose'
  ]
```

`ScreenshotCapability::features()` appends `frozen-compose` to whatever the backend
reports — **unconditionally** (`native-screenshot/src/capability.rs:35-39`):

```rust
fn features(&self) -> Vec<String> {
    let mut features = self.backend.support().features().to_vec();
    if !features.iter().any(|feature| feature == "frozen-compose") {
        features.push("frozen-compose".to_string());
    }
    features
}
```

So the win32/linux branch was asserting a list that **cannot occur**. This isn't
platform drift — the assertion was wrong from the moment `frozen-compose` was added
in `f184d0966`.

The Rust side already agrees on the three-feature list
(`backend_contract_tests.rs:377`: `vec!["display", "region", "frozen-compose"]`).
The JS verifier was the only place that disagreed.

**macOS never caught it** because its branch asserts `engine` and `state` only, never
`features` — which is also why this reads as a Windows-specific bug when it isn't.

### Verified

Built the production screenshot addon locally and ran
`verify-screenshot-production.js` on darwin: passes. That branch is untouched; this
only corrects the win32/linux expectation.

I can't execute the Windows path here, so the Windows leg of `native-protocol.yml` on
this PR is the real check.

### DIFF
--- a/packages/tuff-native/scripts/verify-screenshot-production.js
+++ b/packages/tuff-native/scripts/verify-screenshot-production.js
@@ -35,7 +35,13 @@ try {
     else {
       assert.equal(capability.state, 'degraded')
       assert.equal(capability.reason, 'basic-backend-only')
-      assert.deepEqual(capability.features, ['display', 'region'])
+      // ScreenshotCapability::features() appends frozen-compose to whatever the
+      // backend reports, unconditionally (native-screenshot/src/capability.rs:35-39),
+      // so this list could never be just the backend's two. The Rust contract test
+      // already expects all three (backend_contract_tests.rs:377); this assertion was
+      // the outlier, and it made the Windows leg of native-protocol.yml red on every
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Superseded by #1222, which targets `TalexDreamSoul/app-shell-v2` — the branch this repo actually merges into. This one was opened against `master` by mistake, where it could never land. The replacement is verified on that branch.

Issue #323, opened 2026-07-28
Test: restore the full CoreApp Vitest suite


TalexDreamSoul:
## Priority
P0 — restore the CoreApp release regression baseline.

## Problem
`corepack pnpm -C apps/core-app test` on current `v2.4.13` source reports:

- 467 passing / 6 failing test files;
- 3,286 passing / 18 failing tests.

Observed failure groups:
- packaged runtime closure no longer includes the expected `formdata-node` entry;
- DeviceIdleService timeout;
- Agent session and workflow host-only boundary response drift;
- Intelligence invoke actor-boundary timeout;
- AppProvider Windows watcher/mutation-drain timeouts and missed event assertions.

The targeted PR gate passes, so these failures are not currently release-blocking even though they cover packaging, security boundaries, idle scheduling, and search/index shutdown.

## Required triage
For every failure, classify it as:
1. a product regression to fix;
2. an intentional contract change requiring an updated behavior test; or
3. a nondeterministic test requiring deterministic clocks/events rather than a larger timeout.

## Acceptance criteria
- [ ] `corepack pnpm -C apps/core-app test` passes from a clean checkout with no retries.
- [ ] Runtime-closure tests assert the modules the packaged loader actually resolves.
- [ ] Host-only/plugin actor tests assert canonical error codes and verify no protected service runs.
- [ ] DeviceIdle and AppProvider tests use deterministic timers/events and do not merely raise global timeouts.
- [ ] Windows watcher tests remain runnable on non-Windows CI through explicit adapters/mocks without weakening real Windows coverage.
- [ ] A second consecutive full run passes to detect flakes.
- [ ] No tests are deleted or skipped without a linked replacement contract.

## Verification
```bash
corepack pnpm -C apps/core-app test
corepack pnpm -C apps/core-app run typecheck
corepack pnpm -C apps/core-app run build:vite
```

## Non-goals
- Broad production refactoring unrelated to the failing contracts.
- Treating focused tests as proof that the full suite is restored.

TalexDreamSoul:
已在 **PR #1115** 里给出 #323 要求的分类，并订正了我先前在 #1070 上的一处错误测量（详见该 PR 与 #1070）。

## 分类结果

当前 `app-shell-v2` 干净安装下全量套件：

```
Test Files  23 failed | 603 passed | 3 skipped (629)
Tests       54 failed | 5124 passed | 8 skipped (5186)
```

[...]

PR #1208, opened 2026-08-08
test(quickops): restore the dropped contract coverage, and point the audit at it (#323)

TalexDreamSoul:
Refs #323. `quick-ops-flow-ai-adapter-audit` was the last failing CoreApp suite here.

All **16** implementation markers the audit looks for are present in
`plugins/touch-quickops/index.js`. All **6** test markers were gone — dropped when
`e37c92c8c` cut `index.test.cjs` from 1035 lines to 220.

### Two tests restore real coverage

- **High-risk blocking.** `kill port 3000` publishes the blocked notice and *nothing
  dispatchable* — no flow item, no actions, no payload — and acting on it anyway
  reaches no facade. A blocked action that still ships its payload is one careless
  dispatch away from running.
- **Host projection.** Every published item is projected down to a fixed meta shape
  before it leaves the plugin, and the request text does not survive it.

The second exists because of what the first turned up: `projectItemsForHost` strips
item meta to a **four-key whitelist**, so `mode`, `confirmation`, `result` and
`flowAdapterTrace` never reach the host at all. That's a real protection and nothing
was asserting it. My first attempt at these tests asserted the *pre*-projection shape
and failed — which is how I found it.

### `flowAdapterTrace` is dead data — reported, not deleted

It's built on three code paths, stripped before publication, and has **no consumer
anywhere in the repo** (`grep -rn flowAdapterTrace` outside the file that builds it:
nothing).

So two of the audit's markers were demanding test evidence of a structure nothing
reads. I've retargeted them at the projection test, which is where the protection
actually lives, and flagged the trace itself on #323 rather than quietly deleting it —
whether it should be published or removed is a maintainer call.

### One marker was never a coverage gap

`onItemAction dispatches safe QuickOps Flow action` is covered by a test with a
longer, more specific name (`…dispatches only the fixed Flow payload and awaits the
acknowledgement`). The test was never lost; only the marker went stale. Retargeted the
marker rather than renaming a good test to satisfy a grep.

### Verification

- `apps/core-app` quick-ops: **147 passing across 6 files** (was 1 failing).
- `plugins/touch-quickops/index.test.cjs`: **7/7**.
- The audit's synthetic fixture is updated to match, so it's still proven able to
  detect the signals it looks for — not just passing because it stopped looking.
- **No lint delta**: 112 pre-existing root-config errors in these files before and
  after (core-app's own config has opposite trailing-comma rules).

### DIFF
--- a/apps/core-app/src/main/modules/quick-ops/quick-ops-flow-ai-adapter-audit.test.ts
+++ b/apps/core-app/src/main/modules/quick-ops/quick-ops-flow-ai-adapter-audit.test.ts
@@ -88,9 +88,8 @@ describe('QuickOps Flow and AI adapter audit', () => {
         {
           path: 'plugins/touch-quickops/index.test.cjs',
           source: [
-            'test("buildFlowAdapterTrace redacts request and payload values", () => {})',
-            'expect(trace).toMatchObject({ payloadKeys: [], sensitivePayloadRedacted: true })',
-            'test("onItemAction dispatches safe QuickOps Flow action", () => {})',
+            'test("projectItemsForHost publishes only whitelisted meta, so request detail cannot leak", () => {})',
+            'test("onItemAction dispatches only the fixed Flow payload and awaits the acknowledgement", () => {})',
             'assert.equal(items[0].meta.payload, undefined)',
             'assert.equal(resolveHighRiskFlowAction("kill port 3000").targetId, "quickops.port-kill")'
           ].join('\n')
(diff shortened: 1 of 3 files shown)


TalexDreamSoul:
Superseded by #1227, which targets `TalexDreamSoul/app-shell-v2` — the branch this repo actually merges into. This one was opened against `master` by mistake, where it could never land. The replacement is verified on that branch.

Issue #1098, opened 2026-08-07
nuxt is pinned at 4.4.8 behind an unhead 2 -> 3 migration; 5 prod-audit allowlist entries expire 2026-11-09


TalexDreamSoul:
Split out of #328, whose override work cleared nine of thirteen vulnerable packages but could not touch these two.

## What is stuck

Two advisories in the production audit need Nuxt 4.5.x:

| package | severity | patched | current |
|---|---|---|---|
| `@nuxt/devtools` | **critical** | `>= 3.3.1` | 2.7.0 |
| `nuxt` | high | `>= 4.5.1` | 4.4.8 |

Neither can be bumped in isolation:

- **`nuxt@4.5.2` requires `unhead ^3.3.1`**, while the root `pnpm.overrides` pins `unhead` to **2.1.15**. With the bump applied, `nuxt prepare` fails: `Package subpath './stream/iife' is not defined by "exports" in .../unhead/package.json imported from .../nuxt/dist/index.mjs`.
- **`@nuxt/devtools@3.4.1` requires `@nuxt/kit ^4.5.1`**, i.e. the same Nuxt 4.5.x line. There is no 2.x release carrying the fix — the advisory's patched range starts at 3.3.1.

So this is an `unhead` 2 → 3 major migration reaching every Nuxt-adjacent package, not a version bump. Verified by attempting it and reverting.

## Two routes, and the cheaper one is probably not the migration

**Route A — do the migration.** Lift the `unhead` override to `^3.3.1`, bump the `nuxt` and `@nuxt/devtools` catalog entries, and work through whatever `@unhead/vue` and the Nuxt module ecosystem require. Real breaking-change surface across `apps/nexus`.

**Route B — remove the reason Nuxt is in the production closure at all.** `vue-sonner@2.0.9` declares `nuxt: ^4.5.2`, `@nuxt/kit` and `@nuxt/schema` as optional peer dependencies, and **`apps/core-app` lists `vue-sonner` as a production dependency**. That is how a Nuxt build/server stack ends up inside CoreApp's `pnpm audit --prod` closure — an Electron app that does not run Nuxt.

Replacing or dropping `vue-sonner` would remove `nuxt` and `@nuxt/devtools` from the production audit outright, without any migration. This is also #328's acceptance criterion 3 ("CoreApp package closure no longer contains Nuxt build/server dependencies solely because of `vue-sonner`, or the necessity is proven"), so the two land together.

[...]

Issue #749, opened 2026-08-06
[audit/compat] Catalog pins next-auth ~4.21.1 while apps/nexus installs ~4.24.14


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** medium  ·  **Category:** dependencies  ·  **Domain:** compat

**Location:** `pnpm-workspace.yaml`:54

## What

Two contradictory next-auth pins exist in pnpm-workspace.yaml (the main catalog and a generated conflicts_ catalog), neither is referenced by any package, and the real consumer hard-codes a third literal range.

## Evidence

```
pnpm-workspace.yaml:54 `next-auth: ~4.21.1`, pnpm-workspace.yaml:102-103 `conflicts_next-auth_t4_24_14: next-auth: ~4.24.14`, apps/nexus/package.json:88 `"next-auth": "~4.24.14"`. package.json:208 then widens the peer check with `"next-auth": "~4.21.1 || ~4.24.14"`. Grep for `"next-auth": "catalog:` returns 0 hits.
```

## Why it matters / failure scenario

The catalog claims 4.21.x is the supported version while the app runs 4.24.x. @sidebase/nuxt-auth 0.9.4 peers on next-auth 4.21.x; the allowedVersions escape hatch silences the mismatch instead of resolving it, so a genuine 4.21-vs-4.24 API break in the auth handler (apps/nexus/nuxt.config.ts:48 reaches into next-auth/core/index.js directly) surfaces only at runtime.

## Suggested direction

Delete the conflicts_ catalog, set the single catalog entry to ~4.24.14, and have apps/nexus use `catalog:`.

---
<sub>audit-domain: compat · confidence: confirmed</sub>

TalexDreamSoul:
## 状态已变:矛盾不存在了,剩下的是死配置

### 前提复核

issue 说 catalog 里 `~4.21.1` 与 `conflicts_` 里 `~4.24.14` 互相矛盾。**今天这三处已经一致了**:

| 位置 | 现值 |
|---|---|
| `pnpm-workspace.yaml:54`(默认 catalog) | `~4.24.15` |
| `pnpm-workspace.yaml:102-103`(`conflicts_` catalog) | `~4.24.15` |
| `apps/nexus/package.json:94` | `~4.24.15` |
| `package.json:211` `allowedVersions` | `~4.24.15` |

[...]

Issue #483, opened 2026-08-04
Security: triage 51 Dependabot alerts with reachability and release gates


TalexDreamSoul:
## Problem

The 2026-08-04 `master` push reported 51 Dependabot alerts: 3 critical, 20 high, 25 moderate, and 3 low. This is an alert-count baseline, not a demonstrated runtime-reachability conclusion. No open issue currently owns per-alert disposition.

## Required work

- Inventory every open alert in the GitHub Security dashboard without copying advisory-sensitive data into public logs.
- For each alert, record the affected dependency, reachable runtime/package surface, upgrade path, and either remediation owner or time-bounded exemption rationale.
- Apply upgrades through targeted regression gates; do not suppress an alert solely because it is transitive.
- Establish a release gate requiring new critical/high alerts to be triaged before publication.

## Evidence

- 2026-08-04 GitHub push receipt for `master`: 51 alerts (3 critical, 20 high, 25 moderate, 3 low).
- Maintenance audit: `docs/engineering/reports/maintenance-audit-2026-08-04.md`.

## Acceptance criteria

- [ ] Every alert has a documented disposition: upgraded, not reachable with evidence, accepted exception with expiry, or tracked remediation.
- [ ] Critical/high alert dispositions identify an owner and target date.
- [ ] Dependency upgrades have focused affected-package checks and release-impact verification.
- [ ] The audit report and roadmap reference the authoritative issue without treating alert count as exploitability.

TalexDreamSoul:
## 分诊底稿：51 条告警 = **19 个包**，且**每一条都有可用补丁**

从 Dependabot API 直接拉的当前 open 告警（`state=open`，分页取全），2026-08-07：

**51 条**，与 issue 的基线完全一致：critical 3 · high 20 · medium 25 · low 3。

按 scope 拆开（这是 issue 要的「reachable runtime/package surface」里可机器推导的那一半）：

| | runtime | development |
|---|--:|--:|
| critical | 3 | 0 |
| high | 14 | 6 |
| medium | 22 | 3 |
| low | 3 | 0 |

### 最要紧的一点：51 条其实只落在 19 个包上

| 包 | 告警数 | 最高等级 | scope | 首个已修版本 |
|---|--:|---|---|---|
| `@auth/core` | 6 | **critical** | runtime | 0.41.3 |
| `next-auth` | 3 | **critical** | runtime | 4.24.15 |
| `undici` | 8 | high | runtime | 6.28.0 / 7.29.0 |
[...]

Issue #325, opened 2026-07-28
CI: add sharded full-suite and production-audit release gates


TalexDreamSoul:
## Priority
P1 — make restored full suites enforceable without slowing every local edit loop.

## Problem
`.github/workflows/ci.yml` runs `pnpm quality:pr`, whose test stage is the 104-test targeted set. Current full CoreApp, Nexus, and integration suites fail while the PR gate remains green. A targeted gate is useful for fast feedback, but it cannot be the only release regression signal.

## Proposed CI policy
- Keep `quality:pr` as a fast mandatory PR gate.
- Add sharded full-suite jobs for CoreApp, Nexus, Utils/TuffEx, and `packages/test`.
- Run full suites on release tags and nightly/scheduled builds; make the relevant shard mandatory on changes in its owned paths.
- Upload machine-readable Vitest/JUnit summaries on failure.
- Add a flake policy: no silent retry; quarantines require owner, linked issue, expiry, and deterministic replacement plan.
- Add production dependency audit after lockfile install.

## Dependencies
This gate should become required only after the dedicated CoreApp, Nexus, and integration baseline issues are green.

## Acceptance criteria
- [ ] A deliberate failure in each suite fails its CI shard.
- [ ] Release/tag workflows cannot publish when a required full-suite shard fails.
- [ ] PR path filters run the affected full shard without skipping cross-package contract changes.
- [ ] Test reports identify failing file, test, duration, and artifact link.
- [ ] No global `continue-on-error`, blanket retry, or permanent skip hides a failure.
- [ ] Nightly/release workflow runs `pnpm audit --prod` under the approved exception policy.
- [ ] Workflow duration and cache behavior are documented and kept bounded through sharding.

## Verification
- Open a draft PR with controlled failing fixtures in each shard and confirm the expected gate fails.
- Revert the fixtures and confirm all required checks pass.

## Non-goals
- Replacing focused tests.
- Running every platform-specific packaged acceptance job on every documentation-only PR.

TalexDreamSoul:
Landed one acceptance criterion — *"Upload machine-readable Vitest/JUnit summaries on failure"* and *"Test reports identify failing file, test, duration"* — in #1114 (merged `25997ef5a`).

Not closing this: the issue's own dependency is unmet. *"This gate should become required only after the dedicated CoreApp, Nexus, and integration baseline issues are green"*, and they are not.

## Why the reporting had to come first

[...]

PR #1209, opened 2026-08-08
fix(nexus): take the patched next-auth, closing a critical account-hijack advisory (#325)

TalexDreamSoul:
Refs #325 (the `pnpm audit --prod` criterion), #483.

`pnpm audit --prod` reports `next-auth <4.24.15` as **critical**: Auth.js validates the
email address *before* Unicode normalization, so an attacker can register an address
that normalizes onto an existing account.

That's the same class as the unverified-email adoption bug fixed in #1185, and it sits
in the live auth path.

### It was already in range

`apps/nexus` asks for `~4.24.14`, which **permits** `4.24.15`. The lockfile had simply
never moved. This is a lockfile refresh with the spec floor raised to match — not a
version migration.

```
prod audit   critical 3 → 2      high 35 → 34      moderate 28 → 27
```

The extra high and moderate come along with it: `4.24.15` pulls `cookie ^0.7.0` and
`uuid ^11.1.1` in place of older transitive pins.

### On the peer conflict

`@sidebase/nuxt-auth@0.9.4` peers `next-auth: ~4.21.1`, and the repo already installs
4.24.x against it deliberately — hence the `allowedVersions` entry and the (unused)
`conflicts_next-auth_t4_24_14` catalog key. That conflict is pre-existing and tracked in
#749. **Moving inside 4.24.x neither creates nor worsens it.** I raised the
`allowedVersions` floor to match.

### Verified by A/B, not assumption

Installed and ran the full nexus suite both ways:

| | Test Files | Tests |
|---|---|---|
| HEAD (`4.24.14`) | 10 failed / 169 passed | **13 failed** / 943 passed |
| this branch (`4.24.15`) | 10 failed / 169 passed | **13 failed** / 943 passed |

Identical — no regression. Lockfile delta is a version move, not a graph change:
package key set unchanged at **5953**, and `pnpm install --lockfile-only
--frozen-lockfile` validates against every manifest.

### Separately: nexus is no longer green

Those 13 failures are **pre-existing on master** and unrelated to auth
(`pluginsStore-security-scan`, `intelligence/invoke`, others). That matters because
#1181 made this suite **blocking** when it was green at 963 tests / 0 failures.
Flagged on #325 — it isn't in scope for a dependency bump.

### The other two criticals are left alone

- **`@auth/core`** needs `>=0.41.3` against a declared `^0.34.3` — a breaking 0.x jump
  under `@sidebase/nuxt-auth`'s peer. Not a lockfile refresh.
- **`@nuxt/devtools <3.3.1`** (unauthenticated RPC → arbitrary command execution)
  reaches core-app through `vue-sonner > nuxt`, already tracked in #1098.

## Summary by CodeRabbit

* **Chores**
  * Updated the supported authentication library version to 4.24.15.
  * Retained compatibility with the existing 4.21.1 version range.


TalexDreamSoul:
Closing — opened against `master`, and `apps/nexus` on `TalexDreamSoul/app-shell-v2` does
not carry the `~4.24.14` pin this PR bumps.

**The security finding still needs checking on that branch.** `next-auth <4.24.15` is a
critical Auth.js advisory: the email normalizer validates before Unicode normalization, so
an attacker can register an address that normalizes onto an existing account — same class
as the unverified-email adoption bug fixed in #1185.

I have not verified which version `app-shell-v2` resolves. If it is below `4.24.15` the bump
is still wanted, and it was in-range on `master` (`~4.24.14` permits `4.24.15`), so it may
be a lockfile refresh rather than a migration there too. Cross-referenced on #483.

TalexDreamSoul:
Confirmed after closing: `app-shell-v2` already declares `"next-auth": "~4.24.15"` — the patched version. The critical advisory is **already addressed** on the branch that matters, so nothing is outstanding here. `@auth/core` is no longer declared in `apps/nexus/package.json` there either, so that second critical may also be moot; worth re-running `pnpm audit --prod` on `app-shell-v2` to get a true picture rather than the `master` one I reported on #483.

PR #1212, opened 2026-08-08
test(nexus): restore the blocking suite to green — 13 failures → 0 (#1210)

### LINKED ISSUE #1210: apps/nexus is red on master while #1181 makes it a blocking gate — 13 failures across 10 files
#1181 made the `apps/nexus` suite **blocking** on the strength of it being green:
*"180 files, 963 tests, zero failures — so the suite that was added non-blocking to make
it visible now holds the line."*

It no longer is. Measured on `origin/master`, clean install, this machine:

```
Test Files  10 failed | 169 passed (179)
Tests       13 failed | 943 passed (956)
```

### Failing files

| file | failed / total |
|---|---|
| `test/api/auth/auth-email-channel-contract.test.ts` | 1 / 2 |
| `app/pages/store-page-performance.test.ts` | 2 / 5 |
| `app/pages/dashboard/notifications.test.ts` | 1 / 3 |
| `test/api/app-auth/sign-in-token.post.test.ts` | 1 / 2 |
| `build/check-worker-bundle.test.ts` | 1 / 19 |
| `server/utils/__tests__/authStore.sync-items-merge.test.ts` | 1 / 1 |
| `server/utils/__tests__/plugin-security-scan.test.ts` | 1 / 5 |
| `server/utils/platformGovernanceStore.test.ts` | 2 / 22 |
| `server/utils/__tests__/pluginsStore-security-scan.test.ts` | 2 / 3 |
| `test/api/v1/intelligence/invoke.api.test.ts` | 1 / 3 |

Note the shape: four of the ten are **auth or security-scan** suites
(`auth-email-channel-contract`, `sign-in-token`, `authStore.sync-items-merge`,
`plugin-security-scan`, `pluginsStore-security-scan`). That is not a spread consistent
with flake, and it is not a good area to be running blind in.

Also note the totals moved: **963 → 956 tests**, so seven tests were removed or stopped
being collected between #1181 and now. Worth checking whether any of those were removed
rather than fixed.

### Why this needs attention rather than a mute

The gate is behaving correctly — it is red because the suite is red. The risk is the one
#1125 describes for other gates: a required check that is always red trains everyone to
ignore the checks column, and then it stops being a gate at all.

### How I found it

Running the nexus suite to A/B a `next-auth` dependency bump (#1209). Both sides were
identical at 13 / 943, which is what told me the failures were pre-existing rather than
mine. Not investigated further — I have not looked at the individual causes, only
established that they are on `master` and not from my branch.

I did not measure how long they have been failing. `git log` on the failing paths, or
the last green `App suites (nexus)` run, would place it.

TalexDreamSoul:
Fixes #1210. **`apps/nexus`: 13 failing → 0.** 179 files, 956 tests, all passing.

This suite is a **blocking gate** since #1181 and has been red on `master`. **None of
the 13 were product faults** — that's the headline, because a red required gate reads as
broken code and this wasn't.

---

### Four failed because time passed

| | |
|---|---|
| **governance analytics** (3) | Fixtures pinned to `2026-05-20..22`, queried with a 30-day window measured from `Date.now()`. They fell out of that window **around 2026-06-21** and every metric came back zero. Now anchored to recent days. |
| **security scan waiver** (1) | The waiver fixture carried `expiresAt: 2026-07-19`. It expired, the scan blocked the package, the test failed — **which is the expiry check doing its job**. The fixture was the wrong thing; dates are now relative. |

That also answers the "how long has this been red" question from the issue: at least
since late June, for the governance ones.

### Three were contracts that moved

- **credits error** — `7faea27bf` canonicalised the intelligence error codes, so the
  provider's `CREDITS_EXCEEDED` is translated to `QUOTA_EXHAUSTED` at the boundary
  (along with the two `INTELLIGENCE_PROVIDER_*` quota codes). **402 is asserted
  unchanged** — the canonicalisation moved the code, not the status, so callers
  switching on HTTP status are unaffected.
- **pricing locale** — restructured from flat keys (`freeTitle`, `freePrice`,
  `freeFeature1`…) into `pricing.plans.<plan>.{name,desc,featureN}`, and gained a fourth
  plan. The old flat list matched **nothing**, so the test was asserting an *empty*
  contract rather than a complete one. It now walks the real shape and compares plan key
  sets across locales, so a plan can't be added to one and not the other.
- **app token options** — tokens are issued with `LONG_TERM_APP_TOKEN_OPTIONS`. The whole
  options object is asserted rather than `deviceId` alone, because `grantType` is what
  the sibling test relies on to refuse refreshing a short-term token — issuing something
  other than `'long'` here would defeat that check without failing anything.

### And six from the first commit

Signing envelope (2 — fixture predates the envelope, so the file's actual subject never
ran), four source markers (component moved, options object gained `lazy`/`server`,
a computed was inlined, a call was reformatted), and one D1 double missing `batch` after
the merge moved to a single atomic call.

### Verification

- **179 files / 956 tests / 0 failures**, from 13 failing.
- **No lint delta**: 13 pre-existing errors across the five files in the second commit,
  45 across the five in the first — same before and after.

### DIFF
--- a/apps/nexus/app/pages/dashboard/notifications.test.ts
+++ b/apps/nexus/app/pages/dashboard/notifications.test.ts
@@ -51,7 +51,11 @@ describe('dashboard notification inbox UI contract', () => {
     expect(page).toContain('urlBase64ToArrayBuffer(browserPushPublicKey.value)')
     expect(page).toContain('runtimeConfig.public?.notificationWebPush?.publicKey')
     expect(page).toContain('browserPushStatusLabel')
-    expect(page).toContain('browserNotificationActionDisabled')
+    // The named browserNotificationActionDisabled computed was inlined into the
+    // template, so assert the condition itself rather than the identifier: the test
+    // cares that the action is gated on busy / denied / unsupported, not on where that
+    // expression lives.
+    expect(page).toMatch(/browserNotificationBusy \|\| browserNotificationPermission === .denied. \|\| browserNotificationPermission === .unsupported./)
     expect(page).toContain('browserNotificationPermissionLabel')
     expect(page).toContain('dashboard.notifications.browser.testSent')
     expect(page).toContain('dashboard.notifications.browser.webPushSubscribed')
--- a/apps/nexus/app/pages/store-page-performance.test.ts
+++ b/apps/nexus/app/pages/store-page-performance.test.ts
@@ -31,7 +31,10 @@ describe('store page remote search performance boundaries', () => {
     expect(storePage).toContain('const STORE_PLUGIN_PAGE_SIZE = 50')
     expect(storePage).toMatch(/function resolveStoreSearchQuery\(offset = 0\) \{[\s\S]*compact: 1,[\s\S]*q: activeSearch\.value \|\| undefined,[\s\S]*category: activeCategory\.value === 'all' \? undefined : activeCategory\.value,[\s\S]*limit: STORE_PLUGIN_PAGE_SIZE,[\s\S]*offset,[\s\S]*\}/)
     expect(storePage).toContain('const storeSearchQuery = computed(() => resolveStoreSearchQuery(0))')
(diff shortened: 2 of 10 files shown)


TalexDreamSoul:

Verified on that branch before closing: the pricing locale keys were restructured and the flat `freeFeature1` list is gone.

PR #1214, opened 2026-08-08
test(nexus): unpin the login-history fixture before it detonates (#1210)

TalexDreamSoul:
Refs #1210. Contains #1212's commits plus one more.

Four of the thirteen failures in #1210 were **fixtures that aged out of a window
measured from `Date.now()`** — not code changes. That's a category, not a coincidence,
so I went looking for the rest of it.

### The method

If time is what breaks a test, time is the thing to vary. I ran each suite with the wall
clock advanced and diffed against its own baseline. Anything that fails only under the
shift is pinned to the calendar and will fail on its own eventually.

### What it found

| suite | +90d | verdict |
|---|---|---|
| `apps/nexus` | **1 new failure** | `authStore-login-history` — fixed here |
| `apps/core-app` | no change (47 failing before and after, same tests) | clean |
| `packages/test` | no change (60 failing before and after, same tests) | clean |

`listLoginHistory` prunes rows older than its 90-day window **before listing**
(`authStore.ts:2804-2808`). The fixture row was pinned to `2026-06-21` — inside that
window today, outside it from **2026-09-19**. It would have started failing then and
looked like a login-history regression rather than fixture rot.

### After the fix

`apps/nexus` is green at **+90d, +365d and +1095d**. A three-year horizon finds nothing
else of this kind.

### On the other 145 files

145 test files carry pinned ISO dates. Most are harmless — a fixed date only rots when
something compares it to *now*. I didn't touch them, and I don't think a blanket rule
against pinned dates is worth it; the clock-shift run is a better detector than a lint
rule, because it tests the actual comparison rather than the literal.

Worth considering as a periodic CI job rather than a one-off: it costs one extra suite
run and catches this whole class before it reaches a required gate.

### DIFF
--- a/apps/nexus/app/pages/dashboard/notifications.test.ts
+++ b/apps/nexus/app/pages/dashboard/notifications.test.ts
@@ -51,7 +51,11 @@ describe('dashboard notification inbox UI contract', () => {
     expect(page).toContain('urlBase64ToArrayBuffer(browserPushPublicKey.value)')
     expect(page).toContain('runtimeConfig.public?.notificationWebPush?.publicKey')
     expect(page).toContain('browserPushStatusLabel')
-    expect(page).toContain('browserNotificationActionDisabled')
+    // The named browserNotificationActionDisabled computed was inlined into the
+    // template, so assert the condition itself rather than the identifier: the test
+    // cares that the action is gated on busy / denied / unsupported, not on where that
+    // expression lives.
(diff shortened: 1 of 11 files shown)


TalexDreamSoul:
Superseded by #1223, which targets `TalexDreamSoul/app-shell-v2` — the branch this repo actually merges into. This one was opened against `master` by mistake, where it could never land. The replacement is verified on that branch.

PR #1220, opened 2026-08-08
test(core-app): give plugin-module one filesystem model, inodes included (#323)

TalexDreamSoul:
Refs #323. `plugin-module`: **19 failing → 15**, and the remaining 15 are a different
kind of problem — detail at the end.

### One cause for all 19

Every failure reported `PLUGIN_UNINSTALL_TEARDOWN_FAILED`, and every one died at the
**first** barrier, `closeAdmission`, with `sqliteClosePlugin` never called.

`closeAdmission` pins the data and code roots, and `captureDirectoryIdentity` takes that
pin with **real `node:fs`** while everything else here goes through the **`fs-extra`
mock**. They disagreed by construction:

| | sees |
|---|---|
| pin (real `node:fs`) | `/fixture/plugin-data/calendar` → ENOENT → `state: 'missing'` |
| verify (mocked `fs-extra`) | the mocked directory → exists |

`assertPinnedDirectory` then throws `PLUGIN_DISPOSITION_OWNER_PATH_CHANGED`. The mocked
`lstat` also returned no `dev`/`ino` while the check compares both, so `Number(undefined)`
would have failed it anyway.

### The fix models inodes, rather than stubbing the check out

`node:fs`, `node:fs/promises` and `fs-extra` now read **one** model, so the pinning is
still exercised: `dev`/`ino` must match, a symlinked path still fails, and **an inode must
survive the quarantine rename**.

That last part is the whole point. Deletion is two-phase now — rename to
`<path>.uninstall-<uuid>.recovery`, re-pin against the identity captured at admission,
then remove — and it's safe in production *precisely because an inode follows the
directory rather than the path*. A path-derived identity would pass the admission check
and fail the post-rename one, so the double keeps an inode table that `rename` moves
entries within.

It also means `fse.remove` sees the **recovery** path, so ordering matchers keyed on the
original literals stopped matching anything. Normalised back, so they assert the directory
being removed rather than which phase called remove.

**Deliberately not stubbed away:** this is anti-TOCTOU hardening — it's what stops a
directory being swapped for a symlink between admission and delete. The suite's own
*"refuses a symlinked plugin code owner and never follows it during uninstall"* is exactly
the test that must keep working against a faithful double.

### The remaining 15 need expectations rewritten, not doubles fixed

The cleanup no longer walks `cache`, `temp` and `data` separately — it quarantines the
data root **whole**. Ordering arrays still listing `temp-remove`, `data-remove` and
`cache-remove` describe a design that's gone:

```
expected  … temp-remove, data-remove, cache-remove, data-root-remove, code-remove, finalize …
actual    … secret-legacy, secret-v2, data-root-remove, plugin-row-remove, code-remove, residual-verification
```

That's a judgement about what the new cleanup should guarantee, not a mock repair. Mixing
it in would bury the mechanical part, so I stopped here.

### Verification

`plugin-module`: 15 failing / 20 passing (was 19 / 16). ESLint clean under
`apps/core-app`'s own config with `--max-warnings=0` — which is the gate `pnpm lint`
actually runs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved plugin uninstall handling for renamed files, symbolic links, and quarantine recovery paths.
  * Updated uninstall failure and success scenarios to ensure cleanup targets are resolved consistently.
  * Strengthened coverage for file existence and deletion-order behavior during plugin removal.

### DIFF
--- a/apps/core-app/src/main/modules/plugin/plugin-module.test.ts
+++ b/apps/core-app/src/main/modules/plugin/plugin-module.test.ts
@@ -26,6 +26,10 @@ const mocks = vi.hoisted(() => {
   const eventHandlers = new Map<unknown, (event: unknown) => void>()
   const removedPaths = new Set<string>()
   const symbolicPaths = new Set<string>()
+  const inodes = new Map<string, number>()
+  const renamedAway = new Set<string>()
+  const renamedInto = new Set<string>()
+  let nextInode = 1000
   let capturedManagerFactory: CapturedManagerFactory | null = null
   const transportOn = vi.fn(
     (channel: unknown, handler: (payload: unknown, context: unknown) => unknown) => {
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:

Verified on that branch before closing: `plugin-module.test.ts` passes **34/34** — the filesystem/pinning problem I diagnosed is solved.

The diagnosis still stands as a record of *why* it was failing, if it is ever useful: the pin used real `node:fs` while verification used the mocked `fs-extra`, and deletion is two-phase so an inode has to survive the quarantine rename.

PR #1222, opened 2026-08-08
fix(tuff-native): expect the frozen-compose feature the capability always adds (#1125)

TalexDreamSoul:
Fixes the **Windows leg** of #1125. Retarget of #1207 (opened against `master` by mistake).

`verify:screenshot-production` fails on every branch:

```
AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected
  [ 'display', 'region', + 'frozen-compose' ]
```

`ScreenshotCapability::features()` appends `frozen-compose` **unconditionally**
(`native-screenshot/src/capability.rs:35-39`), so the win32/linux branch was asserting a
list that **cannot occur**. Not platform drift — the assertion has been wrong since
`f184d0966` added the feature.

The Rust side already agrees on the three-feature list
(`backend_contract_tests.rs:377`). The JS verifier was the only place that disagreed.

**macOS never caught it** because its branch asserts `engine` and `state` only, never
`features` — which is also why this reads as a Windows-specific bug when it isn't.

### Verified

Built the production screenshot addon on this branch and ran the verifier on darwin:
passes. That branch is untouched; this only corrects the win32/linux expectation. I can't
execute the Windows path here, so this PR's Windows leg is the real check.

### DIFF
--- a/packages/tuffex/scripts/audit-package-exports.mjs
+++ b/packages/tuffex/scripts/audit-package-exports.mjs
@@ -40,7 +40,13 @@ async function collectFiles(dir, predicate) {
 async function collectComponentSubpaths() {
   const dirents = await readdir(componentSrcRoot, { withFileTypes: true })
   return dirents
-    .filter(dirent => dirent.isDirectory() && dirent.name !== 'utils')
+    // Every directory under components/src is treated as a publishable component
+    // subpath, so anything that is not one has to be excluded here. 'utils' is shared
+    // code; '__tests__' arrived with 14274af8b and is a test directory, which the build
+    // correctly does not emit a bundle for -- leaving the audit to report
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Superseded in effect by #1237, which carries this commit **plus** the `libgbm-dev` fix the Linux leg also needs — the two together are what make ubuntu green, and this alone doesn't.

Worth recording here: with the link error fixed, **linux hit this same `frozen-compose` assertion independently**. I argued on this PR from the Windows failure alone that the `win32 || linux` branch asserts a list that cannot occur; linux has now demonstrated it. That's the confirmation it was a shared bug rather than a Windows quirk.

Leaving this open in case you'd rather take the two changes separately.

PR #1236, opened 2026-08-08
fix(utils): order canonical signing JSON by codepoint, not ICU collation

### LINKED ISSUE #894: [audit/security] Plugin signing canonical JSON sorts keys with localeCompare, so it is not deterministic

**Severity:** medium  ·  **Category:** security  ·  **Domain:** security

**Location:** `packages/utils/plugin/signing.ts`:180

## What

canonicalize() orders object keys with localeCompare(left, right, 'en') instead of a codepoint sort. This canonical form is the exact byte string that gets signed and verified across three different JS runtimes.

## Evidence

```
function canonicalize(value: unknown): unknown {
  ...
  if (isRecord(value)) {
    const normalized: Record<string, unknown> = {}
    for (const key of Object.keys(value).sort((left, right) => left.localeCompare(right, 'en'))) {
      const child = value[key]
      if (child === undefined)
        continue
      normalized[key] = canonicalize(child)
    }
    return normalized
  }
```

serializePluginSigningValue feeds three runtimes with different ICU collation data: the publisher CLI signs via packages/tuff-cli-core/src/plugin-signer.ts:150, Nexus verifies on Cloudflare Workers via apps/nexus/server/utils/pluginSigning.ts:438, and the Electron main process verifies via apps/core-app/src/main/modules/plugin/signature-verifier.ts:242. localeCompare is ICU/locale-data dependent (Node small-icu vs full-icu vs the Workers runtime), so the same object can serialize to different byte strings on signer and verifier, producing spurious signature failures. Worse, ICU collation returns 0 for distinct strings that differ only by collation-ignorable code points (e.g. a soft hyphen U+00AD or ZWJ U+200D inside a path); when the comparator returns 0, Array.prototype.sort leaves those keys in input order, which for serializePluginFileMap comes from attacker-chosen tar entry ordering. A publisher can therefore craft plugin file names whose canonical serialization -- and thus the signed fileMapSha256 -- is not a function of the file map alone.

## Suggested direction

Sort with a plain codepoint comparator (left < right ? -1 : left > right ? 1 : 0) and reject duplicate keys, so canonicalization is byte-deterministic on every runtime.

---
<sub>audit-domain: security · confidence: confirmed</sub>

TalexDreamSoul:
Closes #894.

> ⚠️ **This changes the canonical form, so existing plugin signatures will need re-signing.** Details below — please read before merging.

## The defect

`canonicalize` sorted keys with `localeCompare(left, right, 'en')`. That string is the exact bytes **three runtimes** sign and verify:

| runtime | role |
|---|---|
| publisher CLI (`plugin-signer.ts:150`) | signs |
| Nexus on Cloudflare Workers (`pluginSigning.ts:438`) | verifies |
| Electron main (`signature-verifier.ts:242`) | verifies |

All three import `serializePluginSigningPayload` from this one file, so one comparator decides all of them — and ICU collation data differs between small-icu, full-icu and the Workers runtime.

**The sharper half**, verified on this runtime:

```
'a­b'.localeCompare('ab', 'en')  ===  0
```

A comparator returning `0` leaves those keys in **input order**, and for `serializePluginFileMap` that order comes from tar entries the publisher chooses. So `fileMapSha256` was not a function of the file map alone.

## On "reject duplicate keys"

`Object.keys` already yields unique keys, so there is nothing to reject. What was missing is a **total ordering** — a codepoint comparator never returns 0 for distinct strings, which is precisely what fixes the collision case.

## The breaking part

Codepoint order puts `B` before `a`; `localeCompare('en')` puts `a` before `B`. Any file map with mixed-case names — `README.md` beside `src/index.js` is the normal case — canonicalizes differently now, its `fileMapSha256` changes, and that hash sits **inside the signed payload**.

**Plugins signed before this change will fail verification and need re-signing.**

If that is unacceptable, the alternative is gating the comparator on the payload's existing `policyVersion` so old signatures verify under the old rule. I did not build it: it doubles the canonicalization surface, and choosing between *re-sign* and *keep two canonical forms alive* is a call for whoever owns the signing rollout.

## Tests

Seven, **three failing** against the previous comparator: the codepoint ordering itself, the soft-hyphen collision, and a file map whose serialization must not depend on archive order.

Two are controls — arrays keep their order (that is data, not key order), and an ordinary payload still serializes correctly, since an ordering fix that broke serialization would satisfy the equality assertions by returning the same wrong thing twice.

Consumer suites pass unchanged: **1183** in core-app plugin, **191** in nexus server utils, and the 14 existing signing tests in utils.

### DIFF
--- a/packages/utils/__tests__/plugin-signing-canonical.test.ts
+++ b/packages/utils/__tests__/plugin-signing-canonical.test.ts
@@ -1,0 +1,71 @@
+import { describe, expect, it } from 'vitest'
+import { serializePluginFileMap, serializePluginSigningValue } from '../plugin/signing'
+
+/**
+ * The canonical byte string that plugin signatures cover (#894).
+ *
+ * Key order was decided by `localeCompare(left, right, 'en')`. That string is signed by the
+ * publisher CLI and verified independently by Nexus on Cloudflare Workers and by the Electron
+ * main process, so an ICU difference between those runtimes could make the same object
+ * serialize differently on signer and verifier.
+ *
+ * The sharper problem is collation-ignorable code points: ICU reports distinct strings as
+ * equal, `Array.prototype.sort` then leaves them in input order, and for a file map that order
+ * comes from tar entries the publisher chooses.
+ */
+describe('canonical key ordering', () => {
+  it('is stable regardless of the order keys arrive in', () => {
+    const forward = serializePluginSigningValue({ a: 1, B: 2, c: 3 })
+    const reverse = serializePluginSigningValue({ c: 3, B: 2, a: 1 })
+    expect(forward).toBe(reverse)
+  })
+
+  it('orders by codepoint, so uppercase precedes lowercase', () => {
+    // The observable difference from localeCompare('en'), which puts 'a' before 'B'.
+    expect(serializePluginSigningValue({ a: 1, B: 2 })).toBe('{"B":2,"a":1}')
+  })
+
+  it('separates keys that ICU collation reports as equal', () => {
+    // Confirmed against this runtime: 'a­b'.localeCompare('ab', 'en') === 0. With a
+    // comparator returning 0 the sort is input-order dependent, which is the defect.
+    const soft = 'a­b'
+    expect(soft.localeCompare('ab', 'en')).toBe(0)
+
+    const forward = serializePluginSigningValue({ [soft]: 1, ab: 2 })
+    const reverse = serializePluginSigningValue({ ab: 2, [soft]: 1 })
+    expect(forward).toBe(reverse)
+  })
+
+  it('gives a file map the same hash input whatever order the archive listed it in', () => {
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
**Ready, but deliberately not merged — this needs your call.**

All checks green (Typecheck, PR Quality, App suites core-app + nexus, Integration suite; `Documentation Quality` is the branch-wide `.trellis` failure tracked in #1125).

I am not merging it myself because it invalidates existing plugin signatures, and I asked that question on #894 rather than assuming an answer. Merging over my own question would be incoherent.

What I could and could not establish from the repo:

- the signing flow **has** been exercised — there is a `publisher-signature.json` in the July 2026 acceptance evidence
- whether any plugins are signed **and in circulation** is not something the repository can tell me

If nothing signed is in the field, merge as is. If something is, say so and I will replace this with a `policyVersion`-gated comparator so old signatures keep verifying under the old rule.

Issue #1203, opened 2026-08-08
Partial searchProvider coverage is invisible: one declaration silences the check for every other push feature


TalexDreamSoul:
`touch-intelligence` declares **5 push features** and **1** `searchProvider`. The
resolver reports **zero issues**. The other four push to root results with no provider
descriptor — so no admission policy, no `defaultState: 'ask'`, no
`requiresUserConsent`.

Found by `packages/test/src/plugins/manifest-boundary.test.ts`, which is failing on it.
That test is doing its job; this is not stale fixture drift.

### Why nothing warns

`getSearchProviderManifestCoverage` (`packages/utils/search/indexing-source.ts:507-520`):

```ts
needsExplicitProviderMigration:
  pushFeatureIds.length > 0 &&
  (!Array.isArray(manifestProviders) || manifestProviders.length === 0),
```

The condition is **`length === 0`**, so it is all-or-nothing. Declare a single provider
and the flag goes false no matter how many push features remain uncovered.
`derivedFromPushFeatures` follows it, and `SEARCH_PROVIDER_DERIVED_FROM_PUSH_FEATURE` —
the only warning on this path — fires only when providers were *derived*, which by then
they never are.

The coverage object already carries both numbers it would need:

```ts
pushFeatureCount: 5,
explicitProviderCount: 1,
```

Nothing compares them, and there is no issue code for the partial case.

### Reproduction

Resolving `plugins/touch-intelligence/manifest.json` through
`resolveSearchProviderManifestDescriptors` with its own declared permissions:

```
push features = intelligence-ask, intelligence-rewrite, intelligence-summarize,
                intelligence-explain, intelligence-command-registry
descriptors    = touch-intelligence.intelligence-ask
issues         = []
derivedFromPushFeatures = false
```

`apps/core-app/src/main/modules/plugin/plugin-loaders.ts:519-524` then returns those
descriptors because `coverage.hasExplicitProviders` is true, and pushes no issue.

### What I did not verify

Whether the four uncovered features are actually *permitted* to reach root results at
runtime, or are dropped further down the search pipeline. I traced as far as the
resolver and the loader; the manifest-level boundary is definitely not enforcing them,
but the end-to-end consequence needs someone who knows the search path. That's the
difference between "a declaration is missing" and "four features bypass consent", and I
don't want to overstate which one this is.

### Suggested shape

Add an issue code for partial coverage — push feature ids with no matching provider —
[...]

PR #1240, opened 2026-08-08
feat(search): report push features that no declared provider covers (#1203)

TalexDreamSoul:
Refs #1203. Makes the gap visible; **does not** decide what to do about it.

`touch-intelligence` declares **5 push features and 1 `searchProvider`**, and the resolver
reports **zero issues**. `needsExplicitProviderMigration` is the only check on this path and
it requires `manifestProviders.length === 0` — so declaring a single provider silences it
for every other push feature. The coverage object already carried `pushFeatureCount` and
`explicitProviderCount`; nothing compared them.

### The new code

`SEARCH_PROVIDER_PARTIAL_PUSH_FEATURE_COVERAGE`, raised when providers are declared but some
push feature has none of its own. Matched **by `featureId`, not by count** — counts can
coincide while the ids don't.

### One false positive found and excluded

My first version reported `touch-browser-bookmarks` too. Its manifest pairs **one unnamed
provider with one push feature** — an implicit mapping, and nothing to conclude from. Firing
there would hit every hand-written single-provider manifest.

So the check only engages once *some* provider names a feature. If none carries a
`featureId`, it reports nothing rather than everything.

### Against the real manifests

```
touch-intelligence       intelligence-rewrite, intelligence-summarize,
                         intelligence-explain, intelligence-command-registry
touch-browser-bookmarks  (nothing)
```

Exactly the four from the issue.

### What this deliberately leaves open

Those four either need providers with an admission policy, or shouldn't be declaring
`push: true`. That's a manifest decision and a security one — I've made it visible rather
than guessing which way it goes. The point of #1203 was that nothing reported it at all.

### Verification

- `packages/utils`: **132 files, 1057 tests passing**.
- Extended the coverage contract test with the partial case, and pinned the implicit-mapping
  case so the false positive can't come back.
- Lint clean under `packages/utils/eslint.config.js`.

### DIFF
--- a/apps/core-app/src/main/modules/box-tool/search-engine/search-provider-registry.ts
+++ b/apps/core-app/src/main/modules/box-tool/search-engine/search-provider-registry.ts
@@ -60,6 +60,7 @@ type DestroyableSearchProvider = ISearchProvider<ProviderContext> & {
 
 const SEARCH_PROVIDER_ISSUE_CODES: Record<SearchProviderRegistryIssue['code'], true> = {
   SEARCH_PROVIDER_DERIVED_FROM_PUSH_FEATURE: true,
+  SEARCH_PROVIDER_PARTIAL_PUSH_FEATURE_COVERAGE: true,
   SEARCH_PROVIDER_INVALID: true,
   SEARCH_PROVIDER_POLICY_BLOCKED: true,
   SEARCH_PROVIDER_PERMISSION_MISSING: true,
--- a/apps/core-app/src/main/modules/plugin/plugin-loaders.ts
+++ b/apps/core-app/src/main/modules/plugin/plugin-loaders.ts
@@ -542,6 +542,20 @@ abstract class BasePluginLoader {
       return
     }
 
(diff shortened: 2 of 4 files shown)


TalexDreamSoul:
**Fixed the typecheck failure this PR introduced.**

```
search-provider-registry.ts(61,7): error TS2741: Property
'SEARCH_PROVIDER_PARTIAL_PUSH_FEATURE_COVERAGE' is missing
```

`SEARCH_PROVIDER_ISSUE_CODES` is a `Record` keyed on the issue-code union, so adding a code
is *meant* to break every consumer until it says what to do with it. That's the design
working — my first commit added the code without answering it.

Two places had to answer:

- **`search-provider-registry`** admits the code, so the registry doesn't discard the issue.
- **`plugin-loaders`** surfaces it as a plugin warning, with a suggestion naming **both**
  ways out — declare a provider carrying each push feature's `featureId`, or drop `push`
  from features that shouldn't reach root results.

The suggestion deliberately offers both. Which is right for `touch-intelligence` is a
manifest decision, and a warning shouldn't presume it.

`typecheck:node` is clean. Web typecheck is **154 errors before and after** — all
pre-existing `.vue` import complaints in unrelated modules, verified by A/B rather than
assumed.

Issue #523, opened 2026-08-06
[audit/tech-debt] Circular import between search-core and plugin-features-adapter, both module-scope singletons


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** medium  ·  **Category:** circular-dep  ·  **Domain:** architecture

**Location:** `apps/core-app/src/main/modules/box-tool/search-engine/search-core.ts`:36

## What

search-core.ts and plugin/adapters/plugin-features-adapter.ts import each other's default export, and both construct their singleton at module evaluation time. Whichever module the bundler happens to evaluate second sees a partially-initialized binding for the other.

## Evidence

```
// apps/core-app/src/main/modules/box-tool/search-engine/search-core.ts:36, 2187
import PluginFeaturesAdapter from '../../plugin/adapters/plugin-features-adapter'
...
export default SearchEngineCore.getInstance()

// apps/core-app/src/main/modules/plugin/adapters/plugin-features-adapter.ts:24, 662
import searchEngineCore from '../../box-tool/search-engine/search-core'
...
export default new PluginFeaturesAdapter()
```

## Why it matters / failure scenario

Today this survives only because `SearchEngineCore`'s constructor (search-core.ts:268) does not touch `PluginFeaturesAdapter` — the reference is deferred to `registerDefaults()` (line 332, `this.registerProvider(PluginFeaturesAdapter)`). Moving that single `registerProvider` call into the constructor, or adding any other module-scope use of the adapter, registers `undefined` as a search provider: plugin features silently stop appearing in CoreBox results with no error at startup.

## Suggested direction

Break the cycle by having plugin-features-adapter receive the search engine through registration (search-core calls `adapter.attach(this)`) instead of importing the singleton back.

---
<sub>audit-domain: architecture · confidence: confirmed</sub>

TalexDreamSoul:
Fixed in #1292, taking the direction you suggested — though the cheaper cut went the other way, and it is worth saying why I did not take it.

**The reverse cut was one edit against eight.** search-core references the adapter exactly once (`registerDefaults`, line 332); the adapter references the engine eight times. Cutting `search-core -> adapter` would have meant moving `registerProvider(PluginFeaturesAdapter)` out of `registerDefaults()` to a composition root — one line — but it would have split the six-provider list across two files. Keeping the provider registrations together is worth the extra edits, so `attach()` it is.

[...]

PR #1266, opened 2026-08-08
fix(plugin): make IFeatureCommand.value the string a manifest can actually carry

TalexDreamSoul:
Fixes the two concrete failures in #885. The larger type-split it proposes is left open — see below.

## The defect

`IFeatureCommand.value` admitted `RegExp` and `FeatureCommandMatcher`, but features are declared in `manifest.json` and cross IPC, where both degrade to a plain string or `{}`. The matcher trusted the type:

| command type | before |
|---|---|
| `regex` | `(command.value as RegExp).test(...)` → **`TypeError: value.test is not a function`** on every real declaration |
| `function` | no branch at all → fell through to `default: return false`, so the feature **silently never matched** |

## The fix

- `value` narrowed to `string | string[]`
- `regex` compiles the pattern at match time, cached per pattern; an invalid pattern is ignored with a warning rather than thrown
- `function` now logs why it cannot work instead of failing silently

## Narrowing a published type is safe here

Nothing used the removed arms:

- `FeatureCommandMatcher` was referenced **only** by its own declaration and that union arm
- no manifest in `plugins/` declares `type: "regex"` or `"function"`
- nothing constructs a `RegExp` value anywhere

`typecheck:node` passes unchanged. The type stays exported and is marked `@deprecated` so existing imports still resolve — an external author passing a `RegExp` was already broken at runtime, so this surfaces an existing bug rather than creating one.

## Verified by mutation

Against the unfixed matcher, **5 of the 6 new cases fail**:

```
unfixed -> 5 failed | 16 passed
fixed   -> 21 passed
```

The sixth asserts `match` / `contain` / `over` are untouched — a careless rewrite of the switch would break it.

## Note on a lint trap hit while doing this

`no-console` is an error in core-app, and my first version used `console.warn`. Worth recording that `npx eslint … | tail` reports the **pipe's** exit code, not eslint's — it printed `exit=0` while eslint was actually exiting 1 with 2 errors. The warnings now go through the file's existing `pluginFeaturesLog`.

## Left open

The issue's suggested direction — splitting a manifest-facing type from a runtime type and compiling `regex` once at manifest load rather than at match time — is a design change beyond removing the crash. Reported on #885.

### DIFF
--- a/apps/core-app/src/main/modules/plugin/adapters/plugin-features-adapter.test.ts
+++ b/apps/core-app/src/main/modules/plugin/adapters/plugin-features-adapter.test.ts
@@ -1,11 +1,11 @@
 import { readFileSync } from 'node:fs'
 import path from 'node:path'
 import type { IProviderActivate } from '@talex-touch/utils'
-import type { IPluginFeature, ITouchPlugin } from '@talex-touch/utils/plugin'
+import type { IFeatureCommand, IPluginFeature, ITouchPlugin } from '@talex-touch/utils/plugin'
 import type { CoreBoxInputChangeRequest } from '@talex-touch/utils/transport/events/types'
 import { describe, expect, it, vi } from 'vitest'
 import { PluginStatus } from '@talex-touch/utils/plugin'
-import { PluginFeaturesAdapter } from './plugin-features-adapter'
+import { isCommandMatch, PluginFeaturesAdapter } from './plugin-features-adapter'
 import { pluginModule } from '../plugin-module'
 import { PluginViewLoader } from '../view/plugin-view-loader'
 
@@ -21,6 +21,18 @@ function createAdapter(): PluginFeaturesAdapter {
   adapter.attach(searchEngineHost)
   return adapter
 }
+
+const featureLogWarn = vi.hoisted(() => vi.fn())
+
+vi.mock('@talex-touch/utils/common/logger', async (importOriginal) => ({
+  ...(await importOriginal<Record<string, unknown>>()),
+  getLogger: () => ({
+    warn: featureLogWarn,
+    info: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn()
+  })
+}))
 
 vi.mock('../plugin-module', () => ({
   pluginModule: {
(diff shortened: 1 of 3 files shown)


TalexDreamSoul:
Left unrebased deliberately — this one needs a decision rather than a mechanical resolution.

`plugin-features-adapter.test.ts` has since been rewritten around a different harness: the base now builds each adapter through `createAdapter()` and attaches a `searchEngineHost` double (`getActivationState` / `activateProviders` / `deactivateProvider`), where this branch still imports `searchEngineCore` directly. Porting the new `isCommandMatch` cases across means re-expressing them against that host double, and I did not want to guess at the intended assertions while doing it.

Everything else in the stale queue has been rebased and re-run. Say the word if you want me to port these cases, or push a rebase yourself and I will verify and merge.

TalexDreamSoul:
## 已解决冲突并 rebase 到最新基线

冲突出在 `plugin-features-adapter.test.ts`,原因是这个 PR 写于 **#523 之前**:基线后来把 adapter 与 search-core 解耦了,而本 PR 仍在 `vi.mock('../../box-tool/search-engine/search-core')`。落后基线 **131 个提交**。

### 两处冲突都是可加性的,不是互斥的

**导入与 setup:** 保留基线的 `attach(searchEngineHost)` 写法,只把本 PR 新增的部分并进去 —— `isCommandMatch` 的导入,以及 `featureLogWarn` 的 logger mock。

**丢弃了**本 PR 里的 `import searchEngineCore` 和对 search-core 的 mock —— 基线现在有一条测试专门断言 adapter **不得** import search-core,把它们留下会让那条测试失败。

**describe 块:** 两个都保留。基线的 `search engine coupling (#523)` 和本 PR 的 `plugin feature command matching` 测的是不同的东西。

### 验证

```
✓ plugin-features-adapter.test.ts (24 tests)
  Tests  24 passed (24)
```

并且确认两边的标志性测试都还在(不是被我合掉了):

- 基线侧:`does not import the search engine`
- 本 PR 侧:`compiles a regex string pattern instead of throwing on it`

`tsc -p tsconfig.node.json` 对该文件无报错;core-app 的 lint-staged 命令退出 0。

现在状态:**MERGEABLE**(此前是 CONFLICTING)。ahead 1 / behind 0。

Issue #889, opened 2026-08-06
[audit/sdk] @talex-touch/tuff-core is configured for publishing but its entry point is empty


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** low  ·  **Category:** maintainability  ·  **Domain:** sdk

**Location:** `packages/tuff-core/src/index.ts`:1

## What

The package declares `publishConfig.access: public`, `main: dist/index.js` and `types: dist/index.d.ts`, but `src/index.ts` contains only `export {}`, there is no `dist`, no `prepublishOnly`, and no package in the monorepo depends on it.

## Evidence

```
export {}
```

## Why it matters / failure scenario

Anyone running `npm publish` in `packages/tuff-core` ships a package whose `main` and `types` point at files that were never built, so `import ... from '@talex-touch/tuff-core'` fails to resolve for every consumer. The described 'low-level build and publish primitives for Tuff' do not exist.

## Suggested direction

Either implement the package and add a `prepublishOnly: npm run build`, or remove it from the workspace until it has content.

---
<sub>audit-domain: sdk · confidence: confirmed</sub>

TalexDreamSoul:
## 已处理危害:PR #1272。但**包本身的去留仍未决**,issue 保持 open

### 缺陷确认

`packages/tuff-core/src/index.ts` 确实只有 `export {}`,没有 `dist`,`main`/`module`/`types` 全指向 `dist/*`,没有 `prepublishOnly`,仓内**没有任何包依赖它**。正文逐条准确。

### 真正该修的是那道闸门,不是这个包

仓里**已经有**一道发布校验,而它**放行了这个包**:

```
$ node scripts/validate-publish-manifests.mjs
[publish-manifest] Validation passed (source): …, @talex-touch/tuff-core, …
```

它的 source 阶段只检查依赖协议(`catalog:` / `workspace:`),没有任何一处验证入口文件能否存在。这个假绿才是会重复发生的问题,所以改动加在那里:**入口字段指向一个工作区里不存在的路径时,该包必须有 `prepublishOnly`**,发布就无法跳过构建。

### 加上检查后,立刻发现正文没提的第二个包

```
- package: @talex-touch/tuff-core   field: scripts.prepublishOnly
- package: @talex-touch/tuff-cli    field: scripts.prepublishOnly
```

[...]

PR #1272, opened 2026-08-08
build: fail publish validation when an entry points at an unbuilt output

TalexDreamSoul:
Addresses the harm in #889. The package's *fate* stays open — see the last section.

## The defect

`@talex-touch/tuff-core` declares `main`/`module`/`types` under `dist/`, has no `dist`, no `prepublishOnly`, and an `export {}` entry. `npm publish` would ship a package that fails to resolve for every consumer.

## The durable problem is the gate, not the package

The repo already has a publish gate — and it **passed this package**:

```
$ node scripts/validate-publish-manifests.mjs
[publish-manifest] Validation passed (source): …, @talex-touch/tuff-core, …
```

Its source phase only checked dependency protocols; nothing verified that an entry could exist. So the fix goes there: an entry field pointing at a path absent from the tree now requires a `prepublishOnly` script, so publishing cannot skip the build.

## Adding the check immediately found a second package #889 does not mention

```
- package: @talex-touch/tuff-core   field: scripts.prepublishOnly
- package: @talex-touch/tuff-cli    field: scripts.prepublishOnly
```

`tuff-cli` has the identical shape — `dist/index.js`, a build script, no `prepublishOnly`. Both now declare it.

## Verified, not assumed

- **The build behind the promise actually runs.** `tsup` on `tuff-core` exits 0 and produces `dist/index.js` + `index.d.ts`. A `prepublishOnly` pointing at a broken build would just move the failure. The `dist` was removed again afterwards — it is gitignored build output, not part of this change.
- **The check fires.** Removing `prepublishOnly` from `tuff-core` again fails validation with exit 1, naming the package and the missing field. A gate that cannot fail is what created this situation in the first place.

## Deliberately not decided

`tuff-core`'s entry is still `export {}` and **nothing in the monorepo depends on it**. Whether to implement the "low-level build and publish primitives" or retire the package is a product call, and #889 offers both. This change only makes an accidental publish impossible, which is the concrete harm the issue describes. The issue stays open for the fate question.

### DIFF
--- a/packages/tuff-cli/package.json
+++ b/packages/tuff-cli/package.json
@@ -23,6 +23,7 @@
   },
   "scripts": {
     "build": "tsup",
+    "prepublishOnly": "pnpm run build",
     "dev": "tsup --watch",
     "lint": "eslint \"src/**/*.{ts,tsx}\"",
     "lint:fix": "eslint \"src/**/*.{ts,tsx}\" --fix",
--- a/packages/tuff-core/package.json
+++ b/packages/tuff-core/package.json
@@ -20,6 +20,7 @@
   },
   "scripts": {
     "build": "tsup src/index.ts --format esm --dts --splitting --clean --target node24 --platform node",
+    "prepublishOnly": "pnpm run build",
     "dev": "tsup src/index.ts --format esm --dts --splitting --clean --target node24 --platform node --watch",
     "lint": "eslint \"src/**/*.{ts,tsx}\"",
     "lint:fix": "eslint \"src/**/*.{ts,tsx}\" --fix",
--- a/packages/tuffex/package.json
+++ b/packages/tuffex/package.json
@@ -45,6 +45,7 @@
     }
   },
   "scripts": {
+    "prepublishOnly": "pnpm run build",
     "build": "node ./node_modules/gulp/bin/gulp.js -f packages/script/build/index.ts",
     "build:vite": "vite build",
     "dev": "vite build --watch",
--- a/scripts/validate-publish-manifests.mjs
+++ b/scripts/validate-publish-manifests.mjs
@@ -161,17 +161,59 @@ function assertPublishConfig(packages) {
   return issues
 }
 
+/**
+ * A manifest can point `main` / `module` / `types` at a build output that nothing guarantees to
+ * exist. @talex-touch/tuff-core shipped exactly that shape -- `dist/index.js` with no dist, no
+ * prepublishOnly and an `export {}` entry -- and this validator passed it, so `npm publish`
+ * would have produced a package that fails to resolve for every consumer (#889).
+ *
+ * The rule: if an entry field points into a directory that is not present in the working tree,
+ * the package must have a `prepublishOnly` script, so publishing cannot skip the build.
+ */
+function findUnbuiltEntryIssues({ manifest, manifestPath, packageInfo }) {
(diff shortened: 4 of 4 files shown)


TalexDreamSoul:
Rebased onto the current base, but **not merging** — I cannot cleanly attribute the `PR Quality` failure, and I would rather say so than guess.

What fails is `check-release-gates/local-checks.test.mjs > validates packed publish manifests`: `publish-manifests` passes, `publish-manifests-pack` fails, because the check shells out to `corepack pnpm pack` per publishable package.

Two things stop me concluding it is pre-existing:

- It **does** fail on the base branch in my checkout — but my worktree has a symlinked `node_modules`, and the local error is `ERR_PNPM_CANNOT_RESOLVE_WORKSPACE_PROTOCOL` for `@talex-touch/utils` in `packages/tuffex`. That is my environment, not evidence about CI.
- The CI log for this job records the `AssertionError` but not the underlying `pack` stderr, so I cannot confirm the runner hits the same cause.

`PR Quality` passes on other PRs merged today, so the check is not universally red — it is presumably path-filtered onto `scripts/`, which this PR touches.

The branch is rebased and ready. If you can tell me whether the packed-manifest gate is expected to pass on a clean CI checkout right now, I will finish it — otherwise this needs a look from someone whose `node_modules` is a real install.

TalexDreamSoul:
## CI 失败已修:被漏掉的第三个包是 tuffex

`PR Quality` 报的是 `local-checks.test.mjs` 里的 `'fail' !== 'pass'`。根因不在测试,在**这个分支引入的规则自己没有被满足**。

本分支让 `validate-publish-manifests` 在"入口指向工作树中不存在的目录、且没有 `prepublishOnly`"时判失败。它给 `tuff-cli` 和 `tuff-core` 补了脚本,但 **`@talex-touch/tuffex` 是同一种形状** —— public、`files: ["dist"]`、`main`/`module`/`types` 全在 `dist/` 下 —— 却没有 `prepublishOnly`。

### 为什么本地看不见

开发者机器上有构建好的 `dist`,检查就过了。**CI 是干净检出**:`.gitignore:46` 覆盖 `dist/`,`git ls-files packages/tuffex/dist` 返回 0 个文件,入口路径在那里是真的不存在。

枚举全部可发布包中"入口指向 `dist/` 且无 `prepublishOnly`"的:

| | 结果 |
|---|---|
| 本次提交前 | 仅 `packages/tuffex` |
| 本次提交后 | 无 |

`scripts/check-release-gates/local-checks.test.mjs` 4/4。

原本我为此单独开了 #1439,现在并入本 PR —— 引入规则的改动应当同时让仓库满足它,否则这个 PR 单独合入就是红的。#1439 已关闭。

### 一处我先前报错的观察,需要更正

我一度看到这两条测试在 5000ms 超时,并把它当成脆弱点报了出来。**那是我的命令用错了**:裸跑 `vitest run <file>` 用的是 5s 默认值,而该套件自己的配置 `vitest.workspace-scripts.config.mjs` 设的是 `testTimeout: 120_000`。按仓库的方式跑,不存在超时问题。

TalexDreamSoul:
`PR Quality` is green after the rebase — the packed-manifest gate passes on a clean CI checkout. So the failure I could not attribute earlier was stale-base, not a real break, and my local reproduction was my worktree's symlinked `node_modules` as suspected. Merging.

Issue #504, opened 2026-08-06
[audit/a11y] Document lang is hardcoded 'en' and never updated on locale switch


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** medium  ·  **Category:** a11y  ·  **Domain:** a11y

**Location:** `apps/core-app/src/renderer/index.html`:2

## What

The renderer's root HTML declares lang="en" statically, and nothing in the renderer source ever writes document.documentElement.lang, so the declared document language stays English even when the UI is rendering Chinese.

## Evidence

```
<html lang="en"> — a repo-wide search for documentElement.lang across apps/core-app/src/renderer/src returns no matches, and the language-switching code in src/modules/lang/ does not set it
```

## Why it matters / failure scenario

A screen-reader user who switches the app to zh-CN has every Chinese string announced by an English speech synthesiser, which renders Chinese text as unintelligible sounds or skips it entirely. It also breaks browser-level font selection and hyphenation heuristics for CJK.

## Suggested direction

Set document.documentElement.lang from the active locale wherever the language preference is applied.

---
<sub>audit-domain: a11y · confidence: confirmed</sub>

TalexDreamSoul:
## 正文的主要论断**不成立**——文档语言其实一直在同步。PR #1277 只修了其中真实的那一小块

### 先说错在哪

> 「nothing in the renderer source ever writes `document.documentElement.lang`」

字面上成立,但结论是错的。该属性**一直在被同步**,只是用了另一种写法:

```
apps/core-app/src/renderer/src/modules/lang/i18n.ts:55
  document.querySelector('html')!.setAttribute('lang', locale)
```

它在 `setI18nLanguage` 里,而这个函数**初始化和每次切换都会调**:

```
i18n.ts:35          setI18nLanguage(i18n, options.locale)   // setup
useLanguage.ts:90   setI18nLanguage(i18n, lang)             // 切换语言时
```

所以正文设想的场景——「用户切到 zh-CN,但文档语言仍是 en,屏幕阅读器用英文口音念中文」——**不会发生**。切换后 `<html lang="zh-CN">` 是对的。

审计搜的是 `documentElement.lang` 这一种写法,没搜到就判定行为不存在。这是典型的「用单一写法做存在性判断」的假阴性。

### 真正错的那一小块

[...]

PR #1277, opened 2026-08-08
fix(core-app): make the static document lang match the app's default locale

TalexDreamSoul:
Addresses the one real part of #504. **The issue's main claim does not hold** — details below, since acting on it as written would have added duplicate code.

## The issue's premise is false

> *"nothing in the renderer source ever writes `document.documentElement.lang`"*

Literally true, and misleading. The attribute **is** kept in sync, via a different idiom:

```
apps/core-app/src/renderer/src/modules/lang/i18n.ts:55
  document.querySelector('html')!.setAttribute('lang', locale)
```

…inside `setI18nLanguage`, which runs both at setup and on every switch:

```
i18n.ts:35          setI18nLanguage(i18n, options.locale)   // setup
useLanguage.ts:90   setI18nLanguage(i18n, lang)             // on change
```

So a user switching to zh-CN does get `<html lang="zh-CN">`. The audit searched for one specific idiom, found nothing, and read that as the behaviour being absent.

## What is genuinely wrong

`index.html` ships `lang="en"` while the renderer defaults to `zh-CN` (`useLanguage.ts:17`). So the **first paint**, before i18n initialises, declares the wrong language. Narrow, but real for a screen-reader user — and it is the residue the issue was pointing at without knowing it.

One line: `lang="en"` → `lang="zh-CN"`.

## Why not something cleverer

Injecting the resolved locale into the HTML at build time would be the "complete" fix, but the resolved locale depends on stored settings and the system language, neither of which exist at build time. The static value can only ever be a default, so it should be the app's actual default.

## No test

The changed value is a static attribute in an HTML entry file with no renderer test harness, and the dynamic path it hands over to is already exercised by the existing `modules/lang` suite. Asserting a literal in `index.html` would restate the diff rather than test behaviour.

### DIFF
--- a/apps/core-app/src/renderer/index.html
+++ b/apps/core-app/src/renderer/index.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="zh-CN">
   <head>
     <meta charset="UTF-8" />
     <link rel="icon" type="image/svg+xml" href="/logo.png" />


TalexDreamSoul:
Rebased onto the current base (it was 61 commits behind with CI never re-run).

Checked the premise before pushing it through: `DEFAULT_LANGUAGE` in `language-preferences.ts` is `'zh-CN'`, so the static declaration and the resolved default now agree.

Scope note — the issue this came from, #504, is closed as **not a defect**: `setI18nLanguage` already sets `<html lang>` from the active locale, via `document.querySelector('html')` rather than `documentElement.lang`, and `setupI18n` runs before `createApp`, so the static value is corrected before any UI renders. This change is therefore consistency rather than a fix — the static value is only what a pre-bootstrap failure screen would carry. Still worth having for exactly that case. #1415 added the test for the dynamic path.

PR #1279, opened 2026-08-08
fix(core-app): make sidebar navigation items reachable by keyboard

### LINKED ISSUE #505: [audit/a11y] Primary navigation menu items are plain divs with no role, tabindex or key handler

**Severity:** high  ·  **Category:** a11y  ·  **Domain:** a11y

**Location:** `apps/core-app/src/renderer/src/components/menu/TouchMenuItem.vue`:82

## What

TouchMenuItem is the app's main sidebar navigation item (used by TouchMenu.vue) and routes on click, but its root is a bare div with no role, no tabindex and no keydown handler.

## Evidence

```
<div
    ref="dom"
    v-wave
    :data-route="props.route"
    class="TouchMenuItem-Container fake-background"
    flex
    items-center
    :class="{ active, disabled }"
    @click="handleClick"
  >
```

A keyboard-only user cannot reach or activate any top-level navigation entry — the divs are not in the tab order and Enter/Space do nothing, so the entire app becomes unnavigable without a mouse. A screen-reader user gets no link/button role and no indication that the active item is current, since 'active' is conveyed only by a CSS class.

## Suggested direction

Render as <button> or add role="link", tabindex="0", @keydown.enter/@keydown.space and :aria-current for the active item.

---
<sub>audit-domain: a11y · confidence: confirmed</sub>

TalexDreamSoul:
Fixes #505.

## The defect

`TouchMenuItem` is the primary sidebar navigation item and routes on click, but its root was a bare `div` — no role, no tabindex, no key handler. A keyboard-only user could not reach or activate any top-level destination.

## Why it stays a div

`role="link"` rather than a promoted `<button>`: it **routes**, so link announces what it actually does where a button would promise an action. The cost is that role, tabindex and key handling all have to be supplied by hand — and `Space` is handled alongside `Enter`, because a non-native control gets neither for free.

## Beyond the reported scope, deliberately

| addition | why |
|---|---|
| `aria-current="page"` on the active item | otherwise a screen-reader user can reach every destination but cannot tell which one they are on |
| `aria-disabled` + `tabindex="-1"` when disabled | `handleClick` already bailed on `disabled`; without these the item was still a focus stop that silently did nothing |
| `:focus-visible` outline | **the important one** — without it the item takes focus and shows nothing, which is worse than being unreachable: the user is moving through invisible stops |

Making it focusable without a visible focus ring would have satisfied the issue's literal text while leaving it unusable, so the outline is not optional here.

## Verified by mutation

```
unfixed component -> 4 failed | 1 passed
this change       -> 5 passed
```

The one that passes in both states is deliberate: *aria-current absent on an inactive item*. It guards a fix that marks **every** item as the current page, which would be worse than marking none.

## Gates

- `npm run typecheck:web`: exit 0
- New suite: 5 passing

### DIFF
--- a/apps/core-app/src/renderer/src/components/menu/TouchMenuItem.a11y.test.ts
+++ b/apps/core-app/src/renderer/src/components/menu/TouchMenuItem.a11y.test.ts
@@ -1,0 +1,74 @@
+// @vitest-environment jsdom
+import { mount } from '@vue/test-utils'
+import { describe, expect, it, vi } from 'vitest'
+import TouchMenuItem from './TouchMenuItem.vue'
+
+/**
+ * TouchMenuItem is the sidebar's primary navigation item and routes on click, but its root was a
+ * bare div with no role, no tabindex and no key handler -- reachable only by mouse (#505).
+ */
+
+const push = vi.fn()
+
+vi.mock('vue-router', () => ({
+  useRouter: () => ({ push, afterEach: () => () => {} }),
+  useRoute: () => ({ path: '/elsewhere', matched: [] })
+}))
+
+function mountItem(props: Record<string, unknown> = {}) {
+  return mount(TouchMenuItem, {
+    props: { name: 'Home', route: '/home', ...props },
+    global: {
+      directives: { wave: {} },
+      provide: { changePointer: () => {} }
+    }
+  })
+}
+
+describe('TouchMenuItem keyboard reachability', () => {
+  it('is focusable and announces itself as a link', () => {
+    const root = mountItem().get('.TouchMenuItem-Container')
+
+    expect(root.attributes('role')).toBe('link')
+    expect(root.attributes('tabindex')).toBe('0')
+  })
+
+  it('routes on Enter and on Space, not only on click', async () => {
+    const wrapper = mountItem()
+    const root = wrapper.get('.TouchMenuItem-Container')
+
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
Rebased onto the current base and fixed the two things that had kept this red since 2026-08-08. The a11y change itself is untouched.

1. **`Typecheck (workspace)`** — `mountItem` omitted `icon`, which is a required prop, so tsc failed with TS2322 on the mount options while the suite passed. That failure was only ever visible to tsc, which is why it survived: the tests were green the whole time.
2. **`PR Quality`** — `handleClick` is annotated `MouseEvent`, which this branch's own `@keydown.enter` / `@keydown.space` bindings violate. Widened to `MouseEvent | KeyboardEvent`, which is also what the `active` emit now carries.

Verified locally: 5 tests pass, `typecheck:web` = 9, unchanged from this base's baseline and with nothing in the touched files. ESLint clean.

One note for the record, since I re-derived it while picking up #505 independently: this component is **not** currently mounted. `<TouchMenu` appears in zero files outside `components/menu/` — against 7 for `<FlatInput` under the identical query, as a control — and the component header already directs new code to the TuffEx primitives. The fix is still right (a component this shape gets revived, and reviving it with the defect is how it comes back), but the user-facing severity is lower than the issue states.

PR #1290, opened 2026-08-08
fix(i18n): define the five account-dialog keys that only had Chinese defaults

### LINKED ISSUE #491: [audit/i18n] User profile dialog falls back to hardcoded Chinese for 5 keys missing from both locales

**Severity:** high  ·  **Category:** i18n  ·  **Domain:** i18n

**Location:** `apps/core-app/src/renderer/src/components/base/TuffUserInfo.vue`:228

## What

Five t(key, default) calls in the account dialog supply Chinese literals as the default message, and none of the five keys exist in en-US.json or zh-CN.json, so the Chinese default is what actually renders in every locale.

## Evidence

```
<template #name>{{ t('userProfile.overview', '概览') }}</template> — missing from BOTH locales: userProfile.overview ('概览'), userProfile.security ('安全与登录'), userProfile.securityActions ('登录与安全选项'), settingUser.syncSecurity ('同步安全'), settingUser.stepUp ('二次验证')
```

An English-locale user opens their account/profile dialog and sees Chinese section names ('概览', '安全与登录') mixed into an otherwise English UI. Because the fallback masks the missing key, no missing-translation warning surfaces.

## Suggested direction

Add the 5 keys to both locale files with proper English and Chinese values.

---

TalexDreamSoul:
Fixes #491.

## The defect

`TuffUserInfo.vue` calls `t(key, default)` with a **Chinese literal** as the fallback for five keys that exist in neither locale. Since the key never resolves, the Chinese default is what renders in *every* language — an English user saw 概览 and 安全与登录 in the account dialog.

## The Chinese side is not new translation

The zh values are **byte-identical** to the literals the component already rendered, verified after writing:

```
userProfile.overview: '概览' == default '概览' -> True
userProfile.security: '安全与登录' == default '安全与登录' -> True
...
```

So zh users see no change at all. Only the English side is new:

| key | en-US |
|---|---|
| `userProfile.overview` | Overview |
| `userProfile.security` | Security & sign-in |
| `userProfile.securityActions` | Sign-in and security options |
| `settingUser.stepUp` | Two-step verification |
| `settingUser.syncSecurity` | Sync security |

Two of the five live under `settingUser`, not `userProfile`, despite the issue title — both namespaces already existed, so each key sits with its siblings.

## Scope checked, not taken on trust

All **37** `t(key, default)` calls in the file were resolved against both locales:

- 32 already resolve in both
- **5** missing from both — exactly the ones reported
- 0 missing from only one locale
- 0 English-literal defaults left unresolved

So #491's count is right. Worth stating plainly, because scope accuracy has been the weak point of this batch — this is the second issue in a row whose boundary held.

## Measured against origin as a control

```
origin  : 39 keys referenced, 5 unresolved
this fix: 39 keys referenced, 0 unresolved
```

Without the before-measurement, "0 unresolved" could equally mean the scan matched nothing.

## Edited as text, not reserialised

Both locale files carry uncommitted changes in the working tree; a `json.dump` rewrite would reformat thousands of lines and collide with that work. Their pending edits are one line each and touch neither namespace, so this insertion cannot conflict — checked before editing. The result parses and every added key resolves.

### DIFF
--- a/apps/core-app/src/renderer/src/modules/lang/en-US.json
+++ b/apps/core-app/src/renderer/src/modules/lang/en-US.json
@@ -661,7 +661,9 @@
       "authError": "Authentication failed. Please try again.",
       "unknownError": "Unknown authentication error. Please try again."
     },
-    "runtimeApiServer": "Runtime API server"
+    "runtimeApiServer": "Runtime API server",
+    "stepUp": "Two-step verification",
+    "syncSecurity": "Sync security"
   },
   "creditsSummary": {
     "title": "AI Credits",
@@ -741,7 +743,10 @@
     "teamRole": "Role",
     "manageTeam": "Manage Team",
     "upgradeTeam": "Upgrade to Team",
-    "personalTeam": "Personal Team"
+    "personalTeam": "Personal Team",
+    "overview": "Overview",
+    "security": "Security & sign-in",
+    "securityActions": "Sign-in and security options"
   },
   "layout": {
     "back": "Back"
--- a/apps/core-app/src/renderer/src/modules/lang/zh-CN.json
+++ b/apps/core-app/src/renderer/src/modules/lang/zh-CN.json
@@ -661,7 +661,9 @@
       "authError": "认证失败，请重试。",
       "unknownError": "发生未知认证错误，请重试。"
     },
-    "runtimeApiServer": "运行时 API 服务器"
+    "runtimeApiServer": "运行时 API 服务器",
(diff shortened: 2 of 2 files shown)


TalexDreamSoul:
Superseded by #1412, already merged for #491 (closed). Same five keys, same Chinese values taken from the literals themselves. #1412 additionally removes the `t(key, default)` fallbacks — the issue's own point was that the default masks the missing key, so leaving them in place would preserve the mechanism that hid it.

Issue #503, opened 2026-08-06
[audit/i18n] Eight settings.intelligence keys exist only in en-US.json, leaving zh-CN incomplete


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** low  ·  **Category:** i18n  ·  **Domain:** i18n

**Location:** `apps/core-app/src/renderer/src/modules/lang/zh-CN.json`:1

## What

en-US.json defines 4463 leaf keys and zh-CN.json 4455. Beyond the four capability-test keys reported separately, eight settings.intelligence keys are present only in English, so any future consumer of them will fall back to English for Chinese users.

## Evidence

```
Present in en-US.json but missing from zh-CN.json: settings.intelligence.addBinding, settings.intelligence.removeBinding, settings.intelligence.modelOrder, settings.intelligence.noModelsConfigured, settings.intelligence.dragToReorder, settings.intelligence.addFirstModel, settings.intelligence.providerSelection, settings.intelligence.modelSelection
```

## Why it matters / failure scenario

None of the eight is currently referenced from renderer code, so no user-visible breakage today; the drift means a Chinese user hits untranslated English the moment the model-binding UI starts using them, and it masks real drift in automated key-parity checks.

## Suggested direction

Either add Chinese translations for the eight keys or delete them if the model-binding UI was dropped; add a CI key-parity check between the two locale files.

---
<sub>audit-domain: i18n · confidence: confirmed</sub>

TalexDreamSoul:
## 核实结论:缺口是 12 不是 8,**而且这 12 条全是死键——按正文补翻译等于翻译 12 条没人用的字符串**

### 1. 实际差异是 12 条,方向单一

用叶子键全量对比(不是抽样):

```
en 叶子 4544   zh 叶子 4532
仅存在于 en:12    仅存在于 zh:0
```

正文说的「八条」加上它提到的「另行报告的四条 capability-test 键」正好是 12——但那四条**在这里同样还缺**,所以完整修复要覆盖 12 条,不是 8 条。

### 2. 这 12 条**全部零引用**

```
settings.intelligence.{addBinding, addFirstModel, dragToReorder, modelOrder,
                       modelSelection, noModelsConfigured, providerSelection, removeBinding}
settings.settingAISDK.{promptVariablesLabel, promptVariablesPlaceholder,
                       userMessageLabel, userMessagePlaceholder}
```

[...]

PR #1291, opened 2026-08-08
fix(i18n): define the eight prompt-selector keys missing from both locales

### LINKED ISSUE #487: [audit/i18n] IntelligencePromptSelector references 8 intelligence.* keys missing from both locales

**Severity:** high  ·  **Category:** i18n  ·  **Domain:** i18n

**Location:** `apps/core-app/src/renderer/src/components/intelligence/config/IntelligencePromptSelector.vue`:109

## What

The prompt-selection dropdown labels every group header, empty state and action with keys that resolve in neither en-US.json nor zh-CN.json. An intelligence root namespace does exist, so these are genuinely absent rather than a namespace-root problem.

## Evidence

```
:aria-label="t('intelligence.managePrompts')" (line 109) — missing from BOTH locales: intelligence.managePrompts (109), intelligence.selectPrompt (121), intelligence.builtinPrompts (133), intelligence.builtin (149), intelligence.customPrompts (158), intelligence.custom (174), intelligence.createNewPrompt (180), intelligence.promptSelectedHint (204)
```

Any user picking a prompt for an AI capability sees the selector's placeholder, both option-group headings ('intelligence.builtinPrompts' / 'intelligence.customPrompts') and the create action rendered as dot-paths, making it impossible to distinguish built-in from custom prompts. Line 109 feeds aria-label directly, so a screen-reader user hears the manage-prompts control announced as the literal string 'intelligence.managePrompts'.

## Suggested direction

Add the 8 intelligence.* keys to both locale files.

---

TalexDreamSoul:
Fixes #487.

## The defect

`IntelligencePromptSelector.vue` labels every group header, empty state and action with `intelligence.*` keys present in neither locale, so the dropdown rendered raw key strings.

One of them is an `aria-label` (`:109`), so the manage button was **unnamed for a screen reader** — an a11y consequence the issue does not mention.

## Correcting my earlier comment on this issue

I previously wrote that the file references **ten** bare keys against the issue's eight. It references **eleven**, of which exactly **eight** are unresolved — I had compared a *reference* count against a *missing* count. `intelligence.instructions` and `intelligence.instructionsPlaceholder` both resolve fine.

The issue's figure of eight was correct.

## Inserted by line number, not by anchor text

`"compatible": "Compatible"` occurs **twice** in `en-US.json`, at two different nesting depths. A text replace would have been a coin flip about which object received the keys.

So insertion is by line number, with the target line asserted to be the intended one before writing, and every key read back through the parsed JSON afterwards.

## Measured against origin as a control

```
origin  : 11 referenced, 8 unresolved
this fix: 11 referenced, 0 unresolved
```

Locale parity is **unchanged** at 12 en-only leaves — all eight keys went into both files. Those 12 are the dead keys tracked in #503, not a regression from this change.

## Why this could be done now when I declined earlier

I left this alone in an earlier pass because both locale files carry uncommitted changes. Having since checked what those changes actually are — a single `home.toolInterrupted` key in each — they touch neither `intelligence` nor the insertion point, so there is nothing to collide with.

The larger locale work in #503 stays untouched, because that one is a *deletion* of ~116 dead keys and genuinely needs a decision first.

### DIFF
--- a/apps/core-app/src/renderer/src/modules/lang/en-US.json
+++ b/apps/core-app/src/renderer/src/modules/lang/en-US.json
@@ -4694,6 +4694,14 @@
     "providerName": "Provider Name",
     "providerNamePlaceholder": "Enter provider name",
     "customCompatible": "Custom (OpenAI Compatible)",
+    "managePrompts": "Manage prompts",
+    "selectPrompt": "Select a prompt",
+    "builtinPrompts": "Built-in prompts",
+    "builtin": "Built-in",
+    "customPrompts": "Custom prompts",
+    "custom": "Custom",
+    "createNewPrompt": "Create new prompt",
+    "promptSelectedHint": "This prompt will be used for the current capability.",
     "compatible": "Compatible",
     "search": {
       "placeholder": "Search providers..."
--- a/apps/core-app/src/renderer/src/modules/lang/zh-CN.json
+++ b/apps/core-app/src/renderer/src/modules/lang/zh-CN.json
@@ -4646,6 +4646,14 @@
     "providerName": "提供商名称",
     "providerNamePlaceholder": "输入提供商名称",
     "customCompatible": "自定义（OpenAI 兼容）",
+    "managePrompts": "管理提示词",
+    "selectPrompt": "选择提示词",
+    "builtinPrompts": "内置提示词",
+    "builtin": "内置",
+    "customPrompts": "自定义提示词",
+    "custom": "自定义",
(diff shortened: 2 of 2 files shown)


TalexDreamSoul:
Superseded by #1417, already merged for #487 (closed). Same eight keys. One wording difference worth noting: #1417 words `promptSelectedHint` from what the component does — `watch(selectedPrompt, …)` copies the prompt's content into `customInstructions` and disables the textarea — rather than from the key's name.

PR #1294, opened 2026-08-08
ref(core-app): drop the superseded update-asset ranking helpers

### LINKED ISSUE #529: [audit/tech-debt] update-asset-utils.ts: 8 of 9 exported helpers unused, including the only update-asset scoring logic

**Severity:** medium  ·  **Category:** dead-code  ·  **Domain:** architecture

**Location:** `apps/core-app/src/main/modules/update/update-asset-utils.ts`:8

## What

This 107-line module exports 9 functions. Only normalizeUpdateAssetKey is imported anywhere (update-system.ts:38). The other 8 — including calculateUpdateAssetScore, the app's sole implementation of update-installer ranking — have no callers and no tests.

## Evidence

```
export function calculateUpdateAssetScore(
  filename: string,
  asset: UpdateAssetScoreInput,
  platform: string
): number {
  let score = 0

  if (asset.component === 'core') {
    score += 200
  }
  ...
  score += getInstallerExtensionScore(filename, platform)

  return score
}

// Sole external import of this module:
// apps/core-app/src/main/modules/update/update-system.ts:38
import { normalizeUpdateAssetKey } from './update-asset-utils'
```

A grep for 'score' across the whole update module matches only inside update-asset-utils.ts itself, confirming no other ranking implementation exists — so the updater does not score candidate assets at all. When a user reports that the updater downloaded the wrong artifact on a platform that ships several installers, a developer finds calculateUpdateAssetScore, reads its priority table (core +200, tuff +20, latest-release +10), fixes the weights, and ships a change that provably cannot affect behaviour.

## Suggested direction

Delete the 8 unreferenced helpers, or call calculateUpdateAssetScore from the real asset-selection path in update-system.ts.

---

TalexDreamSoul:
Fixes #529. 107 lines down to 13.

## The finding holds

Nine exports, one caller. `normalizeUpdateAssetKey` is used by `update-system.ts`; the other eight — `calculateUpdateAssetScore`, `getInstallerExtensionScore` and six asset-classification predicates — had no callers and no tests.

Each was searched across `apps`, `packages`, `plugins` and `scripts`, with `normalizeUpdateAssetKey` as the positive control (1 file). `UpdateAssetScoreInput` was exported but referenced nowhere.

## The framing needed checking before deleting

The issue calls `calculateUpdateAssetScore` *"the app's sole implementation of update-installer ranking"* — which reads as though deleting it removes behaviour users depend on. It does not, and confirming that was the point of this pass.

Installers are resolved **by exact name against the release manifest**:

```
update-system.ts  resolveAssetByName        -> normalizeUpdateAssetKey equality
update-system.ts  manifest.artifacts.find   -> component match
update-system.ts  fetchReleaseManifest      -> returns null when no manifest asset exists
```

That last line matters: there is **no heuristic fallback**. A release without a manifest yields nothing rather than a best-guess installer.

So the scoring was the older manifest-less approach, fully superseded — not ranking that should be happening and silently is not. There is no code path that reaches it.

## Why deleting beats leaving it

Left in place, 94 lines of filename heuristics read as live installer-selection policy while having no bearing on which installer users actually receive. The next person tracing an update bug would reasonably start there and lose time.

The surviving function carries a comment recording what was removed and why, so that conclusion does not have to be re-derived.

## Gates

- update suites: **14 files, 102 tests passing**

### DIFF
--- a/apps/core-app/src/main/modules/update/update-asset-utils.ts
+++ b/apps/core-app/src/main/modules/update/update-asset-utils.ts
@@ -1,107 +1,13 @@
-import type { UpdateArtifactComponent } from '@talex-touch/utils'
-import { UPDATE_RELEASE_MANIFEST_NAME } from '@talex-touch/utils'
-
-export interface UpdateAssetScoreInput {
-  component?: UpdateArtifactComponent
-}
-
-export function calculateUpdateAssetScore(
-  filename: string,
-  asset: UpdateAssetScoreInput,
-  platform: string
-): number {
-  let score = 0
-
-  if (asset.component === 'core') {
-    score += 200
-  }
-
-  if (filename.includes('tuff')) {
-    score += 20
-  }
-
-  if (filename.includes('latest-release')) {
-    score += 10
-  }
-
-  score += getInstallerExtensionScore(filename, platform)
-
-  return score
-}
-
-export function getInstallerExtensionScore(filename: string, platform: string): number {
-  if (platform === 'darwin') {
-    if (filename.endsWith('.app.zip')) return 180
-    if (filename.endsWith('.dmg')) return 140
-    if (filename.endsWith('.pkg')) return 130
-    if (filename.endsWith('.zip')) return 90
-    return 0
-  }
-
-  if (platform === 'win32') {
-    if (filename.endsWith('.exe')) return 140
-    if (filename.endsWith('.msi')) return 130
-    if (filename.endsWith('.zip')) return 90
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Superseded by #1296, which merged for #529.

Compared before closing rather than assuming: both delete the same eight orphaned helpers, and #1296 additionally adds `update-asset-utils.test.ts` for the one survivor (`normalizeUpdateAssetKey`), mutation-checked against three mutants. Nothing here is lost.

PR #1298, opened 2026-08-08
ref(nexus): share the runtime-server migration policy between main and renderer

### LINKED ISSUE #522: [audit/tech-debt] Nexus runtime-server resolution duplicated across main and renderer with drifted APIs

**Severity:** medium  ·  **Category:** duplication  ·  **Domain:** architecture

**Location:** `apps/core-app/src/main/modules/nexus/runtime-base.ts`:19

## What

The policy that decides which Nexus backend the app talks to (including the legacy `authServer` -> `runtimeServer` migration and the coercion of unknown values) is implemented twice, once for the main process and once for the renderer. The two copies already have different signatures and different function sets.

## Evidence

```
// apps/core-app/src/main/modules/nexus/runtime-base.ts:19
export function ensureRuntimeServerSettings(appSettings: AppSetting): TuffNexusRuntimeServer {
  const dev = (appSettings.dev ?? {}) as LegacyDevSettings
  const current = dev.runtimeServer
  const next = current ?? dev.authServer ?? 'production'
  dev.runtimeServer = normalizeRuntimeServer(next)
  delete dev.authServer

// apps/core-app/src/renderer/src/modules/nexus/runtime-base.ts:18
export function ensureRuntimeServerSettings(): TuffNexusRuntimeServer {
  if (!appSetting.dev) {
    appSetting.dev = { autoCloseDev: true, runtimeServer: 'production', developerMode: false }
  }
  const dev = appSetting.dev as LegacyDevSettings
```

Adding a third runtime target (e.g. 'staging') requires editing `normalizeRuntimeServer` in both files. Because both copies coerce anything that is not 'local' to 'production' (`return value === 'local' ? 'local' : 'production'`), missing one file does not fail a build or a test — the un-updated process silently keeps pointing at the production Nexus backend while the other process points at staging, so main and renderer disagree about which server the session is authenticated against.

## Suggested direction

Move the normalization and legacy migration into a single shared module (packages/utils/env already owns resolveTuffNexusBaseUrl) and have each process supply only its settings accessor.

---

TalexDreamSoul:
Fixes #522.

## The defect

The policy deciding which Nexus backend the app talks to — *prefer `runtimeServer`, fall back to the retired `authServer`, default to production, then drop `authServer`* — was implemented twice, once per process. The copies had already drifted: the renderer also exported `setRuntimeServerMode`.

**The drift is the part that matters.** Two implementations of "which backend do we trust" can disagree about what a corrupted settings value means, and the main and renderer processes would then talk to different servers with nothing reporting it.

## Split by what actually differs

| | where it lives now | why |
|---|---|---|
| migration policy, `normalize` | `@talex-touch/utils/env` | identical in both, and sits next to `resolveTuffNexusBaseUrl` / `NEXUS_BASE_URL` which they already shared |
| storage plumbing | per-process | genuinely different — main reads/writes the config file and re-saves only when the migration changed something; the renderer mutates the reactive store and seeds `dev` when absent |
| `setRuntimeServerMode` | renderer only | nothing in main sets it; unifying it would invent a capability |

Deduplicating the plumbing too would have forced one process to pretend to be the other. Only the policy was actually duplicated.

## This is a move, not a rewrite

The line

```ts
const next = current ?? dev.authServer ?? 'production'
```

was present **verbatim in both copies** at origin — checked before extracting, so the shared function reproduces both originals rather than reinterpreting them.

## The behaviour now has tests, which it did not before

12 cases: `authServer` adoption, `runtimeServer` precedence, the production default, that `authServer` is always dropped (including when it was the value adopted), and that unrelated `dev` settings survive.

Plus the one that matters most:

```ts
for (const value of [undefined, null, 0, 1, true, {}, []])
  expect(normalizeTuffNexusRuntimeServer(value)).toBe('production')
```

That guards the failure this policy exists to prevent — a corrupted settings file silently selecting a non-production backend. With two copies, that guarantee held in whichever copy you happened to read.

## Gates

- `npm run typecheck` (node + web): exit 0
- `packages/utils`: **135 files, 1084 passing**
- nexus suites: 2 files, 11 passing
- `eslint --max-warnings=0` in both packages: exit 0

### DIFF
--- a/apps/core-app/src/main/modules/nexus/runtime-base.ts
+++ b/apps/core-app/src/main/modules/nexus/runtime-base.ts
@@ -1,29 +1,24 @@
 import type { AppSetting } from '@talex-touch/utils/common/storage/entity/app-settings'
 import { StorageList } from '@talex-touch/utils'
 import {
+  applyTuffNexusRuntimeServerMigration,
   NEXUS_BASE_URL,
   type TuffNexusRuntimeServer,
+  type TuffNexusRuntimeServerSettings,
   resolveTuffNexusBaseUrl
 } from '@talex-touch/utils/env'
 import { getMainConfig, saveMainConfig } from '../storage'
 
-type LegacyDevSettings = AppSetting['dev'] & {
-  authServer?: TuffNexusRuntimeServer
-  runtimeServer?: TuffNexusRuntimeServer
-}
-
-function normalizeRuntimeServer(value: unknown): TuffNexusRuntimeServer {
-  return value === 'local' ? 'local' : 'production'
-}
-
+/**
+ * Only the storage plumbing lives here. The migration policy itself is shared with the renderer
+ * copy of this module via @talex-touch/utils/env -- the two had drifted while claiming to decide
+ * the same thing (#522).
+ */
 export function ensureRuntimeServerSettings(appSettings: AppSetting): TuffNexusRuntimeServer {
-  const dev = (appSettings.dev ?? {}) as LegacyDevSettings
-  const current = dev.runtimeServer
-  const next = current ?? dev.authServer ?? 'production'
-  dev.runtimeServer = normalizeRuntimeServer(next)
-  delete dev.authServer
+  const dev = (appSettings.dev ?? {}) as AppSetting['dev'] & TuffNexusRuntimeServerSettings
+  const runtimeServer = applyTuffNexusRuntimeServerMigration(dev)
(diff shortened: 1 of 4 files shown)


TalexDreamSoul:
Superseded by #1414, already merged for #522 (closed). The shared policy lives in `@talex-touch/utils/env`; each process keeps its own storage access, which was the part that legitimately differed between them.

PR #1302, opened 2026-08-08
fix(core-app): remove the powerMonitor handlers when CommonChannelModule is destroyed

### LINKED ISSUE #532: [audit/tech-debt] CommonChannelModule leaks two powerMonitor listeners that re-arm a global poll after destroy

**Severity:** high  ·  **Category:** lifecycle-leak  ·  **Domain:** architecture

**Location:** `apps/core-app/src/main/channel/common.ts`:914

## What

setupBatteryStatusBroadcast registers powerMonitor 'on-ac' and 'on-battery' handlers without keeping a disposer. onDestroy clears the interval and the transport disposers but never removes them, so the handlers outlive the module and can re-register a task on the global PollingService singleton after teardown.

## Evidence

```
// registration (common.ts:914-923) — no disposer captured
      powerMonitor.on('on-ac', () => {
        stopPolling()
        void broadcastBatteryStatus()
      })

      powerMonitor.on('on-battery', () => {
        startPolling()
        void broadcastBatteryStatus()
      })

// teardown (common.ts:1978-1996) — powerMonitor never touched
  onDestroy(): MaybePromise<void> {
    log.info('CommonChannelModule destroyed')
    if (this.batteryPollTimer) { clearInterval(this.batteryPollTimer); this.batteryPollTimer = undefined }
    for (const dispose of this.transportDisposers) { ... }
    this.transportDisposers = []
    this.transport = null
    this.channel = null
  }
```

Destroy the module (hot-reload in dev, or module teardown on quit) and then unplug the laptop. The surviving 'on-battery' handler runs startPolling(), which calls pollingService.register(BATTERY_POLL_TASK_ID, ...) with a 120s interval and pollingService.start() on the process-wide PollingService.getInstance() (common.ts:141). The only unregister call in the file is inside the local stopPolling closure (line 904), which onDestroy never invokes, so the task runs forever, broadcasting through the `transport` captured in the closure even though onDestroy set this.transport = null. Each init/destroy cycle adds another pair of powerMonitor listeners.

## Suggested direction

Push `() => powerMonitor.off('on-ac', h)` / `off('on-battery', h)` and `stopPolling()` into transportDisposers, exactly as network-module.ts:104 already does for its 'resume' handler.

---

TalexDreamSoul:
Fixes #532.

## The defect

`setupBatteryStatusBroadcast` registered `'on-ac'` and `'on-battery'` on `powerMonitor` without keeping a way to remove them. `powerMonitor` is an **app-lifetime singleton**, so the handlers outlived the module: `onDestroy` cleared the poll interval and the transport disposers, then an `'on-battery'` event arriving afterwards called `startPolling()` again and re-armed a task on the global `PollingService`.

## Follows the shape already in the file

The handlers are captured and removed explicitly, mirroring how `batteryPollTimer` is already handled — a named field cleaned in `onDestroy`, rather than pushed into `transportDisposers`, which is specifically about transport subscriptions.

## Ordering matters, and it is deliberate

Removal runs **before** `clearInterval`. In the other order, an `'on-battery'` event arriving between the two would call `startPolling()` and leave a fresh interval behind — the same leak, just through a smaller window. That is recorded in a comment so the order is not "tidied" later.

## The mock had to be extended too

`vi.mock('electron')` stubbed only `powerMonitor.on`, so calling `off()` would have thrown at teardown. It now stubs both — worth noting because that omission is why nothing caught this.

## Verified by mutation

```
unfixed module -> 2 failed | 26 passed
this change    -> 28 passed
```

The second case is the one that matters:

```ts
for (const [event, handler] of powerMonitor.on.mock.calls) {
  const match = powerMonitor.off.mock.calls.find((call) => call[0] === event)
  expect(match?.[1]).toBe(handler)
}
```

`off()` with a *different* closure is a **silent no-op** — it removes nothing and reports nothing. A test that only checked "off was called twice" would pass against a broken fix, so identity is what gets asserted.

## Gates

- `common.test.ts`: **28 passing**

### DIFF
--- a/apps/core-app/src/main/channel/common.test.ts
+++ b/apps/core-app/src/main/channel/common.test.ts
@@ -160,7 +160,8 @@ vi.mock('electron', () => ({
     }))
   },
   powerMonitor: {
-    on: vi.fn()
+    on: vi.fn(),
+    off: vi.fn()
   },
   shell: {
     openExternal: vi.fn(),
@@ -497,6 +498,8 @@ import { APP_SCHEMA, FILE_SCHEMA } from '../config/default'
 import { CommonChannelModule } from './common'
 
 type CommonChannelModuleTestInstance = {
+  setupBatteryStatusBroadcast: () => void
+  onDestroy: () => unknown
   createSafeOperationHandler: (transport: {
     on: (
       event: unknown,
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
Superseded by #1348, which merged for #532.

Compared before closing. Both capture the powerMonitor handlers and remove them on destroy. #1348 additionally releases the **poll registration** — `onDestroy` never unregistered the `PollingService` task either, so releasing only the listeners still left a global task behind on every normal teardown. That was the half this PR does not cover.

Issue #566, opened 2026-08-06
[audit/compat] Published @talex-touch/utils has main:"index.ts"; Node 24 cannot load it (ERR_UNSUPPORTED_DIR_IMPORT)


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** high  ·  **Category:** compat  ·  **Domain:** compat

**Location:** `packages/utils/package.json`:18

## What

@talex-touch/utils is a public npm package (private:false, v1.0.50) whose entry point is a raw TypeScript file using ESM `export *` with extensionless directory specifiers. Node 24 strips the types, treats the file as ESM, and then cannot resolve the directory re-exports.

## Evidence

```
packages/utils/package.json:18 "main": "index.ts", :19 "module": "./index.ts", :4 "private": false  —  packages/utils/index.ts:2 `export * from './account'` (extensionless directory specifier, no `exports` map and no `types` field in package.json)
```

## Why it matters / failure scenario

Reproduced on Node 24.18.0: `require('@talex-touch/utils')` throws `ERR_UNSUPPORTED_DIR_IMPORT: Directory import '.../packages/utils/account' is not supported resolving ES modules imported from .../packages/utils/index.ts`. Any external plugin developer consuming the published package from plain Node, or from any bundler that does not alias TS sources, cannot import it at all — only the in-repo electron-vite/Vite aliases make it work.

## Suggested direction

Publish a built dist (ESM+CJS+.d.ts) with an `exports` map, or at minimum add explicit file extensions and an `exports` field.

---
<sub>audit-domain: compat · confidence: confirmed</sub>

TalexDreamSoul:
## Reproduced against the published artifact — and it is worse than filed

I installed `@talex-touch/utils@1.0.50` from npm into a clean directory:

```
$ npm view @talex-touch/utils main module types exports
main    = 'index.ts'
module  = './index.ts'
types   = (absent)
exports = (absent)

$ node -e "require('@talex-touch/utils')"
ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING          exit 1

$ node --input-type=module -e "await import('@talex-touch/utils')"
ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING          exit 1
```

**Both CommonJS and ESM entry points fail.** So the package is unloadable from Node by any external consumer, which is who it is published for.

[...]

PR #1305, opened 2026-08-08
build: fail publish validation when a runtime entry is a raw TypeScript file

TalexDreamSoul:
Detects the defect in #566. **Does not fix the packaging** — see the last section.

## The defect

`@talex-touch/utils` is published (`private: false`, v1.0.50) with `main: "index.ts"`, and that file uses extensionless directory specifiers:

```ts
export * from './account'
export * from './analytics'
```

Node strips the types, treats the file as ESM, and cannot resolve the directory re-exports. An external consumer installing from npm gets a package that will not load.

There is also **no build script** in `packages/utils` at all — so this is not a stale `main`, it is the intended shape.

## Nothing detected it

The publish validator checked dependency protocols only. A package could declare a raw `.ts` runtime entry and pass clean.

## Adding the check found a second package #566 does not mention

```
- @talex-touch/utils            main, module -> index.ts
- @talex-touch/tuff-intelligence main, module -> src/index.ts
```

Same problem, same consequence.

## `types` pointing at a .d.ts is not flagged

Deliberate. `types: "dist/index.d.ts"` is correct, and four published packages use it. Without that distinction the check would fire on packages that are fine — which would be worse than no check, since the noise trains people to ignore it.

## This makes `publish:check` red, on purpose

Publishing either package today produces something Node cannot load, so the gate failing *is* the gate working.

Scope of that: `publish:check` is **not wired into CI** (checked), so this does not redden the build. It fails the manual pre-publish gate — which is exactly the moment the failure is worth knowing about, rather than after a broken version is on npm.

## Not fixed here

The packaging itself. Giving `@talex-touch/utils` a build and an `exports` map affects **26 shipped subdirectories** and every subpath consumer (`/env`, `/transport`, `/plugin`, …). That is its own task and #566 stays open for it.

In-repo consumers are unaffected either way: plugins depend via `workspace:^` and bundle through vite, which resolves TypeScript directly. The breakage is specific to external npm consumers using Node.

## Gates

`eslint --max-warnings=0`: exit 0

### DIFF
--- a/scripts/validate-publish-manifests.mjs
+++ b/scripts/validate-publish-manifests.mjs
@@ -161,17 +161,49 @@ function assertPublishConfig(packages) {
   return issues
 }
 
+/**
+ * A published package whose runtime entry is a raw .ts file cannot be loaded by Node: it strips
+ * the types, treats the file as ESM, then fails on the extensionless directory specifiers these
+ * entries use -- ERR_UNSUPPORTED_DIR_IMPORT (#566).
+ *
+ * `types` pointing at a .d.ts is correct and deliberately not flagged; this is only about
+ * main/module, which are runtime entries.
+ */
+function findRawTypeScriptEntryIssues({ manifest, manifestPath, packageInfo }) {
+  const offending = ['main', 'module'].filter((field) => {
+    const value = manifest[field]
+    return typeof value === 'string' && value.endsWith('.ts') && !value.endsWith('.d.ts')
+  })
+
+  if (offending.length === 0)
+    return []
+
+  return [{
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Superseded by #1398, already merged for #566 (closed). #1398 publishes a built ESM entry via `publishConfig` and deliberately adds **no** `exports` map — measured against this repo's 144 deep subpaths, every wildcard form breaks part of them, because subpath patterns resolve to exact files and neither Node nor esbuild falls through a fallback array on a merely-missing target.

Issue #584, opened 2026-08-06
[audit/compat] core-app depends on deprecated xterm@5.3.0 and xterm-addon-fit@0.8.0 (renamed to @xterm/*)


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** medium  ·  **Category:** dependencies  ·  **Domain:** compat

**Location:** `apps/core-app/package.json`:135

## What

The terminal UI uses the retired `xterm` package name, which upstream deprecated in favour of the @xterm scope; the old name receives no fixes.

## Evidence

```
apps/core-app/package.json:135 `"xterm": "^5.3.0",` :136 `"xterm-addon-fit": "^0.8.0"`. pnpm-lock.yaml:13819 `xterm@5.3.0: ... deprecated: This package is now deprecated. Move to @xterm/xterm instead.` and pnpm-lock.yaml:13813 `xterm-addon-fit@0.8.0: ... deprecated: This package is now deprecated. Move to @xterm/addon-fit instead.` No @xterm scope is installed anywhere in the workspace.
```

## Why it matters / failure scenario

xterm 5.3.0 is frozen at the 2023 release. The terminal module (apps/core-app/src/main/modules/terminal) is stuck without the @xterm/xterm 5.5+ fixes for Chromium canvas/WebGL renderer changes, and every install emits a deprecation warning. Any future Electron/Chromium change to canvas text metrics cannot be fixed by upgrading, only by migrating the package name.

## Suggested direction

Migrate to @xterm/xterm and @xterm/addon-fit (import path and addon class names change).

---
<sub>audit-domain: compat · confidence: confirmed</sub>

TalexDreamSoul:
Fixed in #1324.

**Went to `^5.5.0` rather than the current 6.0.0**, deliberately. 5.5 is the direct rename the deprecation notice asks for and carries exactly the renderer fixes you name, which keeps this a rename instead of a major upgrade. That matters here because these two components have **no test coverage** and cannot be exercised outside a running Electron window — the smaller delta is worth more than the newer version. 6.0.0 is API-identical for everything used here (I checked both), so that step stays available and separate.

[...]

PR #1310, opened 2026-08-08
build(touch-music): drop the unused xterm dependencies

### LINKED ISSUE #585: [audit/compat] plugins/touch-music also pins deprecated xterm@5.3.0 / xterm-addon-fit@0.8.0

**Severity:** low  ·  **Category:** dependencies  ·  **Domain:** compat

**Location:** `plugins/touch-music/package.json`:26

## What

The music plugin carries the same retired xterm packages as core-app, and it does not appear to render a terminal at all.

## Evidence

```
plugins/touch-music/package.json:26 `"xterm": "^5.3.0",` :27 `"xterm-addon-fit": "^0.8.0"`, both flagged in pnpm-lock.yaml:13819/13813 as `deprecated: This package is now deprecated. Move to @xterm/xterm instead.`
```

The plugin bundles a deprecated, unmaintained terminal emulator (~300KB) into its Surface build. Because the dependency is unused by the plugin's feature set, it is pure deprecated surface area that will keep failing dependency audits and blocks the @xterm migration from being completed repo-wide.

## Suggested direction

Remove xterm/xterm-addon-fit from touch-music if unused, otherwise migrate with core-app.

---

TalexDreamSoul:
Fixes #585 — but not the way it is framed.

## Not the same defect as #584

#585 says touch-music *"also pins deprecated xterm"*, grouping it with #584. They are different problems:

| | uses xterm? | correct fix |
|---|---|---|
| `apps/core-app` (#584) | **yes** — `InteractiveTerminal.vue`, `LogTerminal.vue` | migrate to the `@xterm` scope |
| `plugins/touch-music` (#585) | **no** — zero references | remove |

Migrating touch-music to `@xterm/*` would have kept an unused dependency and made it look deliberate — a music plugin carrying a terminal emulator, now with a current version number so nothing flags it again.

## Verified before removing

Every reference to `xterm` in `plugins/touch-music`, any syntax, was the two manifest lines removed here. No `peerDependencies` declare them either.

Checked with a **positive control**, so that "no references" could not simply mean a search that finds nothing:

```
howler → plugins/touch-music/src/modules/entity/song-manager.ts
         plugins/touch-music/src/modules/entity/song-resolver.ts
xterm  → (nothing outside package.json)
```

## Gates

- plugin `eslint --max-warnings=0`: exit 0
- manifest diff: 1 insertion / 3 deletions, 12 dependencies remain

### DIFF
--- a/plugins/touch-music/package.json
+++ b/plugins/touch-music/package.json
@@ -22,9 +22,7 @@
     "rgbaster": "^2.1.1",
     "typescript": "^5.9.3",
     "vue": "^3.5.39",
-    "wavesurfer.js": "^7.12.10",
-    "xterm": "^5.3.0",
-    "xterm-addon-fit": "^0.8.0"
+    "wavesurfer.js": "^7.12.10"
   },
   "devDependencies": {
     "@vitejs/plugin-vue": "^6.0.7",


TalexDreamSoul:
Superseded by #1320, which merged for #585.

Compared before closing. Same two dependencies removed. #1320 also updates `pnpm-lock.yaml` (needed for `--frozen-lockfile`) and records the finding that the removal changes nothing about the bundle — the build output is byte-identical, which disproves the "~300KB bundled" claim in the issue.

PR #1313, opened 2026-08-08
docs: describe the real module startup shape in CLAUDE.md

### LINKED ISSUE #611: [audit/documentation] CLAUDE.md module loading order lists 15 modules; startup loads 39 in a different order

**Severity:** high  ·  **Category:** doc-drift  ·  **Domain:** documentation

**Location:** `CLAUDE.md`:74

## What

The documented startup sequence is missing 24 modules, names one module that does not exist, and gets the relative order of the ones it does list wrong.

## Evidence

```
CLAUDE.md:74: '1. DatabaseModule → 2. StorageModule → 3. ShortcutModule → 4. ExtensionLoaderModule → 5. CommonChannelModule → 6. PluginModule → 7. PluginLogModule → 8. CoreBoxModule → 9. TrayHolderModule → 10. AddonOpenerModule → 11. ClipboardModule → 12. TuffDashboardModule → 13. FileSystemWatcher → 14. FileProtocolModule → 15. TerminalModule'. apps/core-app/src/main/index.ts:171-210 defines foregroundModulesToLoad with 37 entries (databaseModule, conversationModule, toolGatewayModule, storageModule, fileProtocolModule, shortcutModule, commonChannelModule, trayManagerModule, ... downloadCenterModule) and index.ts:212 deferredModulesToLoad = [extensionLoaderModule, FileSystemWatcher]. There is no 'TrayHolderModule' — it is trayManagerModule and loads 8th, not 9th; fileProtocolModule is 5th, not 14th; extensionLoaderModule is deferred, not 4th.
```

A developer adding a module that must run after permissions/auth uses the documented 15-step order to pick an insertion point, lands it before permissionModule/authModule (index.ts:184,196) which the doc never mentions, and gets an ordering bug that only reproduces at startup.

## Suggested direction

Either regenerate the list from index.ts:171-212 or replace it with a pointer to that array as the single source of truth.

---

TalexDreamSoul:
Fixes #611.

## Why this one matters more than a typical doc fix

`CLAUDE.md` is the file that steers future work in this repo. A wrong startup order there does not just mislead a reader once — it gets acted on.

## What was wrong

The doc described startup as one flat sequence of 15 modules. It is three groups:

| group | where | contents |
|---|---|---|
| foreground | `index.ts:171` | **38 modules**, loaded in array order before the renderer is released |
| deferred | `index.ts:212` | `extensionLoaderModule`, `FileSystemWatcher` — after the window is up |
| optional | `index.ts:215` | `trayManagerModule` — failure logs and startup continues |

Wrong in every dimension the issue names:

- **15 listed, 40 exist**
- **`TrayHolderModule` does not exist** — 0 occurrences in `src/main`; the real module is `trayManagerModule`
- **`ExtensionLoaderModule` was listed 4th**, ahead of `CommonChannelModule` and `PluginModule`. It is actually *deferred until after the window opens*. That is the most misleading part, because it inverts when extensions become available relative to plugins.

## What replaces it

Both source arrays named with line numbers, the three groups distinguished, and the two ordering constraints the code itself comments on recorded:

```ts
permissionModule,  // Plugin permission management - before plugin module
flowBusModule,     // Flow Transfer system - after plugin module
```

It also tells the reader to check the arrays when order matters, rather than trusting the prose — this list drifted silently once already.

## Verified, not transcribed

The 38 names and their order were extracted from `index.ts` and compared against what I wrote:

```
real foreground modules : 38
documented in the chain : 38
order matches exactly   : True
```

Copying a list of 38 names by hand is exactly where an off-by-one slips in, so it is checked mechanically rather than by eye.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -70,8 +70,33 @@ This is a pnpm workspace monorepo with the main application in `apps/core-app/`,
 - **OCR** ([modules/ocr/](apps/core-app/src/main/modules/ocr/)): Native OCR via @talex-touch/tuff-native (Apple Vision / Windows OCR) with AI-provider fallback
 - **Clipboard** ([modules/clipboard/](apps/core-app/src/main/modules/clipboard/)): System clipboard operations
 
-**Module Loading Order** (sequential after Electron ready):
-1. DatabaseModule → 2. StorageModule → 3. ShortcutModule → 4. ExtensionLoaderModule → 5. CommonChannelModule → 6. PluginModule → 7. PluginLogModule → 8. CoreBoxModule → 9. TrayHolderModule → 10. AddonOpenerModule → 11. ClipboardModule → 12. TuffDashboardModule → 13. FileSystemWatcher → 14. FileProtocolModule → 15. TerminalModule
+**Module Loading Order:** startup is not one flat sequence. `main/index.ts` loads three groups
+through `loadStartupModules` (`main/core/startup-module-loader.ts`):
+
+- **Foreground** (`foregroundModulesToLoad`, `index.ts:171`) — 38 modules loaded in array order
+  before the renderer is released. The order encodes real dependencies, and two of them carry a
+  comment saying so: `permissionModule` must precede `pluginModule`, and `flowBusModule` must
+  follow it.
+
+  databaseModule → conversationModule → toolGatewayModule → storageModule → fileProtocolModule →
+  shortcutModule → commonChannelModule → trayManagerModule → nativeCapabilitiesModule →
+  networkModule → catalogModule → analyticsModule → platformPermissionModule → permissionModule →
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Superseded by #1400, already merged for #611 (which is closed).

The two overlap almost entirely. Where they differ is the enumerated list: this PR spells out all 38 foreground modules, #1400 points at the array instead. I went with the pointer because a 40-item list in prose is what produced the defect being fixed — the previous one drifted 24 modules out of date without anyone noticing — and because the counts are now guarded by `startup-order-doc.test.ts`, which fails the moment a module is added.

One detail from here that #1400 does not have and is worth keeping: the loader itself, `loadStartupModules` in `main/core/startup-module-loader.ts`. Happy to add that line if you want it.

PR #1314, opened 2026-08-08
docs: correct the channel API block and mark it as the legacy surface

### LINKED ISSUE #612: [audit/documentation] CLAUDE.md channel API block mixes main-process and renderer signatures

**Severity:** high  ·  **Category:** correctness  ·  **Domain:** documentation

**Location:** `CLAUDE.md`:248

## What

The six "Key APIs" are presented as one surface, but they come from two different classes with incompatible signatures, and sendTo as written matches neither.

## Evidence

```
CLAUDE.md:245-254 lists, under one "Key APIs" block: 'regChannel(type: ChannelType, eventName: string, callback)', 'send(eventName: string, arg?: any)', 'sendTo(window: BrowserWindow, eventName: string, arg?: any)', 'sendPlugin(pluginName, eventName, arg?)', 'requestKey(name)', 'revokeKey(key)'. Main process (apps/core-app/src/main/core/channel-core.ts): regChannel matches (:570), but send is 'send(type: ChannelType, eventName: string, arg: unknown)' (:800) and sendTo is 'sendTo(win, type: ChannelType, eventName, arg)' (:772-777) — both require a ChannelType the doc omits. Renderer (apps/core-app/src/renderer/src/modules/channel/channel-core.ts): send matches (:205) but regChannel takes no ChannelType (:177), and the class has no sendTo, sendPlugin, requestKey or revokeKey at all. So the documented sendTo(window, eventName, arg) signature exists on neither side.
```

A main-process developer copies `channel.send('my-event', payload)` from the doc; TypeScript rejects it because the first argument must be a ChannelType. A renderer developer copies `regChannel(ChannelType.MAIN, 'evt', cb)` and passes the ChannelType where the event name is expected, silently registering a handler under the wrong key that never fires.

## Suggested direction

Split the block into main-process and renderer subsections with each class's real signatures, and drop the sendTo form that matches neither.

---

TalexDreamSoul:
Fixes #612.

## Three of six signatures were wrong

| API | documented | actual |
|---|---|---|
| `send` | `(eventName, arg?)` | `(type: ChannelType, eventName, arg)` |
| `sendTo` | `(window, eventName, arg?)` | `(win, type: ChannelType, eventName, arg)` |
| `requestKey` | `(name)` | `(name, activation?)` |

`regChannel`, `sendPlugin` and `revokeKey` were correct.

## The issue's explanation is not what is there

> *"they come from two different classes with incompatible signatures"*

`channel-core.ts` declares **one** class, `TouchChannel`, and all six are its methods. So the two-class framing does not hold — but the drift is real, and it has a more useful cause.

## Nothing calls them

Outside `channel-core.ts` and its own tests:

```
.regChannel(  → 0 hits in apps/core-app/src/main
.sendPlugin(  → 0 hits
.requestKey(  → only via plugin-channel-key-registry
```

What main-process modules actually call is `transport.sendTo(webContents, event, payload)` — the **TuffTransport SDK**, a different method on a different object, taking a typed event instead of a `ChannelType` plus a string.

That is why the signatures drifted: no caller would have broken.

Checked with a positive control, since "0 hits" and "a search that finds nothing" look identical — the same scan resolves `.sendTo(` to three real call sites, all of them on `transport`.

## So the fix is more than corrected parameters

The block now names the class and file it describes, and says plainly that this is the **legacy** surface with TuffTransport as the current one.

Documenting a retired API as *the* channel system is worse than documenting it with wrong parameters: wrong parameters fail loudly the first time someone compiles, whereas a reader who believes this block writes new code against the wrong layer and it works.

## Verified mechanically

Each documented signature was regex-matched against `channel-core.ts` — transcribing six signatures by hand is exactly where a parameter goes missing, which is how this block got here.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -241,20 +241,30 @@ The application uses a custom channel system abstracting Electron IPC:
 - `ChannelType.MAIN`: Main process ↔ renderer communication
 - `ChannelType.PLUGIN`: Plugin-specific isolated communication
 
-**Key APIs:**
+**Key APIs** (`main/core/channel-core.ts`, class `TouchChannel`):
 ```typescript
 // Register handlers
-regChannel(type: ChannelType, eventName: string, callback): () => void
+regChannel(type: ChannelType, eventName: string, callback: ChannelCallback): () => void
 
-// Send messages
-send(eventName: string, arg?: any): Promise<any>
-sendTo(window: BrowserWindow, eventName: string, arg?: any): Promise<any>
-sendPlugin(pluginName: string, eventName: string, arg?: any): Promise<any>
+// Send messages -- note every one of these takes ChannelType, including sendTo
+send(type: ChannelType, eventName: string, arg: unknown): Promise<unknown>
+sendTo(win: BrowserWindow, type: ChannelType, eventName: string, arg: unknown): Promise<unknown>
+sendPlugin(pluginName: string, eventName: string, arg?: unknown): Promise<unknown>
 
 // Key management (encryption for plugin isolation)
-requestKey(name: string): string
+requestKey(name: string, activation?: Pick<PluginActivationIdentity, 'pluginInstanceId' | 'activationGeneration'>): string
 revokeKey(key: string): boolean
 ```
+
+> **This is the legacy surface.** New main-process code uses the TuffTransport SDK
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Superseded by #1406, already merged for #612 (closed). Same split by process, plus `requestKey`'s second parameter and the renderer's `unRegChannel`, and a guard that asserts the renderer's four absences against the class body rather than against the doc.

PR #1316, opened 2026-08-08
docs: make the BaseModule snippet something a subclass can be written against

### LINKED ISSUE #627: [audit/documentation] CLAUDE.md BaseModule snippet omits the generic and the required protected constructor

**Severity:** medium  ·  **Category:** correctness  ·  **Domain:** documentation

**Location:** `CLAUDE.md`:93

## What

The documented base-class shape has no type parameter and no constructor, so a subclass written against it does not compile and has no module key.

## Evidence

```
CLAUDE.md:93-99 shows 'abstract class BaseModule {' with only onInit/onDestroy/created/start/stop and un-parameterised contexts. apps/core-app/src/main/modules/abstract-base-module.ts:27 is 'export abstract class BaseModule<E = TalexEvents> implements TalexTouch.IModule<E>' and :36 declares 'protected constructor(key: ModuleKey, file?: ModuleFileConfig, env?: ModuleEnvFlag)' which sets 'this.name = key' (:37) — the ModuleKey the doc itself calls the singleton symbol key at CLAUDE.md:104. All five lifecycle contexts are generic (ModuleInitContext<E> etc.).
```

A developer writes `class MyModule extends BaseModule { onInit() {} onDestroy() {} }` from the snippet; TypeScript rejects the implicit `super()` (expected 1-3 arguments) and, once that is patched blindly, the module has no name so ModuleManager cannot register it as a singleton.

## Suggested direction

Show the real signature including the protected constructor and the ModuleKey argument, or show a minimal working subclass.

---

TalexDreamSoul:
Fixes #627.

## What was missing

The snippet showed five methods and nothing else. Written against it, a subclass does not compile and has no module key:

| missing | consequence |
|---|---|
| `<E = TalexEvents>` type parameter | every `ctx` type in the snippet was unparameterised |
| `implements TalexTouch.IModule<E>` | the contract it satisfies was invisible |
| `name: ModuleKey` | — |
| **`protected constructor(key, file?, env?)`** | **a subclass never passes a key up, so the manager has nothing to register it under** |
| `file` / `env` fields | directory creation and env gating undiscoverable |
| `init` / `destroy` | the snippet showed `onInit`/`onDestroy` as the entry points, when they are the hooks these wrappers delegate to |

## A real subclass is included

The constructor is the part easiest to miss and hardest to infer from a signature alone, so the corrected section carries `CommonChannelModule` verbatim — static symbol key, `name` bound to it, `super()` passing both the key and the directory config.

That is what makes this section usable rather than merely accurate.

## Verified mechanically

Each documented line was matched against `abstract-base-module.ts` and `channel/common.ts`:

```
class signature  in source=True  in doc=True
constructor      in source=True  in doc=True
name field       in source=True  in doc=True
example key      in source=True  in doc=True
example super    in source=True  in doc=True
```

Transcribing a class shape by eye is how the original lost its constructor.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -88,14 +88,45 @@ This is a pnpm workspace monorepo with the main application in `apps/core-app/`,
 4. **Stop**: Module deactivation (optional)
 5. **Destroy**: Cleanup and resource release (required)
 
-**Base Module Interface:**
+**Base Module Interface** (`main/modules/abstract-base-module.ts`):
 ```typescript
-abstract class BaseModule {
-  abstract onInit(ctx: ModuleInitContext): MaybePromise<void>
-  abstract onDestroy(ctx: ModuleDestroyContext): MaybePromise<void>
-  created?(ctx: ModuleCreateContext): MaybePromise<void>
-  start?(ctx: ModuleStartContext): MaybePromise<void>
-  stop?(ctx: ModuleStopContext): MaybePromise<void>
+abstract class BaseModule<E = TalexEvents> implements TalexTouch.IModule<E> {
+  /** Singleton id the manager keys on. Set from the constructor argument. */
+  public readonly name: ModuleKey
+  public readonly file?: ModuleFileConfig   // declarative <root>/modules/<dirName>
+  public readonly env?: ModuleEnvFlag       // env flag(s) gating auto-load
+
+  protected constructor(key: ModuleKey, file?: ModuleFileConfig, env?: ModuleEnvFlag)
+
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Superseded by #1403, already merged for #627 (which is closed). Same correction — the generic, the protected constructor, the parameterised contexts and a minimal subclass — plus the fact that `init` / `destroy` are concrete wrappers whose override silently drops the `filePath` assignment, and a guard (`base-module-doc.test.ts`) that compares every documented signature against the real declaration.

Issue #630, opened 2026-08-06
[audit/documentation] ROADMAP links three archived tasks as active with in_progress/planning status


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** medium  ·  **Category:** broken-link  ·  **Domain:** documentation

**Location:** `ROADMAP.md`:64

## What

Three rows of the active-task table link to .trellis/tasks/<id>/prd.md paths that were moved to the archive, so the links 404 and the stated statuses are wrong.

## Evidence

```
ROADMAP.md:64 '| [batch-commit-release-v2-4-14-beta-1](.trellis/tasks/07-27-batch-commit-release-v2-4-14-beta-1/prd.md) | P1 | 🔄 planning [1/2] |', :65 the same for 07-27-release-v2-4-14-beta-1 ('planning'), :83 '| [fix-file-index-update-redaction-476](.trellis/tasks/07-30-fix-file-index-update-redaction-476/prd.md) | P2 | 🔄 in_progress |'. All three directories now live under .trellis/tasks/archive/ and are absent from .trellis/tasks/; the other 24 rows resolve.
```

## Why it matters / failure scenario

A maintainer picking up 'in_progress' work clicks fix-file-index-update-redaction-476, gets a dead path, and cannot tell whether the task was completed, renamed or lost.

## Suggested direction

Drop archived rows from the active table (or link into .trellis/tasks/archive/<month>/).

---
<sub>audit-domain: documentation · confidence: confirmed</sub>

TalexDreamSoul:
## 断链只有 1 条,不是 3 条;而且它**归档里也没有**

本条说三行链接指向已归档路径。逐条核验 `ROADMAP.md` 里全部 25 处 `.trellis/tasks/*/prd.md` 链接:

```
链接总数 : 25
断链     : 1  → 07-26-install-launch-v2-4-13-beta-23
```

其余 24 处都能解析。

### 那一条不是「被移到归档」

`.trellis` 下(含 `archive/`)**任何位置都不存在**这个任务目录。所以它不是链接指错了归档路径,而是目标整个消失了——**没有可以重指的对象**,只能删掉该行。

这个区别影响修法:如果只是移到了归档,正确做法是把链接改到 `archive/` 路径、保留该行;而现在只能删。

### 已处理

[...]

PR #1319, opened 2026-08-08
docs(roadmap): stop the task table claiming to be a complete view

### LINKED ISSUE #631: [audit/documentation] ROADMAP calls its 27-row table the full active-task panorama; 63 tasks are active

**Severity:** medium  ·  **Category:** doc-drift  ·  **Domain:** documentation

**Location:** `ROADMAP.md`:53

## What

The heading promises a complete view of active Trellis tasks, but more than half of them — every task created since 2026-07-31 — are missing.

## Evidence

```
ROADMAP.md:53: '## 📁 活跃 Trellis 任务全景（27 个）'. `.trellis/tasks/` currently holds 63 task directories excluding archive/ and README.md; the table's newest entry is dated 07-30 while 30+ tasks dated 07-28 through 08-05 (e.g. 08-05-search-audit-remediation, 08-05-reco-signal-substrate, 08-04-settings-rewrite, 07-29-screenshot-*) have no row. ROADMAP.md:87 does offer the escape hatch '完整列表 → `python3 .trellis/scripts/task.py list`'.
```

A contributor checking whether work on the recommendation engine or the screenshot suite is already claimed reads the 'panorama', sees nothing, and starts a duplicate task.

## Suggested direction

Either regenerate the table from task.py or relabel it as a P0/P1 excerpt rather than 全景.

---

TalexDreamSoul:
Fixes #631, and the real part of #630.

## The completeness claim was false by 60%

The heading read 活跃 Trellis 任务全景（24 个）:

```
active tasks with a prd.md : 60
distinct tasks in the table: 23
absent                     : 37
newest row                 : 07-30
tasks created 08-03 onward : 29, none listed
```

The table stopped being maintained around 07-30 and the heading never noticed.

A list that **claims** completeness and delivers 40% is worse than one that says it is partial: a reader checking whether a task is tracked gets a wrong answer with no reason to look further. The heading now says 部分, the note carries the measured numbers and the date it stopped, and it points at the directory and `task.py list` as the real source.

## #630's count does not hold

It describes **three** rows pointing at archived paths. There is **one**: `07-26-install-launch-v2-4-13-beta-23`.

And it is not archived either — the directory exists nowhere under `.trellis`, archive included. So the link had no target to repoint at, and the row is removed rather than fixed.

## Deliberately not done: adding the 37 missing rows

Each needs a priority and a status. Inferring those from directory names would put **fabricated state** into the very document this issue exists because people trust. Completing the table is a separate job that should read `task.json` rather than guess.

## Verified after editing, not before

```
distinct tasks listed : 23
broken links          : 0
active tasks          : 60
absent                : 37
```

Counted from the file afterwards rather than trusting the numbers I had just written — the heading this PR fixes was wrong by exactly that kind of drift, and I had in fact written "23" while the raw link count said 24 (one task appears twice for parent/child indentation).


TalexDreamSoul:
Rebased onto the current base, where #1405 (for #630) had already changed the same heading and rows. Resolved to keep both points rather than one:

- The heading is now `活跃 Trellis 任务（部分，本表 22 个）` — this PR's correction that the table is **not** a complete view, plus the count #630's guard checks against the actual row count.
- The note's figure is corrected from 23 to **22**, since #1405 removed two rows whose committed `task.json` says `completed`. Re-verified the other number independently: `.trellis/tasks/*/prd.md` is exactly **60**.
- `install-launch-v2-4-13-beta-23` keeps #1405's treatment — listed but unlinked. It is a real `in_progress` task that is excluded from git via `.git/info/exclude`, so a link 404s for every clone but its author's; dropping the row would delete live work from the roadmap, and linking it ships a dead link.

`roadmap-task-links.test.ts` needed its heading regex updated to match, and passes: 5 tests.

PR #1326, opened 2026-08-08
fix(ocr): claim the processQueue guard before the first await

### LINKED ISSUE #644: [audit/main-process] OCR processQueue re-entrancy guard is set after an await, allowing duplicate job dispatch

**Severity:** medium  ·  **Category:** race  ·  **Domain:** main/modules

**Location:** `apps/core-app/src/main/modules/ocr/ocr-service.ts`:929

## What

processQueue() checks `if (this.processing) return`, then awaits isQueueDisabled() (a DB/config read) before setting this.processing = true. Two callers that arrive during that await both observe processing === false and both enter the dispatch loop.

## Evidence

```
private async processQueue(): Promise<void> {
    if (this.processing) return
    if (!this.db) return
    if (await this.isQueueDisabled()) return

    this.processing = true
```

The 5s polling task (registered at line 382) fires while failJob()'s setImmediate(() => this.processQueue()) at line 1516 is still inside the isQueueDisabled() await. Both loops select the same single ready pending job, both write status='processing' and both call runAgentJob for it — two OCR worker threads on the same image, two attempts increments, and two ocr_results rows for one job.

## Suggested direction

Set this.processing = true synchronously immediately after the guard, before any await, and clear it in the existing finally.

---
<sub>audit-domain: main/modules · confidence: confirmed</sub>

TalexDreamSoul:
Fixes #644.

## The defect

```ts
if (this.processing) return
if (!this.db) return
if (await this.isQueueDisabled()) return   // ← yields

this.processing = true                     // ← too late
```

`isQueueDisabled()` is a config read. Two callers arriving during that await both saw `processing === false`, both set it, and both ran the dispatch loop — the same pending jobs handed to workers twice.

## The fix, and the trap in it

The check and the claim are now one synchronous step. But moving the claim above an early return introduces a second failure mode: if the `isQueueDisabled()` path returns without releasing, the queue is wedged shut for the rest of the session — **worse than the bug being fixed**.

So the disabled check moves *inside* the `try`, where the existing `finally` still releases it. The second test exists specifically to hold that down.

## Verified by mutation

Against the unfixed service:

```
× lets only one concurrent caller past the guard
  → expected "spy" to be called 1 times, but got 2 times
```

Both callers really did get through. The test observes the config read rather than job dispatch, because that is the first thing past the guard and needs no database.

## Gates

- `ocr-service` suite: **10 tests passing**

### DIFF
--- a/apps/core-app/src/main/modules/ocr/ocr-service.test.ts
+++ b/apps/core-app/src/main/modules/ocr/ocr-service.test.ts
@@ -129,6 +129,10 @@ vi.mock('../ai/intelligence-sdk', () => ({
 import { ocrService } from './ocr-service'
 
 interface OcrServiceTestAccess {
+  processQueue: () => Promise<void>
+  processing: boolean
+  isQueueDisabled: () => Promise<boolean>
+  db: unknown
   runAgentJob: (jobId: number, job: Record<string, unknown>) => Promise<void>
   updateClipboardMeta: (...args: unknown[]) => Promise<void>
   normalizeSourceForAgent: (...args: unknown[]) => Promise<{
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
Rebased onto the current base. The conflict was in `ocr-service.test.ts`, where a second top-level `describe` had landed in the same region since this branch was cut. Rebuilt from the base file with this branch's `OcrService processQueue re-entrancy` block re-inserted, rather than splicing the markers — the naive splice left a block unterminated, because the closing `})` for the other side's last test sits below the conflict in the shared tail.

Verified: 11 tests pass, ESLint clean.

TalexDreamSoul:
Corrected my own rebase. Rebuilding the test file from the base and re-inserting this branch's `describe` block dropped the accompanying extension to `OcrServiceTestAccess` — `processQueue`, `processing`, `isQueueDisabled` and `db` — which lives outside the block. That failed `Typecheck (workspace)` with 20 TS2339s while the suite itself passed, so it was visible only to tsc.

Restored. `typecheck:node` = 6, unchanged from baseline; 11 tests pass; ESLint clean.

PR #1330, opened 2026-08-08
fix(terminal): contain stdin failures instead of letting them reach the process

### LINKED ISSUE #641: [audit/main-process] Terminal write() can raise an unhandled stdin 'error' and crash the main process
>
> ⚠ **Lower confidence (plausible)** — needs a repro / verification before acting; may depend on runtime behavior not confirmed by static reading.

**Severity:** medium  ·  **Category:** error-handling  ·  **Domain:** main/modules

**Location:** `apps/core-app/src/main/modules/terminal/terminal.manager.ts`:149

## What

write() calls proc.stdin.write(data) with no callback and no 'error' listener on the stdin stream. A write to a child that has already exited (or closed stdin) emits EPIPE/ERR_STREAM_DESTROYED on that stream, and an 'error' event with no listener is rethrown as an uncaught exception in the main process.

## Evidence

```
const proc = this.processes.get(id)
    if (proc && proc.stdin) {
      proc.stdin.write(data)
```

The user runs a command that ignores stdin and exits quickly (e.g. `ls`), then types into the panel in the window between the child's exit and the 'close' event that removes it from this.processes. proc.stdin.write() emits EPIPE, no listener exists, and the whole Electron main process dies.

## Suggested direction

Pass an error callback to write() (or attach proc.stdin.on('error', ...)) and drop the write when the stream is no longer writable.

---
<sub>audit-domain: main/modules · confidence: plausible</sub>

TalexDreamSoul:
Closes #641.

`proc.stdin` is a stream with its own EventEmitter and ships with **zero** `error` listeners (checked on node v24). The existing `proc.on(error)` handler is on the *child process* — it covers spawn failures, not stream failures. So an EPIPE on a session whose child had exited had nowhere to go. `write()` can also throw synchronously once the stream is destroyed, which no listener would ever see.

Both shapes are now contained and logged. The listener registration is itself optional-called (`stdin?.on?.(...)`) so it cannot become a new crash source.

### Verified
- both guards mutation-tested — removing either turns the new test red, restoring it green
- 9/9 terminal tests, `typecheck:node`, `eslint --max-warnings=0` clean

### Not verified
The issue was filed at lower confidence and asked for a repro. I could not produce one: three attempts to trigger a real unhandled EPIPE from a child stdin on node v24 did not crash the process. **The missing listener is confirmed; the crash is not reproduced.** The guard is cheap and correct regardless, but reviewers should know the severity is unproven.

Also updates the test stdin mock from `{ write }` to `{ write, on }` — the old mock could not represent a stream and so could not exercise this path.

### DIFF
--- a/apps/core-app/src/main/modules/terminal/terminal.manager.test.ts
+++ b/apps/core-app/src/main/modules/terminal/terminal.manager.test.ts
@@ -72,7 +72,8 @@ function fakeStream() {
 
 function fakeProcess() {
   return {
-    stdin: { write: vi.fn() },
+    // Mirrors a real ChildProcess.stdin: a stream, so 'error' can be listened for (#641).
+    stdin: { write: vi.fn(), on: vi.fn() },
     stdout: fakeStream(),
     stderr: fakeStream(),
     on: vi.fn(),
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
Rebased onto the current base. One conflict in `terminal.manager.test.ts`: this branch inlined `{ on: vi.fn() }` for stdout/stderr, while the base has since grown a richer `fakeStream()` helper. Resolved by keeping the base's helper and taking only this branch's actual change — `on` on `stdin`, which is the point of the fix.

Verified: 17 tests pass, ESLint clean.

Issue #655, opened 2026-08-06
[audit/main-process] TimeStatsAggregator.parseTimeStats throws on any non-JSON distribution column


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** medium  ·  **Category:** error-handling  ·  **Domain:** main/search

**Location:** `apps/core-app/src/main/modules/box-tool/search-engine/time-stats-aggregator.ts`:190

## What

parseTimeStats JSON.parses three columns unguarded and is reached from the public getItemTimeStats / getItemTimeStatsBatch API. The tolerant parser in the sibling item-time-stats-buckets.ts is not used here.

## Evidence

```
private parseTimeStats(raw: typeof schema.itemTimeStats.$inferSelect): ParsedItemTimeStats {
  return {
    sourceId: raw.sourceId,
    itemId: raw.itemId,
    hourDistribution: JSON.parse(raw.hourDistribution),
    dayOfWeekDistribution: JSON.parse(raw.dayOfWeekDistribution),
    timeSlotDistribution: JSON.parse(raw.timeSlotDistribution),
    lastUpdated: raw.lastUpdated
  }
}
```

## Why it matters / failure scenario

A row whose time_slot_distribution is an empty string (written by a legacy code path, or truncated) makes JSON.parse('') throw 'Unexpected end of JSON input'. getItemTimeStatsBatch rejects and every caller loses its whole batch rather than one row.

## Suggested direction

Delegate to parseStoredTimeBuckets from item-time-stats-buckets.ts, which already returns zero-filled buckets on parse failure.

---
<sub>audit-domain: main/search · confidence: confirmed</sub>

TalexDreamSoul:
Fixed in #1372. Confirmed as filed — `parseTimeStats` now delegates to `parseStoredTimeBuckets`, the tolerant parser already sitting in the sibling file.

**Extracted rather than fixed in place.** It is now an exported `toParsedItemTimeStats`, because **#649 is the same defect at a second call site** (`getTimeBasedTopItems`). Repeating the mapping there is precisely how the two would drift, so #649 will consume this instead of reimplementing it — I will take that one next.

**Five tests, none of which existed.** The first parses a well-formed row, placed deliberately *before* the tolerance cases: without it every assertion below would also hold for a parser that always returned zeros.

Two of them go past what a bare `try/catch` would give you:

[...]

PR #1331, opened 2026-08-08
perf(search): stop holding the WAL writer lock across the time-stats rebuild

### LINKED ISSUE #659: [audit/main-process] TimeStatsAggregator holds a write transaction open across event-loop turns

**Severity:** low  ·  **Category:** perf  ·  **Domain:** main/search

**Location:** `apps/core-app/src/main/modules/box-tool/search-engine/time-stats-aggregator.ts`:140

## What

The rebuild's transaction body awaits setImmediate every 20 upserts. Yielding inside an open SQLite write transaction keeps the WAL writer lock held while unrelated main-process work runs, widening the SQLITE_BUSY window for the search-index worker thread.

## Evidence

```
db.transaction(async (tx) => {
  for (let i = 0; i < allStats.length; i++) {
    ...
    // 每 20 条 upsert 让出事件循环
    if ((i + 1) % 20 === 0) {
      await new Promise<void>((resolve) => setImmediate(resolve))
    }
  }
})
```

With 5,000 items the rebuild yields 250 times inside one transaction. Each yield lets other main-process I/O run while the writer lock is still held, so the search-index worker's concurrent writes hit SQLITE_BUSY for the full duration of the rebuild instead of for one short transaction.

## Suggested direction

Chunk the upserts into separate scheduleDbWrite transactions and yield between chunks, the pattern backfillTrendDay already uses.

---
<sub>audit-domain: main/search · confidence: confirmed</sub>

TalexDreamSoul:
Closes #659.

The rebuild ran every upsert inside one `db.transaction` and awaited `setImmediate` every 20 rows. Yielding there hands the event loop to unrelated main-process I/O **while the WAL writer lock is still held**, so the search-index worker's concurrent writes hit `SQLITE_BUSY` for the duration of the whole rebuild rather than for one short transaction. At 5,000 items that is 250 yields inside a single open transaction.

Now chunked the way `backfillTrendDay` already does it (the pattern the issue pointed at): bounded chunks of 500, one `scheduleDbWrite` each, yield moved *between* transactions where no lock is held. Each chunk is one multi-row upsert instead of 500 round trips.

### Tradeoff worth stating explicitly
The rebuild is **no longer atomic**. A failure part-way leaves some items on fresh stats and some on stale ones, corrected by the next rebuild. This is derived data recomputed from `usage_logs`, and `backfillTrendDay` accepts the same tradeoff for the same class of data — but it is a real behaviour change, not a pure refactor.

### Verified
- new test file mutation-tested against the HEAD implementation: all 4 red on the old code, green on the new
- 579/579 search-engine tests, `typecheck:node`, `eslint --max-warnings=0` clean

The new tests pin the structural property rather than the timing: no transaction spanning the rebuild, chunks bounded at 500, every item written exactly once, and no chunking when it already fits in one write.

### DIFF
--- a/apps/core-app/src/main/modules/box-tool/search-engine/time-stats-aggregator.test.ts
+++ b/apps/core-app/src/main/modules/box-tool/search-engine/time-stats-aggregator.test.ts
@@ -1,87 +1,97 @@
-import { describe, expect, it } from 'vitest'
-import { toParsedItemTimeStats } from './time-stats-aggregator'
+import type { DbUtils } from '../../../db/utils'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 /**
- * A corrupt row must cost that row its history, not take down every caller (#655).
- *
- * hourDistribution / dayOfWeekDistribution / timeSlotDistribution are plain TEXT columns holding
- * JSON. `parseTimeStats` ran raw JSON.parse on all three, and it sits under the public
- * getItemTimeStats / getItemTimeStatsBatch API — so one partially written or truncated row threw
- * out of every consumer at once.
+ * The rebuild used to run every upsert inside one `db.transaction`, awaiting
+ * `setImmediate` every 20 rows. Yielding there keeps the WAL writer lock held
+ * while unrelated main-process work runs, so the search-index worker's
+ * concurrent writes hit SQLITE_BUSY for the whole rebuild rather than for one
+ * short transaction (#659). What these tests pin is the structural property:
+ * bounded chunks, one scheduled write each, and no transaction spanning them.
  */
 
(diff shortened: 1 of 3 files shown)


TalexDreamSoul:
Rebased onto the current base. One conflict, resolved by splitting rather than choosing:

`time-stats-aggregator.test.ts` was created independently on both sides — this branch's write-shape suite, and a JSON-tolerance suite that landed with #655. The two are disjoint but this branch's file carries module-scope `vi.mock` calls for `db-write` and `perf-context`, which would then apply to the tolerance tests as well. So the write-shape suite keeps the filename and the tolerance suite moves to `time-stats-parse.test.ts`; neither inherits the other's mocks.

Verified after the rebase: 9 tests pass across the two files, ESLint clean.

PR #1350, opened 2026-08-08
perf(search): poll plugin recommendation providers concurrently

### LINKED ISSUE #674: [audit/main-process] Plugin recommendation providers are polled sequentially with an uncleared timeout each

**Severity:** medium  ·  **Category:** perf  ·  **Domain:** main/search

**Location:** `apps/core-app/src/main/modules/box-tool/search-engine/recommendation/recommendation-engine.ts`:1741

## What

getPluginCandidates awaits each provider inside a for-of loop, so the 200ms PLUGIN_PROVIDER_TIMEOUT_MS budget is per provider rather than shared, and the setTimeout that arms each race is never cleared when the provider wins.

## Evidence

```
for (const [_id, { provider }] of this.pluginProviders) {
  try {
    if (!provider.canProvide(context)) continue

    const result = await Promise.race([
      Promise.resolve(provider.getCandidates(context)),
      new Promise<PluginRecommendCandidate[]>((_, reject) =>
        setTimeout(
          () => reject(new Error(`Provider ${provider.id} timed out`)),
          PLUGIN_PROVIDER_TIMEOUT_MS
        )
      )
    ])
```

Six plugins register recommendation providers and each takes the full 200ms (or hangs to the timeout). getCandidates blocks for 1.2s on every uncached recommend(), which is the CoreBox empty-query path — the user sees an empty grid for over a second on every open. Each iteration also leaves an unreferenced 200ms timer pending even when the provider resolved immediately.

## Suggested direction

Run the providers with Promise.allSettled so the timeout budget is shared, and clear the race timer in a finally block.

---
<sub>audit-domain: main/search · confidence: confirmed</sub>

TalexDreamSoul:
Closes #674.

`getPluginCandidates` awaited each provider inside a `for-of`, so `PLUGIN_PROVIDER_TIMEOUT_MS` was a **per-provider** budget rather than a shared one. Six plugins each taking the full 200ms blocked `getCandidates` for **1.2s on every uncached `recommend()`** — the CoreBox empty-query path, so the user watched an empty grid for over a second on every open.

The `setTimeout` arming each race was also never cleared when the provider won, leaving one pending 200ms timer per provider per call.

Providers now run under `Promise.all` with the timer cleared in a `finally`.

### Ordering is unchanged, and pinned
`Promise.all` preserves array order, so the candidate sequence is identical to the sequential version. There is a test with **deliberately inverted delays** (90/60/30/0ms) that would come back reversed if ordering had quietly started following completion instead of registration.

### Verified
- **concurrency test** red on HEAD — the sequential run measures ~250ms for 4×60ms of work
- **timer test** red on HEAD — the pending timer count stays raised after the call returns
- **ordering test** green on both, which is exactly what a control should do

578/578 search-engine tests, `typecheck:node`, `eslint --max-warnings=0` clean.

### DIFF
--- a/apps/core-app/src/main/modules/box-tool/search-engine/recommendation/recommendation-engine.test.ts
+++ b/apps/core-app/src/main/modules/box-tool/search-engine/recommendation/recommendation-engine.test.ts
@@ -1787,4 +1787,102 @@ describe('RecommendationEngine', () => {
     // is promoted because the foreground app is an IDE.
     expect(result.items[0]?.id).toBe('/Applications/Terminal.app')
   })
+
+  it('polls plugin recommendation providers concurrently, not one after another', async () => {
+    // Awaiting each provider in a for-of made PLUGIN_PROVIDER_TIMEOUT_MS a
+    // per-provider budget: six slow plugins meant 1.2s of empty grid on every
+    // uncached open of the CoreBox empty-query path (#674).
+    const engine = new RecommendationEngine(createDbUtils() as never)
+    const DELAY = 60
+    const PROVIDERS = 4
+
+    let live = 0
+    let peakConcurrent = 0
+
+    for (let i = 0; i < PROVIDERS; i++) {
+      engine.registerPluginProvider('demo-plugin', {
+        id: `provider-${i}`,
+        canProvide: () => true,
+        getCandidates: async () => {
+          live += 1
+          peakConcurrent = Math.max(peakConcurrent, live)
+          await new Promise((resolve) => setTimeout(resolve, DELAY))
+          live -= 1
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
Rebased onto the current base. The conflict was in `recommendation-engine.test.ts`, where #1347's diversity-filter tests had landed in the same region.

Both sides are purely additive, so both are kept. I rebuilt the file rather than hand-resolving the markers — a first attempt at splicing the two hunks left a test block unterminated, because the closing `})` for the last test on one side lived in the shared tail below the conflict. Taking the base file and re-inserting the three new tests inside `describe('RecommendationEngine')` avoids that class of mistake entirely.

Verified: 38 tests pass in that file, ESLint clean.

PR #1358, opened 2026-08-08
fix(nexus): stop the push worker fighting the PWA worker for scope '/'

### LINKED ISSUE #680: [audit/nexus] Web Push handler SW is registered at the same scope as the PWA SW and gets replaced

**Severity:** high  ·  **Category:** correctness  ·  **Domain:** nexus

**Location:** `apps/nexus/app/pages/dashboard/notifications.vue`:234

## What

`/notification-sw.js` is registered with no `scope` option, so it claims scope `/`. The @vite-pwa/nuxt worker `/sw.js` is configured with `scope: '/'` too and re-registers on every page load because `registerType: 'autoUpdate'`. A service-worker registration is keyed by (origin, scope), so the two scripts overwrite each other; whichever registered last owns the scope.

## Evidence

```
app/pages/dashboard/notifications.vue:234 `return await navigator.serviceWorker.register('/notification-sw.js')`
app/config/pwa.ts:5-12 `const scope = '/'` … `registerType: 'autoUpdate', scope, base: scope,`
Built bundle contains both: `navigator.serviceWorker.register("/notification-sw.js")` and the vite-pwa `register(c.sn,c.nn)` with `scope:"/"`.
```

A user enables Web Push on /dashboard/notifications: `notification-sw.js` takes scope `/` and `pushManager.subscribe()` stores a subscription on that registration. On the next page load the PWA plugin re-registers `/sw.js` at scope `/`, replacing the script. The push subscription survives on the registration but is now served by the generated `sw.js`, which has no `push` listener (public/notification-sw.js:1 is the only `addEventListener('push', ...)` in the repo), so every delivered push is silently dropped and no notification is ever shown.

## Suggested direction

Register the push handler under a dedicated scope (e.g. `/push/`) or move the `push`/`notificationclick` listeners into the vite-pwa worker via `injectManifest`/`importScripts`.

---
<sub>audit-domain: nexus · confidence: confirmed</sub>

TalexDreamSoul:
Closes #680.

`/notification-sw.js` was registered with no `scope` option, so it claimed `/`. The @vite-pwa worker is configured with `scope: "/"` too and re-registers on **every page load** because `registerType` is `autoUpdate`. A registration is keyed by (origin, scope), so the two scripts overwrote each other. The push subscription survived on the registration but was then served by the generated `sw.js`, which has no `push` listener — `public/notification-sw.js` is the only one in the repo — so delivery stopped with no error anywhere.

Registered under its own `/push/` scope instead. A scope narrower than the script's own path is always permitted, and push delivery is not scope-limited, so `subscribe()` and `showNotification()` are unaffected.

### ⚠️ Tradeoff worth a second opinion
The tidier fix is folding the push listener into the PWA worker via `workbox.importScripts` — one worker, no possible collision.

I did not do that because it **breaks push entirely when the PWA is disabled** through `NUXT_DISABLE_PWA`: there would then be no worker at all, where today the page registers its own. Two non-colliding registrations survive that case.

If `NUXT_DISABLE_PWA` is never set in any real deployment, the single-worker version is better and I am happy to redo it that way.

### Why the registration moved to a util
Nexus has no component-mounting test setup, so an SFC-internal function could not be covered without introducing one. `app/utils/push-service-worker.ts` makes the scope assertable. The page imports it explicitly via `~/utils/...`, matching the convention in `pages/docs/[...slug].vue` rather than relying on auto-import.

### Verified
Two of the four new tests are red against the pre-fix shape (scope-less `register()`). 68/68 nexus `app/utils` tests, `eslint --max-warnings=0` clean.

**Nexus typecheck still cannot run in my shell** — mise's `command_not_found` handler kills `nuxi` and `vue-tsc` before the compiler, so both exit non-zero with zero `error TS` lines. CI is the first real check, which matters more here because of that explicit import path.

### DIFF
--- a/apps/nexus/app/pages/dashboard/notifications.test.ts
+++ b/apps/nexus/app/pages/dashboard/notifications.test.ts
@@ -46,7 +46,12 @@ describe('dashboard notification inbox UI contract', () => {
     expect(page).toContain('window.Notification.requestPermission()')
     expect(page).toContain('new window.Notification')
     expect(page).toContain("tag: 'tuff-dashboard-notification-test'")
-    expect(page).toContain("navigator.serviceWorker.register('/notification-sw.js')")
+    // The page no longer registers the worker itself: doing so with no scope claimed '/' and
+    // fought the PWA worker for it. It delegates instead, and the URL and scope of that
+    // registration are asserted in push-service-worker.test.ts.
+    expect(page).toContain("import { registerPushServiceWorker } from '~/utils/push-service-worker'")
+    expect(page).toContain('registerPushServiceWorker(navigator.serviceWorker)')
+    expect(page).not.toContain('navigator.serviceWorker.register(')
     expect(page).toContain('registration.pushManager.subscribe')
     expect(page).toContain('urlBase64ToArrayBuffer(browserPushPublicKey.value)')
     expect(page).toContain('runtimeConfig.public?.notificationWebPush?.publicKey')
--- a/apps/nexus/app/pages/dashboard/notifications.vue
+++ b/apps/nexus/app/pages/dashboard/notifications.vue
@@ -5,6 +5,7 @@ import { TxRadio, TxRadioGroup, type TxRadioValue } from '@talex-touch/tuffex/ra
 import { TxSpinner } from '@talex-touch/tuffex/spinner'
 import { TxTag } from '@talex-touch/tuffex/tag'
 import { hasWindow } from '@talex-touch/utils/env'
+import { registerPushServiceWorker } from '~/utils/push-service-worker'
 import { requestJson } from '~/utils/request'
 
 const LazyFlipDialog = defineAsyncComponent(() => import('~/components/base/dialog/FlipDialog.vue'))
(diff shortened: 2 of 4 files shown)


TalexDreamSoul:
## CI 失败已修:一条断言在描述已经不存在的代码

`App suites (nexus)` 挂在 `notifications.test.ts`:

```
AssertionError: expected '<script setup lang="ts">…' to contain
  "navigator.serviceWorker.register('/n…"
```

这条断言检查的是**页面源码文本**。而本分支正是要把注册搬进 `registerPushServiceWorker`,好让它用自己的 scope 而不是和 PWA worker 抢 `/` —— 所以页面里当然不再有那段字面量。**断言在描述已经不存在的代码**,这也是这里唯一的 CI 失败项。

### 改成断言页面真正该负责的事

- 导入 `registerPushServiceWorker`
- 委托调用它
- **不再**自己调 `navigator.serviceWorker.register(`

注册用的 URL 和 scope 由它们现在所在的地方来断言 —— `push-service-worker.test.ts` 已经覆盖了 URL、scope 以及返回值透传。

### 验证

Mutation 而非假设:把页面改回内联的 register 调用,这条测试失败;还原后两个套件 7/7 通过。

Issue #716, opened 2026-08-06
[audit/tech-debt] Four copies of the 'is Nexus provider' rule; the privacy copy uses AND, the others OR


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** high  ·  **Category:** duplication  ·  **Domain:** architecture

**Location:** `apps/core-app/src/main/modules/privacy/provider-disclosure.ts`:96

## What

The predicate that decides whether an AI provider is Nexus-managed is implemented independently in four places. Three combine the id check and the origin-metadata check with OR; the fourth — the one that classifies a provider for the privacy disclosure — combines them with AND, so it recognizes strictly fewer providers than the rest of the app.

## Evidence

```
// privacy/provider-disclosure.ts:94-96 — AND
function isNexusProvider(provider: Record<string, unknown>): boolean {
  const metadata = ownValues(provider.metadata)
  return provider.id === 'tuff-nexus-default' && metadata?.origin === 'tuff-nexus'
}

// ai/provider-runtime-shared.ts:7 — OR
  return provider.id === TUFF_NEXUS_PROVIDER_ID || provider.metadata?.origin === 'tuff-nexus'

// ai/providers/nexus-provider.ts:238 — OR
  return config.id === 'tuff-nexus-default' || config.metadata?.origin === 'tuff-nexus'

// renderer/modules/intelligence/nexus-provider.ts:16 — OR
  return provider.id === TUFF_NEXUS_PROVIDER_ID || provider.metadata?.origin === 'tuff-nexus'
```

## Why it matters / failure scenario

provider-disclosure.ts:125 uses this predicate as the first branch of destinationFor: `if (isNexusProvider(provider)) return 'nexus-managed'`, falling through to `hasLoopbackEndpoint(...) -> 'local'` and finally `return 'remote'`. A provider carrying metadata.origin === 'tuff-nexus' under any id other than the literal 'tuff-nexus-default' — which the AI runtime, the model resolver and the renderer all already treat as Nexus-managed — is classified 'remote' by the disclosure instead of 'nexus-managed'. The user is then shown a privacy disclosure naming the wrong data destination for a provider the rest of the app knows is Nexus-managed.

## Suggested direction

Export one predicate from packages/utils and have all four sites call it; if the disclosure genuinely needs the stricter AND, name it differently and document why.

---
<sub>audit-domain: architecture · confidence: confirmed</sub>

[...]

PR #1380, opened 2026-08-09
ref: give the Nexus provider id a single definition

### LINKED ISSUE #717: [audit/tech-debt] 'tuff-nexus-default' provider id is a magic string repeated across 4 modules and 6 config entries

**Severity:** medium  ·  **Category:** duplication  ·  **Domain:** architecture

**Location:** `packages/utils/types/intelligence.ts`:2940

## What

The identity of the built-in Nexus provider is written as a bare string literal in the shared default-capability table (6 occurrences) and re-declared as a separate exported constant in both the main process and the renderer, with two further inline literals in the privacy and provider modules. No single definition exists.

## Evidence

```
// packages/utils/types/intelligence.ts — 6 literal occurrences (2940, 2992, 3051, 3061, 3074, 3087)
    id: "tuff-nexus-default",
      { providerId: "tuff-nexus-default", priority: 1, enabled: false },

// two independent constant declarations of the same literal
apps/core-app/src/main/modules/ai/provider-runtime-shared.ts:1
export const TUFF_NEXUS_PROVIDER_ID = 'tuff-nexus-default'
apps/core-app/src/renderer/src/modules/intelligence/nexus-provider.ts:4
export const TUFF_NEXUS_PROVIDER_ID = 'tuff-nexus-default'

// two further inline literals
apps/core-app/src/main/modules/privacy/provider-disclosure.ts:96
apps/core-app/src/main/modules/ai/providers/nexus-provider.ts:238
```

Renaming the built-in provider requires finding all ten sites across three packages, two of which are the constants that a developer would reasonably assume are the single source of truth. Missing the six literals inside DEFAULT_CAPABILITIES leaves every capability route pointing at a provider id that no longer resolves, so text.chat, embedding.generate and the four other capabilities silently fall through to their next-priority provider — a behaviour change with no error and no failing test, since DEFAULT_CAPABILITIES is data rather than code.

## Suggested direction

Define the id once in packages/utils (next to DEFAULT_CAPABILITIES), export it, and replace all ten sites with references to it.

---

TalexDreamSoul:
Closes #717.

`'tuff-nexus-default'` was a bare literal in **13** places in the shared default-capability table and in two main-process predicates, while `TUFF_NEXUS_PROVIDER_ID` already existed in `packages/utils/intelligence/nexus-provider.ts` and was used only by the renderer. All 15 now reference it, and the origin literal `'tuff-nexus'` uses `TUFF_NEXUS_PROVIDER_ORIGIN` alongside it.

Checked each of the 13 before replacing rather than running a blanket substitution: one is an `id:` field and twelve are `providerId:` entries, all naming the same provider.

No import cycle — `intelligence/nexus-provider.ts` has no imports at all, and `types/intelligence.ts` already imports from `../env`.

### ⚠️ This deliberately does NOT unify the two predicates
#716 proposes exactly that, and **it is wrong**. `privacy/provider-disclosure.ts` combines the id and origin checks with `&&` **on purpose**, so a third-party provider that declares `metadata.origin = 'tuff-nexus'` is not disclosed to the user as Nexus-managed. The fixture that pins this is literally named `custom-nexus-spoof`.

I tried the unification while working #716, and that test caught it — the spoofed provider flipped from `remote` to `nexus-managed`. Reverted there; see that issue for the detail.

Here **only the literals changed**. The `&&` and the `||` are untouched, and the spoof test still passes.

### Verified the single definition is load-bearing
Corrupting `TUFF_NEXUS_PROVIDER_ID` turns **2 tests red** — so this is not a rename that silently detaches from what the tests actually check.

1095/1096 utils tests, 229/229 privacy and ai-provider tests, `typecheck:node` 0 errors, `eslint --max-warnings=0` clean in both packages.

### DIFF
--- a/packages/utils/types/intelligence.ts
+++ b/packages/utils/types/intelligence.ts
@@ -1,4 +1,5 @@
 import { NEXUS_BASE_URL } from "../env";
+import { TUFF_NEXUS_PROVIDER_ID } from "../intelligence/nexus-provider";
 
 /**
  * Supported intelligence provider types.
@@ -2937,7 +2938,7 @@ export const DEFAULT_PROVIDERS: IntelligenceProviderConfig[] = [
     rateLimit: {},
   },
   {
-    id: "tuff-nexus-default",
+    id: TUFF_NEXUS_PROVIDER_ID,
     type: IntelligenceProviderType.CUSTOM,
     name: "Tuff Nexus",
     enabled: false,
@@ -2989,7 +2990,7 @@ export const DEFAULT_CAPABILITIES: Record<
     label: "对话 / Chat",
     description: "通用对话、问答、助理类能力",
     providers: [
-      { providerId: "tuff-nexus-default", priority: 1, enabled: false },
+      { providerId: TUFF_NEXUS_PROVIDER_ID, priority: 1, enabled: false },
       { providerId: "openai-default", priority: 2, enabled: false },
       { providerId: "anthropic-default", priority: 3, enabled: false },
       { providerId: "deepseek-default", priority: 4, enabled: true },
@@ -3048,7 +3049,7 @@ export const DEFAULT_CAPABILITIES: Record<
     label: "图片翻译 / Image Translation",
     description: "将图片中的文字翻译并返回译后图片",
     providers: [
-      { providerId: "tuff-nexus-default", priority: 1, enabled: false },
+      { providerId: TUFF_NEXUS_PROVIDER_ID, priority: 1, enabled: false },
     ],
   },
   "text.translate": {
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Rebased and **narrowed**: this now touches only `packages/utils/types/intelligence.ts`.

Its other two files were resolved independently while I was working #716. `ai/providers/nexus-provider.ts` now delegates to `isNexusManagedProvider` (via #1395), and `privacy/provider-disclosure.ts` already imports `TUFF_NEXUS_PROVIDER_ID` / `TUFF_NEXUS_PROVIDER_ORIGIN` — deliberately keeping its stricter `AND` combinator, which is anti-spoofing rather than drift, since that class renders the provider to the user as "Tuff Nexus".

What was genuinely left is the half nobody had touched: **13 occurrences** of the literal `"tuff-nexus-default"` in the shared type module — the "6 config entries" the issue names. Those now reference the constant.

Verified: 13 literals → 0, `packages/utils` nexus tests pass, `typecheck:node` unchanged at 6, ESLint clean.

PR #1386, opened 2026-08-09
ci: give every workflow job a timeout-minutes budget

### LINKED ISSUE #724: [audit/build-ci] No timeout-minutes on any release or CI job except one

**Severity:** medium  ·  **Category:** ci  ·  **Domain:** build-ci

**Location:** `.github/workflows/build-and-release.yml`:48

## What

Across all 18 workflows only windows-everything-production.yml sets timeout-minutes; the three-platform Electron release matrix, the Rust matrix and every publish job can hang for the 6-hour runner default.

## Evidence

```
$ grep -rn "timeout-minutes" .github/workflows/
.github/workflows/windows-everything-production.yml:35:    timeout-minutes: 60

.github/workflows/build-and-release.yml:48-50
  build-and-release:
    name: Build and Release Tuff App
    runs-on: ${{ matrix.os }}
(no timeout-minutes)
```

macOS notarization waits on an Apple service that stops responding, or electron-builder blocks on a download from the npmmirror mirror. The job sits idle until GitHub's 6-hour default kills it, once per matrix leg on three runners (macOS minutes bill at 10x), and `concurrency.cancel-in-progress: false` means a follow-up tag push queues behind it rather than superseding it.

## Suggested direction

Set a realistic timeout-minutes on each job (e.g. 90 for the build matrix, 20 for publish jobs).

---
<sub>audit-domain: build-ci · confidence: confirmed</sub>

TalexDreamSoul:
Closes #724.

Only `windows-everything-production.yml` set `timeout-minutes`. The other **21 jobs** — including the Electron release matrix and every publish job — could hang for the runner's 6-hour default.

### Budgets come from measured run history, not a guess
A timeout set too low is worse than none at all, because it kills legitimate slow runs. So each band is ~3–6× the observed maximum:

| workflow | budget | observed |
|---|---|---|
| `build-and-release` | 60m | successful runs at 17.7–18.2 min |
| `ci` | 30m | median 4.3, max 5.5 over 15 runs |
| `package-*` | 20m | 1.3–3.0 min |
| everything else | 15m | sub-minute helpers |

`build-and-release` had **no runs in the last 60 workflow executions**, so its number comes from a workflow-scoped query instead. Successes there are ~18 minutes, so 60 leaves room for the macOS notarization stall the issue is concerned about without being unbounded.

### Verified with the repo's own parsers
pyyaml is not installed here, so instead of a YAML library I used the checks this repo already ships:

- `check-action-pins` — reads all 18 workflows, reports clean
- `check-workflow-injection` — same, reports clean
- both self-tests pass

If either file had been left structurally broken, those parsers would have failed rather than silently skipping.

**22/22 jobs now carry a timeout.**

### DIFF
--- a/.github/workflows/ai-review.yml
+++ b/.github/workflows/ai-review.yml
@@ -16,6 +16,7 @@ jobs:
   ai-review:
     name: AI PR Review
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
       - name: Checkout trusted base commit
         uses: actions/checkout@v7
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -48,6 +48,7 @@ jobs:
   build-and-release:
     name: Build and Release Tuff App
     runs-on: ${{ matrix.image }}
+    timeout-minutes: 60
     if: github.event_name != 'workflow_dispatch' || github.event.inputs.sync_tag == ''
     permissions:
       contents: read
@@ -838,6 +839,7 @@ jobs:
     name: Create Release
     needs: build-and-release
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     if: >
       always() && (
         needs.build-and-release.result == 'success'
@@ -1272,6 +1274,7 @@ jobs:
     name: Sync Nexus Release
     needs: create-release
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     permissions:
       contents: read
     if: >
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
   job1:
     name: Check Not Allowed File Changes
     runs-on: ubuntu-latest
+    timeout-minutes: 30
     outputs:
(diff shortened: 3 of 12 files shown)


TalexDreamSoul:
Rebased onto the current base. One conflict with #1418, which pinned the release runners: it changed `runs-on: ${{ matrix.os }}` to `${{ matrix.image }}`, while this branch added `timeout-minutes: 60` to the same job. Both kept — the pinned image and the timeout budget are independent.

Also worth noting for the CI legs: #1237 has since merged, which fixes the `Protocol` failures (`libgbm`) that were red on this branch for reasons unrelated to it.

TalexDreamSoul:
Checked the completeness claim after rebasing, since "every job" is the whole point. My first scan flagged 6 jobs without `timeout-minutes` — all in `package-*-ci.yml`. That was my scan being wrong, not the PR: those are reusable-workflow callers (`uses: ./.github/workflows/package-ci.yml`), where GitHub does not accept the key, and the budget is correctly set inside `package-ci.yml` itself (`timeout-minutes: 20`).

So the claim holds: 28 jobs, all covered either directly or through the reusable workflow.

PR #1398, opened 2026-08-09
fix(utils): publish a loadable entry point [#566]

### LINKED ISSUE #566: [audit/compat] Published @talex-touch/utils has main:"index.ts"; Node 24 cannot load it (ERR_UNSUPPORTED_DIR_IMPORT)

TalexDreamSoul:
Closes #566.

Reproduced on Node 24.18.0 exactly as reported — importing the published package throws `ERR_UNSUPPORTED_DIR_IMPORT`, because `main` is a TypeScript barrel whose re-exports are extensionless directory specifiers. Only the in-repo Vite aliases hid it.

## Approach

A `tsup` build, with the published `main`/`module`/`types` pointed at it **through `publishConfig`**. In-repo resolution is byte-for-byte unchanged — the app keeps compiling the sources it always did, rather than bundling a second prebuilt copy of every singleton alongside them.

## Three measurements that contradicted the obvious fix

**1. Publishing TypeScript cannot be made to work.** Writing the directory specifiers out in full doesn't help: Node refuses with `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING` for anything under `node_modules`. And adding `.ts` extensions in source breaks the core-app typecheck, which has no `allowImportingTsExtensions` (TS5097).

**2. An `exports` map — the issue's suggestion — is a regression.** Subpath patterns resolve to exact files, and neither Node nor esbuild falls through a fallback array when the first target is merely missing. Measured against this repo's 144 distinct deep subpaths, using the current published shape as a control:

| `"./*"` form | `intelligence/nexus-provider` | `transport/events/types` |
|---|---|---|
| none (control = today) | OK | OK |
| `"./*"` | FAIL | FAIL |
| `["./*.ts", "./*/index.ts"]` | OK | FAIL |
| `["./*", "./*.ts", "./*/index.ts"]` | FAIL | FAIL |

The file half and the directory half cannot both be served by one pattern. With **no** exports map, classic resolution keeps every subpath working exactly as before.

**3. `main` has to be the ESM output.** Pointed at the CJS build, `import` goes through `cjs-module-lexer` and yields **2** named exports instead of 610. Pointed at `.mjs`, `require()` still works — Node ≥22.12 can require ESM, and `engines` already demands ≥24.15.

Also: the two `dayjs/plugin/*` imports needed an explicit `.js`. They stay external in the bundle, and Node ESM does not resolve those bare subpaths without an extension — the same failure one level down.

## Verification

Packed with `pnpm pack` and installed into a clean project:

```
main= ./dist/index.mjs  types= ./dist/index.d.mts  exports= ABSENT
ESM OK — exports: 610
CJS OK — exports: 610
```

- Bundler probe over deep subpaths matches the control exactly (the two probe failures are `node:fs/promises` under `platform: neutral`, present on both sides).
- `packages/utils`: 140 test files, 1095 passed. New `publish-shape.test.ts` (7) pins the entry shape, that the in-repo manifest stays on source, and that publishing goes through **pnpm** — `npm publish` ignores `publishConfig` field replacement, which would silently ship the bug again with every part of the fix present in the repo. Mutation-tested: restoring `main: index.ts` + `npm publish` fails 2.
- `typecheck:node` = 6, unchanged. Positive control: breaking the dayjs specifier drives it to 35, proving the file is in the program.
- `typecheck:web` = 9 on this branch, none in touched files (6 in `tuff-intelligence`, 2 `LogTerminal.vue`, 1 `theme-style.ts`) — pre-existing.
- Lockfile diff is 3 lines: `tsup` added to `packages/utils`, resolved from the existing store.

### DIFF
--- a/packages/utils/__tests__/publish-shape.test.ts
+++ b/packages/utils/__tests__/publish-shape.test.ts
@@ -1,0 +1,100 @@
+import { readFileSync } from 'node:fs'
+import path from 'node:path'
+import { describe, expect, it } from 'vitest'
+
+/**
+ * The published package must be loadable by Node (#566).
+ *
+ * @talex-touch/utils is public, and its entry was `index.ts` — a raw TypeScript barrel whose
+ * re-exports are extensionless directory specifiers. Node 24 strips the types, reparses as ESM and
+ * then cannot resolve `./account`, so `import '@talex-touch/utils'` failed outright with
+ * ERR_UNSUPPORTED_DIR_IMPORT. Only the in-repo Vite aliases hid it.
+ *
+ * The fix publishes a bundle. What is worth pinning is not that decision but the three things
+ * measured along the way, each of which looks like an improvement and is not:
+ *
+ * 1. Publishing TypeScript can never work, whatever the specifiers say. Even with `./account/index.ts`
+ *    written out in full, Node refuses: ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING. And adding those
+ *    extensions in source breaks the core-app typecheck, which has no allowImportingTsExtensions.
+ * 2. An `exports` map is the usual advice and would be a serious regression here. Subpath patterns
+ *    resolve to exact files: neither Node nor esbuild falls through a fallback array when the first
+ *    target is merely missing. Mapping the wildcard to itself breaks every deep import, and a
+ *    fallback of star-dot-ts then star-slash-index-dot-ts breaks the directory half — measured
+ *    against 144 in-repo subpaths. With no exports map, classic resolution keeps them all working
+ *    exactly as before.
+ * 3. `main` has to be the ESM output. Pointed at the CJS build, `import` goes through
+ *    cjs-module-lexer and yields 2 named exports instead of 610; pointed at `.mjs`, `require()`
+ *    still works because Node ≥22.12 can require ESM, and engines already demands ≥24.15.
+ */
+
+const ROOT = path.resolve(__dirname, '..')
+const manifest = JSON.parse(readFileSync(path.join(ROOT, 'package.json'), 'utf8')) as {
+  main: string
+  files: string[]
+  scripts: Record<string, string>
+  exports?: unknown
+  publishConfig: Record<string, string>
+  engines: { node: string }
+}
+
+describe('published entry points', () => {
+  it('resolves to built JavaScript, not to a source barrel', () => {
+    const { publishConfig } = manifest
+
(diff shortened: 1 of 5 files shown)


TalexDreamSoul:
Pushed a lint fix: `ci / CI - utils` runs ESLint over JSON too, and the manifest needed `jsonc/sort-array-values` (the `files` array) and `jsonc/sort-keys` (the build tsconfig). My local run only covered the `.ts` files.

The other three red checks are **pre-existing and unrelated** — this PR only triggered them because it touches `packages/utils`, and the path filters rarely fire them:

- **`Native Protocol` (all three OSes)** fails at *Test Rust workspace* with `rust-lld: error: unable to find library -lgbm`. A missing system library on the runners, no Rust in this PR.
- **`SDK, CLI, unavailable, and packaged native`** fails at *Build unpacked Windows snapshot*, inside the `clipboard-history` plugin build: `ERR_MODULE_NOT_FOUND` for `glob` — *"Did you mean to import glob/dist/commonjs/index.min.js?"*, thrown by `tuff builder`.

The lockfile change cannot be the cause of either: it is +3/−0 lines adding a `tsup` reference to the `packages/utils` importer. `tsup@8.5.1` appears twice in the lockfile both before and after, and no `packages:` entry is added or removed — the installed dependency graph is byte-identical, so hoisting cannot have shifted.

Both look worth their own issues; say the word and I'll file them.

Issue #767, opened 2026-08-06
[audit/main-process] A throwing post() inside the Pi run executor leaks the run timer and wedges that run id


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** medium  ·  **Category:** resource-leak  ·  **Domain:** main/modules

**Location:** `apps/core-app/src/main/modules/ai/pi-agent-runtime-host.ts`:386

## What

execute() registers the activeRuns entry (with its timeout handle) and only then calls this.post(...) inside the promise executor. post() throws when the child is gone, and postMessage throws on a non-cloneable payload; the synchronous throw rejects the promise but leaves the map entry and the timer behind.

## Evidence

```
this.activeRuns.set(payload.run.id, {
        ...
        resolve,
        reject,
        timeout
      })
      this.post({ type: 'run.start', payload })
```

## Why it matters / failure scenario

The utility process exits between `if (!this.child)` at line 370 and the post (or the request payload contains a non-structured-cloneable value). execute() rejects, but activeRuns still holds the run id, so every retry of that run fails with 'Run <id> is already active', and the timeoutMs timer fires later into settleRun on an entry whose promise is already rejected.

## Suggested direction

Wrap the post in try/catch (or post before registering) and delete the activeRuns entry plus clearTimeout on failure.

---
<sub>audit-domain: main/modules · confidence: confirmed</sub>

TalexDreamSoul:
## 已确认并修复

### 前提核实

`execute()` 先把 `activeRuns` 条目和它的超时定时器登记好,**然后**在 promise executor 内部投递 `run.start`。

而 `post()` 在"检查之后、投递之前"子进程消失时会抛错;`postMessage` 对不可结构化克隆的负载也会抛错。这个同步抛出**只**拒绝了 promise,两样东西都留了下来:

- run id 仍在 `activeRuns` 里 → **之后每一次重试都报 `Run <id> is already active`**
- 定时器仍在 → 稍后触发 `settleRun`,作用在一个已经 reject 的 promise 上

### 修法

把投递包起来:清定时器、删条目、再 reject。

**条目仍然登记在投递之前**,没有改成之后 —— 回复可能在 post 调用返回前就到达,需要能找到它。

### 验证

[...]

PR #1441, opened 2026-08-10
fix(intelligence): bound how long the Pi runtime may take to report ready

### LINKED ISSUE #766: [audit/main-process] Pi runtime start() has no readiness timeout, so a stuck worker hangs every agent run forever

**Severity:** medium  ·  **Category:** correctness  ·  **Domain:** main/modules

**Location:** `apps/core-app/src/main/modules/ai/pi-agent-runtime-host.ts`:333

## What

readyPromise is settled only by a 'runtime.ready' message or by the process exiting. There is no timer, and execute() awaits start() before arming its own run timeout, so a forked utility process that spawns but never reports ready blocks every call indefinitely.

## Evidence

```
this.readyPromise = new Promise<void>((resolve, reject) => {
      this.readyResolve = resolve
      this.readyReject = reject
    })
    const child = utilityProcess.fork(workerPath, [], {...})
// execute(): await this.start()  -> then the timeoutMs timer is created
```

The pi runtime worker hangs during module load (a slow/never-resolving dynamic import, a native module stuck on first load) and never posts runtime.ready. The process is alive, so no 'exit' fires and readyReject is never called. Every subsequent execute() awaits the same pending readyPromise; the conversation UI shows a request that never completes and never times out, and the run timeout that would have rescued it is never installed.

## Suggested direction

Race readyPromise against a bounded timer that kills the child and rejects, mirroring the run-level timeout already used in execute().

---
<sub>audit-domain: main/modules · confidence: confirmed</sub>

TalexDreamSoul:
Closes #766.

`readyPromise` was settled only by a `runtime.ready` message or by the process exiting. A worker that spawns but hangs during module load posts neither — the process is alive so no `exit` fires — and `execute()` awaits `start()` **before** arming its own run timeout, so every call parked on the same pending promise with nothing left to rescue it. The conversation UI showed a request that never completed and never timed out.

`start()` now races a **30s** timer that rejects and kills the child, letting the existing exit path do its normal cleanup. 30s is generous for what readiness actually costs — module load and one message — and far below the run timeout's ten-minute default. The point is that a stuck worker surfaces in seconds instead of never.

### Cleared at all three places readiness settles

The `runtime.ready` message, `handleExit`, and `stop()`. Missing any one of them leaves a timer that later kills a **healthy** worker.

### Two tests, each mutation caught by its own

| mutation | fails |
|---|---|
| remove the timer | the stuck-worker test — it hangs, exactly as the bug does |
| never clear it on ready | the healthy-worker test — the timer kills a runtime that had already reported ready |

The second is the control: a fix written as *"always kill after 30s"* passes the first test and fails this one.

Lint and `tsc --noEmit -p tsconfig.node.json --composite false` both exit 0. **13/13.**

### DIFF
--- a/apps/core-app/src/main/modules/ai/pi-agent-runtime-host.test.ts
+++ b/apps/core-app/src/main/modules/ai/pi-agent-runtime-host.test.ts
@@ -713,4 +713,31 @@ describe('piAgentRuntimeHost protocol boundary', () => {
     await settleAsyncWork()
     await completeRun(payload, execution)
   })
+
+  it('worker 迟迟不 ready 时,start() 有界失败并杀掉子进程', async () => {
+    // A worker that spawns but never posts runtime.ready: the process is alive so no 'exit'
+    // fires, and execute() awaits start() before arming its own run timeout, so before #766
+    // this promise stayed pending forever.
+    hostMocks.fork.mockImplementationOnce(() => hostMocks.child)
+
+    const host = new PiAgentRuntimeHost()
+    const started = host.start()
+    const assertion = expect(started).rejects.toThrow(/not ready within/i)
+
+    await vi.advanceTimersByTimeAsync(30_000)
+    await assertion
+
+    expect(hostMocks.child.kill).toHaveBeenCalled()
+  })
+
+  it('正常 ready 的 worker 不会被这个定时器杀掉', async () => {
+    // The control: a fix written as "always kill after 30s" would fail here.
+    const host = new PiAgentRuntimeHost()
+
+    await expect(host.start()).resolves.toBeUndefined()
+
+    await vi.advanceTimersByTimeAsync(60_000)
+
+    expect(hostMocks.child.kill).not.toHaveBeenCalled()
+  })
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
Rebased onto the current base — it conflicted with #1442 (#767), which landed in the same executor file.

Flagging a process hazard I hit here: **#766 had already been closed automatically** while this PR was still open and conflicting, so the issue read as fixed while `READY_TIMEOUT_MS` and `readyTimeout` appeared nowhere on the base. I reopened it.

TalexDreamSoul:
Rebased. The conflict was in `pi-agent-runtime-host.test.ts`, where #1442 (#767) added its own cases to the same block since this branch was cut.

Both sides are additive, so both are kept — rebuilt from the base file with this branch's two readiness-timeout cases re-inserted, rather than splicing the markers, because the naive splice leaves a block unterminated when the closing `})` for one side sits below the conflict in the shared tail.

Verified: 15 tests pass in that file, ESLint clean.

PR #1462, opened 2026-08-10
fix(preload): reject message events this window did not send itself

### LINKED ISSUE #797: [audit/main-process] Preload message listeners accept cross-origin messages without any origin check

**Severity:** low  ·  **Category:** security  ·  **Domain:** electron-core

**Location:** `apps/core-app/src/preload/index.ts`:730

## What

Both this handler and the loading-overlay messageListener at lines 634-640 dispatch on ev.data fields alone, never checking ev.origin or ev.source, while running in the privileged preload context. The direct assignment form also clobbers any handler the page may have registered.

## Evidence

```
window.onmessage = (ev) => {
  if (!ev.data) return
  if (ev.data.payload === 'removeLoading') {
```

A cross-origin frame posts { payload: 'removeLoading' } and dismisses the loading overlay early, or posts crafted PRELOAD_LOADING_CHANNEL events to inject text into the debug panel. Impact is currently cosmetic, but this is an unvalidated privileged-context message sink that will grow more dangerous if the handler ever gains capability.

## Suggested direction

Switch to addEventListener('message', ...) and validate ev.source === window && ev.origin === window.location.origin before dispatching.

---
<sub>audit-domain: electron-core · confidence: high</sub>

TalexDreamSoul:
Closes #797.

Both loading-overlay sinks dispatched on `ev.data` alone, in the **privileged preload context**, so any cross-origin frame embedded in the page could post `{ payload: 'removeLoading' }` or a crafted channel event and be obeyed.

### `source`, not origin, is the decisive test

Legitimate events come from `sendPreloadEvent` calling `window.postMessage` on **this same window**, so `source === window`. A frame posting to its parent carries that frame's window instead — whatever it puts in `data`. Origin is checked as well when both sides have a real one.

**Null origins are accepted deliberately.** The packaged app loads from `file:`, whose origin serialises as the string `'null'`. Rejecting on that would break the real app rather than an attacker — there is nothing to compare against, and `source` has already done the work.

### `window.onmessage =` became `addEventListener`

The assignment form replaces whatever the page had registered, which is not this preload's call to make. The issue is right to flag it.

### Six tests, three mutations, each hitting only its own

| mutation | fails |
|---|---|
| drop the `source` check | **2** — including a message with no source at all |
| drop the origin check | **1** — a same-window claim with a foreign origin |
| reject null origins | **1** — the packaged `file:` case |

That last mutation is the one worth keeping: it is the shape a *stricter-looking* fix would take, and it breaks the shipped app.

eslint **0**, tsc **0** with zero TS errors, gated before committing. **6/6**.

### DIFF
--- a/apps/core-app/src/preload/index.ts
+++ b/apps/core-app/src/preload/index.ts
@@ -10,6 +10,7 @@ import appLogoSvgRaw from '../../public/logo.svg?raw'
 import { contextBridge, ipcRenderer } from 'electron'
 import { RAW_MAIN_PROCESS_CHANNEL, RAW_PLUGIN_PROCESS_CHANNEL } from '../shared/ipc/raw-channel'
 import { resolvePreloadTargetOrigin } from './preload-target-origin'
+import { isTrustedPreloadMessage } from './trusted-preload-message'
 
 interface StartupHandshakePayload {
   rendererStartTime: number
@@ -684,6 +685,7 @@ function useLoading(options: LoadingOptions) {
   }
 
   const messageListener = (ev: MessageEvent) => {
+    if (!isTrustedPreloadMessage(ev, window)) return
     if (ev.data?.channel !== PRELOAD_LOADING_CHANNEL) return
     const payload = ev.data.data as LoadingEvent | undefined
     if (!payload) return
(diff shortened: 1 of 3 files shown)


TalexDreamSoul:
Rebased onto the current base. The conflict was a one-line import collision with #1437 (#798), which landed in the same preload file — the two fixes are independent (`resolvePreloadTargetOrigin` narrows the outgoing `postMessage` target, `isTrustedPreloadMessage` filters incoming events), so both imports are kept.

Verified after rebasing: both preload suites pass, ESLint clean.

PR #1467, opened 2026-08-10
fix(utils): make three SupportOS members name the OS they carry

### LINKED ISSUE #872: [audit/sdk] SupportOS.WATCHOS_9 and WATCHOS_10 have their values swapped

**Severity:** low  ·  **Category:** correctness  ·  **Domain:** sdk

**Location:** `packages/utils/base/index.ts`:171

## What

The `WATCHOS_9` member is assigned the string `'watchOS 10'` and `WATCHOS_10` is assigned `'watchOS 9'`. `WINDOWS_11` is likewise assigned `'Windows 10 Pro'`.

## Evidence

```
  WATCHOS = 'watchOS',
  WATCHOS_9 = 'watchOS 10',
  WATCHOS_10 = 'watchOS 9',
```

Any consumer that compares a detected OS string against `SupportOS.WATCHOS_10` matches devices running watchOS 9 and misses watchOS 10, inverting platform gating. `SupportOS.WINDOWS_11 === 'Windows 10 Pro'` similarly never matches a Windows 11 string.

## Suggested direction

Swap the two watchOS values and correct `WINDOWS_11` to `'Windows 11'`.

---
<sub>audit-domain: sdk · confidence: confirmed</sub>

TalexDreamSoul:
Closes #872.

`SupportOS` is the vocabulary a plugin manifest declares its platforms in, published through the SDK. Three members named one OS version and carried another:

| member | was | now |
|---|---|---|
| `WATCHOS_9` | `'watchOS 10'` | `'watchOS 9'` |
| `WATCHOS_10` | `'watchOS 9'` | `'watchOS 10'` |
| `WINDOWS_11` | `'Windows 10 Pro'` | `'Windows 11'` |

The watchOS pair is **inverted**, not merely wrong: gating on `SupportOS.WATCHOS_10` matched watchOS 9 devices and missed watchOS 10.

### Blast radius: none

No manifest in the repo declares `os`, and nothing compares against these values at runtime — the only `SupportOS` reference outside the enum is the `IPlatformInfo.os` field type.

`OSIcon.vue:18` does test the literal `'Windows 10 Pro'`, which is what made this worth checking before touching the enum. It is unrelated: it receives `os?.version` — a real `os.version()` string, where `'Windows 10 Pro'` is genuinely what a Windows 10 Pro machine reports. Left alone.

### Pinned as a rule, not as three literals

This is a copy-paste slip, and the next one will land on a different member. So the test asserts: **every underscore-separated segment of a key must appear in its value**, case-insensitively. `WINDOWS_11` has to say `Windows` and `11` somewhere.

That needs no table of how each OS capitalises itself (`macOS`, `iPadOS`, `WearOS`, `VisionOS` …) and it flagged **exactly these three out of 46 members** — no collateral, which is why the fix stops here.

### Five tests, four mutations

| mutation | fails |
|---|---|
| restore all three original values | the rule, both literal pins |
| re-swap the watchOS pair only | the rule, the watchOS pin |
| **over-correct**: `WINDOWS_11 = 'Windows 10'` | the rule, the Windows pin, **and uniqueness** |
| **blind the rule** | the positive control only |

The last one is the point of the control: with the filter neutered, "46 members are all consistent" passes **vacuously**.

utils suite **156 files / 1185 passed**, eslint **0** at `--max-warnings=0`.

### DIFF
--- a/packages/utils/__tests__/support-os-labels.test.ts
+++ b/packages/utils/__tests__/support-os-labels.test.ts
@@ -1,0 +1,54 @@
+/**
+ * SupportOS is the vocabulary a plugin manifest declares its platforms in, published through the
+ * SDK. Three members named one OS version and carried another: WATCHOS_9 and WATCHOS_10 had their
+ * values swapped, and WINDOWS_11 was 'Windows 10 Pro' (#872). A consumer gating on
+ * SupportOS.WATCHOS_10 matched watchOS 9 devices and missed watchOS 10 - inverted, not merely
+ * broken.
+ *
+ * The pin is the general rule rather than three literals, because the bug is a copy-paste slip and
+ * the next one will land on a different member. Every underscore-separated segment of the key must
+ * appear in the value: WINDOWS_11 must say 'Windows' and '11' somewhere, whatever the casing. That
+ * needs no table of how each OS capitalises itself, and it caught exactly these three out of 46.
+ */
+import { describe, expect, it } from 'vitest'
+
+import { SupportOS } from '../base/index'
+
+const MEMBERS = Object.entries(SupportOS) as [string, string][]
+
+describe('SupportOS values match the member they are named for', () => {
+  it('规则本身有效:构造一个错值会被判为违规(缺这条,下面的空结果什么都不证明)', () => {
+    expect(violations([['WATCHOS_10', 'watchOS 9']])).toEqual(['WATCHOS_10 = "watchOS 9"'])
+    expect(violations([['WATCHOS_10', 'watchOS 10']])).toEqual([])
+  })
+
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
Rebased onto the current base after #1466 landed — this branch carried its commits too, so the diff is now just the `SupportOS` correction.

Checked the one consumer before merging: `OSIcon.vue` compares against `'Windows 10 Pro'`, which looks like it depends on the wrong value. It does not — it matches **OS-reported strings**, and both call sites feed it runtime values (`os?.version` in SettingAbout, a plugin manifest's `platforms` key in PluginOverview), never a `SupportOS` member. So correcting the enum does not change any icon.

Issue #915, opened 2026-08-06
[audit/security] Permission guard allows any API name that is missing from the mapping table


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** medium  ·  **Category:** security  ·  **Domain:** security

**Location:** `apps/core-app/src/main/modules/permission/permission-guard.ts`:159

## What

PermissionGuard.check() returns allowed:true whenever an API name matches no entry in API_PERMISSION_MAPPINGS, so the permission model is an opt-in denylist keyed on a hand-maintained string table rather than a default-deny gate.

## Evidence

```
    const requiredPermissions = this.getRequiredPermissions(apiName)

    if (requiredPermissions.length === 0) {
      // No permission required for this API
      const duration = performance.now() - startTime
      this.recordPerformance(duration)
      return {
        allowed: true,
        permissionId: '',
        pluginId,
        durationMs: duration
      }
    }
```

## Why it matters / failure scenario

API_PERMISSION_MAPPINGS (lines 52-126) enumerates colon-separated names such as 'network:request' and 'system:exec'. Any capability whose call site passes a name not in that list -- including every capability added since the table was last updated -- is granted unconditionally to any plugin. Combined with the fact that only six main-process files call withPermission/enforcePermission at all, a plugin's manifest permission declarations bound almost nothing: the network, download, terminal-write and shell handlers reported above never reach this function in the first place, and anything that does reach it with an unmapped name is waved through.

## Suggested direction

Default-deny: return allowed:false for unmapped API names and require every guarded call site to name a permission explicitly, with a test that fails when a transport event has no mapping.

---
<sub>audit-domain: security · confidence: confirmed</sub>

TalexDreamSoul:
## Instrumented, not flipped — PR #1182, merged as `48f3cdf10`

```
permission-guard.ts (13112 bytes — control that it was read)
  recordUnmappedApi : 2   getUnmappedApis : 1   console. : 0
```

The premise is confirmed: `check()` returns `allowed: true` for any name with no mapping entry, so the model is an opt-in denylist keyed on a hand-maintained table.

**I did not flip the default, and that is the substance of this comment.**

## Why flipping is not currently possible

Four call sites reach the guard, and two pass a caller-supplied string:

```
plugin-storage-transport-service.ts:359   checkPermission(plugin.name, apiName, …)
[...]

PR #1484, opened 2026-08-10
feat(permission): gate new call sites on having a mapping entry

TalexDreamSoul:
Addresses #915. The default-deny flip is **not** made here, and the PR says why.

### What was already done

`PermissionGuard.check()` returns `allowed: true` for any API name that matches no entry in `API_PERMISSION_MAPPINGS`. PR #1182 already landed the first half of the answer: every unmapped name is recorded and warn-logged on first sighting, so the allow leaves a trace.

I started to add that logging before noticing it exists — the recording is not inert, and the code comment there already states the reasoning.

### What still blocks the flip

Returning `allowed: false` for unmapped names needs an inventory of every name that legitimately reaches the guard, and only production traffic can build it. That is a judgement call about which capabilities are deliberately public, not something derivable from the repo.

### What this PR does — the half that needs no inventory

**Every API name written at a call site today must already be in the table.** All **19** literals reaching `enforcePermission` match one of the **50** patterns, so nothing is broken right now. The value is prospective: adding a call site with an unmapped name stops being a silent allow found later in production logs and becomes a red PR.

Scope note: `withPermission({ permissionId })` is deliberately not scanned — it names a permission directly and never consults the table, so it cannot have a mapping gap. Terminal's three handlers use that form.

### Why a script and not a vitest file

`App suites (core-app)` is `continue-on-error: ${{ matrix.app != 'nexus' }}`, so **a test there cannot fail a PR**. `PR Quality` runs scripts as gates, which is where this goes, following `check:search-index-writers` — including its `--self-test` convention, which is the positive control institutionalised.

### Evidence it actually gates

- `--self-test`: 5 cases + a reachability check. Exit **0**.
- **negative control on the real tree**: renaming `native:screenshot:capture` to an unmapped name in `screenshot-session/index.ts` makes the gate print that file and line and exit **1**. Restored, exit **0**.

The reachability case matters on its own: "zero unmapped names" is vacuous if the scan found zero names, so the self-test asserts it sees both the real table and the real call sites.

### One unrelated thing left alone

`apps/core-app/package.json` has `"zod"` indented 12 spaces. My JSON round-trip normalised it; I restored it, because a reviewer should not have to wonder why zod moved in a permissions PR.

eslint **0** with no warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,6 +124,11 @@ jobs:
       - name: Verify search-index writer ownership
         run: pnpm -C apps/core-app check:search-index-writers && pnpm -C apps/core-app check:search-index-writers:self-test
 
+      # Runs here rather than as a core-app vitest file: the App suites job is
+      # continue-on-error, so a test there cannot fail a PR (#915).
+      - name: Verify permission API mapping coverage
+        run: pnpm -C apps/core-app check:permission-api-mappings && pnpm -C apps/core-app check:permission-api-mappings:self-test
+
       # The script existed but no workflow ran it (#555). ~1s: re-derives the sensitive-data
       # inventory's 28 evidence anchors from the TypeScript AST, so it needs the root
       # `typescript` dependency and nothing else.
--- a/apps/core-app/package.json
+++ b/apps/core-app/package.json
@@ -87,7 +87,9 @@
     "quickops:evidence:verify": "pnpm exec tsx \"scripts/quickops-evidence-verify.ts\"",
     "context:entrypoint:evidence:verify": "pnpm exec tsx \"scripts/context-entrypoint-evidence.ts\"",
     "check:search-index-writers": "node scripts/check-search-index-writers.mjs",
-    "check:search-index-writers:self-test": "node scripts/check-search-index-writers.mjs --self-test"
+    "check:search-index-writers:self-test": "node scripts/check-search-index-writers.mjs --self-test",
+    "check:permission-api-mappings": "node scripts/check-permission-api-mappings.mjs",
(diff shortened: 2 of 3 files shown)


TalexDreamSoul:
Superseded — its substance is already on `TalexDreamSoul/app-shell-v2`.

The gate script, the two `package.json` scripts and the CI step all landed with #1485, whose diff carried this work alongside its own download fix. This branch's remaining commit is the docblock count correction (12 → 19/50), and the base already has the corrected numbers — which is exactly what the rebase conflicted on.

Verified on the base rather than assuming: `check:permission-api-mappings` reports *19 guarded API name(s) all match one of 50 mapping pattern(s)*, and the self-test passes its four cases including the one that proves the scan reads the real table and the real call sites.

TalexDreamSoul:
自我更正一句,并附上后续:**这个 PR 正文把闸门的覆盖面说大了。**

正文写的是「今天代码里写死的每个 API 名都必须已在表里」。实际上当时它只看见**直接**以字面量调用 `enforcePermission(x, '...')` 的那一类 —— 45 个调用点里的 5 个。绝大多数调用点走的是包装函数,把 API 名当参数传进去,静态扫描那一版看不见。

也就是说,合入时闸门是真的,但它守住的范围比正文承诺的窄得多。

**这一条已经被 #1523 补上了**(2026-08-11 合并)。我刚在最新的 `TalexDreamSoul/app-shell-v2` 上就地跑了一次闸门核实,不是照抄 PR 描述:

```
[permission-api-mappings] 70 guarded API name(s) all match one of 50 mapping pattern(s)
exit 0
```

70 个,不是 5 个。所以留这条只为把记录摆正 —— 别人回头翻这个 PR 时,不该以为它当时就覆盖了全部调用点。

PR #1508, opened 2026-08-10
fix(plugins): make plugins:validate reject committed dev leakage

### LINKED ISSUE #812: [audit/plugins] plugins:validate has no dev.enable check, so committed dev leakage passes the gate

**Severity:** medium  ·  **Category:** config  ·  **Domain:** plugins

**Location:** `scripts/validate-plugins.mjs`:126

## What

AGENTS.md prescribes `pnpm plugins:validate` as the manifest gate, but the validator never inspects the dev block. Running it against the current tree reports 24/24 passed while three manifests ship dev.enable:true with localhost addresses.

## Evidence

```
Actual run output: `24/24 plugins passed validation.` / `All plugins validated successfully.` A grep for dev handling in the script returns only line 126-127, which is about a missing manifest, not the dev block:
  logWarn(pluginName, hasPackageJson ? 'manifest.json not found — skipping Surface-only plugin' : 'manifest.json not found — skipping')
```

A contributor follows plugins/AGENTS.md:52 ('Manifest 校验：pnpm plugins:validate'), sees an all-green result, and ships clipboard-history / touch-intelligence / touch-translation with dev.enable:true. The prescribed gate cannot catch the exact class of defect it is cited for.

## Suggested direction

Fail validation when a committed manifest has dev.enable true.

---
<sub>audit-domain: plugins · confidence: confirmed</sub>

### LINKED ISSUE #815: [audit/plugins] touch-image: package.json embeds talex-touch.plugin.dev.enable:true with 127.0.0.1:6001

**Severity:** low  ·  **Category:** config  ·  **Domain:** plugins

**Location:** `plugins/touch-image/package.json`:13

## What

The committed package.json carries a talex-touch block that turns on dev mode and hardcodes a localhost address; this block is what the build copies into dist/manifest.json.

## Evidence

```
plugins/touch-image/package.json:11-18  "talex-touch": {
    "icon": {...},
    "plugin": {
      "dev": {
        "enable": true,
        "address": "http://127.0.0.1:6001"
      }
    },
```

Anyone building touch-image gets dist/manifest.json with dev.enable:true baked in, reproducing the same localhost-dev leakage as the three shipped manifests, and there is no validator step that inspects package.json talex-touch blocks.

## Suggested direction

Set enable to false, or move the dev address into a git-ignored local override.

---
<sub>audit-domain: plugins · confidence: confirmed</sub>

TalexDreamSoul:
Closes #812. Closes #815.

Fixed together because a gate that lands red is not a gate: `touch-image` was the **only** violation the new check found, so the check and its one casualty go in one commit.

### The gate could not catch what it is cited for

`plugins/AGENTS.md` prescribes `pnpm plugins:validate` as the manifest gate. Neither `validate-plugins.mjs` nor `validate-plugin-package-policy.ts` mentioned `dev` at all — grep returns **zero** matches in both.

Now checked in `manifest.json` **and** in `package.json`'s `talex-touch.plugin` block, because the build copies the latter into `dist/manifest.json` — which is exactly how `touch-image` was leaking.

Only `enable` is policed. A `dev.address` alongside `enable: false` is the normal way to keep a local server configured without turning it on, and every other manifest in the tree uses it that way.

### The issue's evidence is stale — the conclusion is not

#812 cites *"three manifests ship dev.enable:true"* (clipboard-history / touch-intelligence / touch-translation). Read directly, **all three are `enable: false` today**; they keep localhost addresses, which is the supported configuration. Someone fixed them after the sweep.

I verified that with a positive control first, because an empty scan is exactly what a broken scan looks like: the same query does find their `"dev"` blocks.

So the leakage claim is out of date. **The gate claim is not** — and that is the part worth fixing, since it is what lets the next one through.

### Evidence the gate actually gates

- **negative control on the real tree**: with `touch-image` restored to `enable: true`, `plugins:validate` names the file and exits **1**. Fixed, exits **0**.
- **`--self-test`**: 5 cases including `enable: true` with no address, `enable: false` with an address, and a non-object `dev` block. Wired into the same blocking `PR Quality` step, following the repo's `check:X && check:X:self-test` convention.

The self-test exists because "no plugin declares `dev.enable`" reads identically whether the check works or does nothing — which is the state this script was in.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,7 +135,7 @@ jobs:
       # until #1099 removed it (#1128). If the declaration this commit adds is wrong, this
       # step is where it shows up rather than in whichever release script runs next.
       - name: Validate plugin manifests and package policy
-        run: pnpm plugins:validate
+        run: pnpm plugins:validate && pnpm plugins:validate:self-test
 
       # `pnpm lint` reaches plugins either as workspaces or through a hand-maintained brace
       # list, and touch-dictation was in neither (#562). The list is the kind that drifts
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "release:notes:prepare": "node scripts/release-notes-gateway.mjs prepare",
     "release:notes:verify": "node scripts/release-notes-gateway.mjs verify",
     "plugins:validate": "tsx scripts/validate-plugin-package-policy.ts && node scripts/validate-plugins.mjs",
+    "plugins:validate:self-test": "node scripts/validate-plugins.mjs --self-test",
     "plugins:release:audit": "tsx scripts/plugin-source-package-audit.ts",
     "publish:check": "node scripts/validate-publish-manifests.mjs",
     "publish:check:pack": "node scripts/validate-publish-manifests.mjs --pack",
--- a/plugins/touch-image/package.json
+++ b/plugins/touch-image/package.json
@@ -17,7 +17,7 @@
     },
     "plugin": {
       "dev": {
-        "enable": true,
+        "enable": false,
         "address": "http://127.0.0.1:6001"
       }
     },
--- a/scripts/validate-plugins.mjs
+++ b/scripts/validate-plugins.mjs
@@ -51,6 +51,31 @@ const KNOWN_PERMISSION_IDS = new Set([
   'lexicon.register',
 ])
 
+/**
+ * A committed `dev.enable: true` points the loader at a developer's local server, so a shipped
+ * plugin either fails to load or loads whatever is answering on that port.
+ *
+ * AGENTS.md prescribes `pnpm plugins:validate` as the manifest gate, but neither validator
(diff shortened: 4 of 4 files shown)


TalexDreamSoul:
Closing as a duplicate. #1504 landed the same fix ~12 minutes before I started this branch — the dev.enable check in `validate-plugins.mjs` reading both `manifest.json` and `package.json`'s `talex-touch.plugin.dev`, plus the `touch-image` block it turns red.

My working base was 47 commits behind `origin/TalexDreamSoul/app-shell-v2`, so the defect still looked live in my tree: the validator genuinely had zero `dev` handling and `touch-image` genuinely still had `enable: true` — in a stale checkout. I verified the tree, not the branch, and that was the mistake.

Nothing here is worth salvaging over #1504: it ships the same two-source check with the same enable-only policy, and covers it with a dedicated test file (`validate-plugins.dev-leakage.test.mjs`) rather than the `--self-test` flag I added. I am not going to re-litigate that difference for an already-merged fix.

PR #1511, opened 2026-08-10
fix(tuff-cli): declare the glob its bundle imports at runtime

### LINKED ISSUE #1435: [audit/build-ci] Windows snapshot build cannot resolve 'glob' from the bundled tuff-cli chunk
Every PR that reaches the packaged-native job fails the same way, on the *Build unpacked Windows snapshot* step:

```
Cannot find package 'glob' imported from D:\a\tuff\tuff\packages\tuff-cli\dist\chunk-BREXGMSJ.js
    at packageResolve (node:internal/modules/esm/resolve:768:81)
  code: 'ERR_MODULE_NOT_FOUND'
```

thrown while running `pnpm --filter @talex-touch/clipboard-history-plugin run build`, whose script is `vite build && tuff builder && …`, from `apps/core-app/scripts/lib/touch-translation-runtime-sync.js:61`.

## It is not a missing declaration

`packages/tuff-cli/package.json` **does** declare `"glob": "^10.5.0"`, and `glob@10.5.0` has a correct exports map (`import` → `./dist/esm/index.js`). So the manifest is right and the failure is in how the build target's isolated install lays out `node_modules` — the step logs `[build-target] Removed temporary pnpm-lock.yaml`, so `tuff-cli`'s dependencies are being resolved against a synthesised tree rather than the workspace one.

## Why it is worth its own issue

It is **pre-existing and unrelated to whatever PR triggers it** — confirmed by the identical error, down to the same `chunk-BREXGMSJ.js` hash, on PRs as unrelated as a `packages/utils` packaging change, a lint-glob removal, and a nested-peer resolution fix. Six PRs were sitting behind it today.

Windows-only, so it cannot be reproduced on a macOS or Linux checkout; it needs someone with a Windows runner or a local reproduction of the build-target isolated install.

Filed while working through the audit backlog; the six PRs above were merged after confirming the failure was not theirs.

### LINKED ISSUE #1509: Windows production gate fails on every PR: tuff-cli imports 'glob' without declaring it
Every pull request opened since **2026-08-10 05:01 UTC** fails the `Windows Everything Production Gate` check, on branches with nothing in common:

| branch | result |
|---|---|
| `dependabot/npm_and_yarn/electron-43.2.0` (03:46) | success |
| `ci/721-core-app-max-warnings` (05:01) | failure |
| `fix/915-permission-mapping-gate` | failure |
| `fix/1457-single-stream-download-atomicity` | failure |
| `fix/850-native-error-cause` | failure |
| `build/602-extract-file-icon-optional` | failure |
| `build/801-drop-dead-snap-config` | failure |
| `chore/816-drop-unused-music-deps` | failure |

Identical error in each:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'glob'
  imported from D:\a\tuff\tuff\packages\tuff-cli\dist\chunk-KINUHW5U.js
```

It surfaces during `Build unpacked Windows snapshot`, when `after-pack` builds the official plugin seeds and `@talex-touch/clipboard-history-plugin`'s build runs `vite build && tuff builder && …`. `vite build` succeeds; `tuff builder` dies on the import.

## Root cause

`packages/tuff-cli/tsup.config.ts` lists `glob` in `external`, so the emitted bundle keeps `import … from 'glob'` as a runtime import. `packages/tuff-cli/package.json` does not declare `glob` anywhere — not in `dependencies`, not in `peerDependencies`.

The import comes from `@talex-touch/tuff-cli-core`, which is in `noExternal` and therefore inlined into the bundle. `tuff-cli-core` declares `glob: ^13.0.6` correctly, but it is a **devDependency** of `tuff-cli`, so its declaration does not follow the code that gets bundled into `dist/`.

I built `dist/` and scanned every bare import against the declared set. `glob` is the only real one missing (19 distinct bare imports found, so the scan is not silently empty; the other unmatched string is a `${args.importer}` template literal in bundled esbuild plugin code, not an import). `rollup` is also in `external` without being declared, but nothing imports it, so esbuild never emits it.

## Why it broke now rather than at introduction

Nothing about `tuff-cli` changed. `glob` had been resolving through hoisting into the workspace root — every developer machine and every other CI job does a full workspace install, which puts `glob` where `dist/` can find it.

The Windows gate does not. `build-target.js` writes a temporary pruned `pnpm-lock.yaml` (`[build-target] Removed temporary pnpm-lock.yaml` in the log) and installs from it, which drops dev-only packages. Without `tuff-cli-core` installed, nothing pulls `glob` in, and the undeclared import has nothing left to resolve against.

So this is a latent packaging defect that a lockfile change in the 03:46–05:01 window exposed, not a regression in any one of those commits. The same defect would hit anyone installing the published `@talex-touch/tuff-cli` from npm.

## Fix

Declare `glob` in `packages/tuff-cli` `dependencies`, and add a guard that fails when `tsup.config.ts` marks a package `external` that the built `dist/` actually imports and `package.json` does not declare. That guard is the part that matters: the current arrangement cannot detect the drift, which is why an undeclared runtime import survived long enough to block the whole queue.

TalexDreamSoul:
Closes #1509. Closes #1435 — same failure, and the two issues disagree about its cause; see below.

`tuff-cli` bundles `tuff-cli-core` (tsup `noExternal`) but keeps `glob` **external**, so the emitted chunk carries `import … from 'glob'` at runtime. `tuff-cli-core` declares `glob: ^13.0.6` correctly — but it is a **devDependency** of `tuff-cli`, so that declaration does not follow the code into `dist/`. `tuff-cli`'s own manifest mentioned `glob` nowhere.

`glob` is now a dependency of `tuff-cli`. The lockfile delta is **3 lines** — the importer entry only; `glob@13.0.6` was already resolved for `tuff-cli-core`, so no package-set change.

### The two issues disagree, and #1435 is out of date

| | claim about `tuff-cli/package.json` |
|---|---|
| #1435 (02:39) | *"**does** declare `"glob": "^10.5.0"`"* → concludes it is an install-layout problem |
| #1509 (09:10) | declares it nowhere → concludes it is a missing declaration |

**#1509 is right for the current tree.** `git log -S '"glob"'` shows #1312 (Aug 8, *"move off the deprecated glob line"*) **removed** `"glob": "^10.5.0"` — it moved the dependency to `tuff-cli-core@^13.0.6` and did not re-add it to `tuff-cli`, whose bundle still needs it.

So #1435's "the manifest is right, the isolated install is wrong" reading sent the investigation at the wrong layer, and its "needs a Windows runner" conclusion follows from that: the missing declaration is not platform-specific — Windows was just where the isolated install exposed it first.

### The guard reads the artifact, not the config

`external` also lists `rollup` (**never imported** — checked) and ~20 optional template engines that only `@vue/compiler-sfc` reaches dynamically. A config-based check would demand declarations for all of them and be wrong.

Scanning `dist/` names only what is actually required: **10 imports today, all declared**.

### Evidence it gates

- **negative control on the real artifact**: removing the declaration again makes the check name `chunk-BREXGMSJ.js` — **the same chunk in #1435's error message**, which is also what ties the two reports to one cause.
- **`--self-test`**: 7 cases — undeclared import, declared import, subpath→package, node builtins, relative paths, a template-literal dynamic import, and `require()`.

Wired into the `tuff-cli` CI job right after its build, where `dist/` exists. That workflow is path-filtered to the tuff-cli packages, which is where this can regress.

### Not verified here

I cannot run the Windows gate. What I verified is the layer the failure lives at: the built bundle imports `glob`, and the package that ships it did not declare it.

### DIFF
--- a/.github/workflows/package-tuff-cli-ci.yml
+++ b/.github/workflows/package-tuff-cli-ci.yml
@@ -51,3 +51,5 @@ jobs:
         pnpm --dir ../.. --filter "@talex-touch/tuff-cli-core" run build
         pnpm --dir ../.. --filter "@talex-touch/unplugin-export-plugin" run build
         pnpm build
+        pnpm check:externals
+        pnpm check:externals:self-test
--- a/packages/tuff-cli/package.json
+++ b/packages/tuff-cli/package.json
@@ -27,7 +27,9 @@
     "dev": "tsup --watch",
     "lint": "eslint \"src/**/*.{ts,tsx}\"",
     "lint:fix": "eslint \"src/**/*.{ts,tsx}\" --fix",
-    "clean": "rimraf dist"
+    "clean": "rimraf dist",
+    "check:externals": "node ../../scripts/check-bundled-externals.mjs",
+    "check:externals:self-test": "node ../../scripts/check-bundled-externals.mjs --self-test"
   },
   "dependencies": {
     "@talex-touch/utils": "workspace:^",
@@ -38,6 +40,7 @@
     "debug": "^4.4.3",
     "esbuild": "^0.28.1",
     "fs-extra": "^11.3.6",
+    "glob": "^13.0.6",
     "pathe": "^1.1.2",
     "unplugin": "^2.3.11",
     "vite": "^7.3.6"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -921,6 +921,9 @@ importers:
       fs-extra:
         specifier: ^11.3.6
         version: 11.3.6
+      glob:
+        specifier: ^13.0.6
+        version: 13.0.6
       pathe:
         specifier: ^1.1.2
         version: 1.1.2
--- a/scripts/check-bundled-externals.mjs
+++ b/scripts/check-bundled-externals.mjs
@@ -1,0 +1,194 @@
+#!/usr/bin/env node
+/**
+ * Every bare import that survives into a published bundle must be declared by the package that
+ * ships it.
+ *
+ * `@talex-touch/tuff-cli` bundles `@talex-touch/tuff-cli-core` (it is in tsup's `noExternal`) but
+ * keeps `glob` external, so the emitted chunk carries `import … from 'glob'` at runtime.
+ * `tuff-cli-core` declares `glob` correctly — but it is a *devDependency* of `tuff-cli`, so that
+ * declaration does not follow the code into `dist/`. Nothing in `tuff-cli`'s own manifest mentioned
+ * `glob`, and the Windows production gate failed on every PR with `ERR_MODULE_NOT_FOUND` (#1509).
+ *
+ * It reads the built artifact rather than the tsup config on purpose. `external` also lists
+ * `rollup` (never imported) and ~20 optional template engines that only `@vue/compiler-sfc`
+ * reaches dynamically. A config-based check would demand declarations for all of them; the bundle
+ * names only what is actually required.
+ *
+ * Read-only. `--self-test` proves the detector fires, because "every import is declared" looks the
+ * same whether the check works or matched nothing at all.
+ */
+
+import fs from 'node:fs'
+import { builtinModules } from 'node:module'
+import path from 'node:path'
+import process from 'node:process'
+import { fileURLToPath } from 'node:url'
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url))
+const rootDir = path.resolve(__dirname, '..')
+
+/** Packages whose bundles are published and therefore have to stand on their own manifest. */
+const TARGETS = [{ name: '@talex-touch/tuff-cli', dir: 'packages/tuff-cli' }]
+
+const BUILTINS = new Set([...builtinModules, ...builtinModules.map(name => `node:${name}`)])
+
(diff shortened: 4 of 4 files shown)


TalexDreamSoul:
Closing as a duplicate of #1510, which was opened 6 minutes before this one and is the better fix.

Two things it does that this does not:

- it also removes `rollup` from `external` — a dead entry I noticed and only mentioned
- its check lives in `packages/utils`, whose `ci / CI - utils` job is **blocking**. Mine runs in `Tuff CLI Package CI`, which is path-filtered to the tuff-cli packages, so it would not fire on a change elsewhere that breaks the same invariant

I had assumed a config-based rule would over-report, because `external` also lists `rollup` and ~20 optional template engines. #1510 answers both: it slices `external: [` to the first `]`, which excludes the engines' separate const, and it deletes `rollup`. Its stated invariant is also sound — tsup already externalises declared dependencies and node builtins are external under `platform: 'node'`, so a bare import can only reach `dist/` via that array, a declaration, or a builtin.

The one thing only an artefact scan would catch is a bare import that reaches `dist/` *without* going through `external` — a future inlined package doing a dynamic `require`, say. That is speculative against the current config and not worth a competing gate. If it ever bites, `scripts/check-bundled-externals.mjs` in this branch's history is a starting point.

My process failure, not a disagreement: I did not check for an existing PR on the issue before starting.

Issue #839, opened 2026-08-06
[audit/rust] MacFrameSource::stop() builds its Notified after the flag check and can hang forever


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** high  ·  **Category:** correctness  ·  **Domain:** rust

**Location:** `packages/tuff-native/native-screenshot/src/backend/macos/actor.rs`:167

## What

`stop()` checks `stopped` and only then constructs `notified()`. tokio's Notified snapshots the notify_waiters counter at construction, so a `notify_waiters()` landing between the flag load and that construction is already accounted for and the future parks with no further wakeup coming.

## Evidence

```
        Box::pin(async move {
            while !control.stopped.load(Ordering::Acquire) {
                control.stop_notify.notified().await;
            }
            Ok(())
        })
// producer (actor.rs:887-889):
        control.slot.request_stop();
        control.stopped.store(true, Ordering::Release);
        control.stop_notify.notify_waiters();
```

## Why it matters / failure scenario

`ScreenshotCapability::invoke_stream` calls `source.request_stop()` then awaits `source.stop()`. The awaiting task loads `stopped` (false); the actor thread then runs `stopped.store(true)` + `notify_waiters()`; only afterwards does the task evaluate `control.stop_notify.notified()`, whose snapshot of num_notify_waiters_calls already includes that call, so poll_notified registers and parks. `stop()` never resolves: the frames operation never emits its terminal frame, its entry stays in NativeRuntime::streams, and finish_dispose() eventually fails with NATIVE_DISPOSE_TIMEOUT. The same crate gets this right elsewhere — native-core/src/stream.rs:53, native-core/src/runtime.rs:597 and backend/mod.rs:314 all build the Notified before checking state.

## Suggested direction

Construct the `Notified` before reading `stopped` (or pin it and call `.enable()`), matching the pattern already used in stream.rs/runtime.rs.

---
<sub>audit-domain: rust · confidence: confirmed</sub>

TalexDreamSoul:
Confirmed and fixed — PR #1536. Same class as #840, which is #1535.

The read-then-construct order is exactly as reported, and so is the consequence chain: `stop()` never resolves → no terminal frame → the entry stays in `NativeRuntime::streams` → `finish_dispose()` returns `NATIVE_DISPOSE_TIMEOUT`.

## The three sites you cited are a useful control

[...]

PR #1534, opened 2026-08-11
fix(native-core): build the cancel Notified before reading the state

### LINKED ISSUE #840: [audit/rust] CancellationToken::cancelled() builds its Notified too late and can miss a one-shot cancel

**Severity:** high  ·  **Category:** correctness  ·  **Domain:** rust

**Location:** `packages/tuff-native/native-core/src/cancellation.rs`:70

## What

The wait loop reads `reason()` first and only then constructs `notified()`. Because tokio's Notified snapshots the notify_waiters counter at construction and cancellation fires exactly once, a cancel landing in that window is lost permanently.

## Evidence

```
    pub async fn cancelled(&self) -> CancelReason {
        loop {
            if let Some(reason) = self.reason() {
                return reason;
            }
            self.state.notify.notified().await;
        }
    }
// cancel() notifies exactly once and never again:
        self.state.reason.store(reason_to_u8(reason), Ordering::Release);
        self.state.notify.notify_waiters();
```

An idle `frames` stream (static content, no new frames arriving) is torn down: begin_dispose() calls credit.cancel(Dispose) -> CancellationToken::cancel -> notify_waiters(). The stream task sits in `run_frame_stream` awaiting `LatestFrameSlot::next_frame`, whose `tokio::select!` arm is `cancellation.cancelled()`. That task had already read reason()==None and constructs its Notified after notify_waiters() ran, so the snapshot matches and it parks forever. Nothing publishes to the slot and nothing closes it, so the stream never terminates, `finish_dispose` returns NATIVE_DISPOSE_TIMEOUT, and the entry leaks in NativeRuntime::streams. Secondary hazard: `cancel()` stores `cancelled` before `reason`, so a concurrent `reason()` can observe cancelled==true with reason==REASON_NONE and wrongly report None.

## Suggested direction

Build the Notified before checking reason (pin + `.enable()`), and store the reason before flipping the `cancelled` flag (or pack both into a single atomic).

---

TalexDreamSoul:
Closes #840

## What was wrong

`cancelled()` read `reason()` first and only then constructed its `Notified`:

```rust
loop {
    if let Some(reason) = self.reason() { return reason; }
    self.state.notify.notified().await;   // built after the read
}
```

tokio snapshots the `notify_waiters` counter when a `Notified` is **constructed**, and `cancel()` notifies exactly once. A cancel landing between the read and the construction is therefore already accounted for by the snapshot: the future registers as a waiter and parks, with no wakeup left to come.

`cancelled()` is awaited in `tokio::select!` arms across the screenshot backend (`backend/mod.rs:332`, `macos/actor.rs:278`, `macos/ax_actor.rs:109/129`, `xcap.rs:246/265`), so a lost cancel there leaves the stream entry in `NativeRuntime::streams` until `finish_dispose` gives up with `NATIVE_DISPOSE_TIMEOUT`.

The same crate already gets this right in `native-core/src/stream.rs:53`, `runtime.rs:597` and `backend/mod.rs:314` — all build the `Notified` first.

## Second half

`cancel()` flipped a separate `cancelled: AtomicBool` **before** storing the reason, so a reader could observe `is_cancelled() == true` with `reason() == None`. `reason` now doubles as the flag — `REASON_NONE` means live — so one compare-exchange publishes both. `is_cancelled()`, `reason()`, `cancel()` keep their signatures and semantics; only the internal representation changed.

## Test

The window is a few instructions wide, so the test hammers it: a barrier releases **eight** waiters and the canceller together so they all race the same cancel, and a timeout turns a parked waiter into a failure rather than a hung suite.

Calibrated against the defect deliberately restored:

| configuration | detection |
|---|---|
| 8 waiters/iteration, 20k iterations | **6 of 6 runs**, latest at iteration 4268 |
| 1 waiter/iteration, 20k iterations | 5 of 6 runs — **missed once** |

So the 20k bound carries roughly a 5x margin, and the concurrency is load-bearing rather than decorative. Runtime on the fixed code: 1.5s, stable across 5 runs.

Mutations, each applied through a patch asserting it matched exactly once:

| mutation | result |
|---|---|
| `Notified` built after the state read | the race test parks and fails |
| drop the first-cancel-wins gate | the pre-existing reason test fails |
| `reason()` always `Some` | both tests fail |

## Not covered by a test

The second half has none, deliberately. Its window is two adjacent stores; a tight-spin observer on its own OS thread — 2,000 rounds x 200,000 loads, with the defect restored — never landed inside it. The single atomic **removes** the intermediate state rather than narrowing it, which is a construction-level guarantee, not an observable one. Claiming a test for it would be claiming coverage that does not exist.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` and `cargo test --workspace` all pass.

### DIFF
--- a/packages/tuff-native/native-core/src/cancellation.rs
+++ b/packages/tuff-native/native-core/src/cancellation.rs
@@ -169,6 +169,30 @@ mod tests {
         }
     }
 
+    /// The control for the test above: it only requires that `cancelled()` *resolves*, and
+    /// a "return immediately" implementation satisfies that too. A live token has to keep
+    /// it pending, or the race test would be measuring nothing.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn cancelled_stays_pending_while_the_token_is_live() {
+        let token = CancellationToken::new();
+        let waiter = token.clone();
+
+        assert!(
+            tokio::time::timeout(Duration::from_millis(250), waiter.cancelled())
+                .await
+                .is_err(),
+            "cancelled() resolved on a token that was never cancelled"
+        );
+
+        token.cancel(CancelReason::Caller);
+        assert_eq!(
+            tokio::time::timeout(Duration::from_secs(5), token.cancelled())
+                .await
+                .expect("cancelled() must resolve once the token is cancelled"),
+            CancelReason::Caller
+        );
+    }
+
     // #840's second half -- `cancel` flipping a separate `cancelled` bool before storing
     // the reason, so a reader could see a cancelled token reporting `None` -- has no test
     // here on purpose. The window is two adjacent stores; a tight-spin observer on its own
--- a/packages/tuff-native/native-core/src/cancellation.rs
+++ b/packages/tuff-native/native-core/src/cancellation.rs
@@ -1,5 +1,5 @@
 use std::sync::Arc;
-use std::sync::atomic::{AtomicBool, AtomicU8, Ordering};
+use std::sync::atomic::{AtomicU8, Ordering};
 
 use tokio::sync::Notify;
 
@@ -11,9 +11,12 @@ const REASON_CONSUMER_CLOSED: u8 = 2;
 const REASON_DEADLINE: u8 = 3;
 const REASON_DISPOSE: u8 = 4;
 
+/// `reason` doubles as the cancelled flag: `REASON_NONE` means live. Keeping them in
+/// one atomic is what makes "cancelled" and "which reason" indivisible -- with a
+/// separate bool, a reader could see the flag set while the reason was still
+/// `REASON_NONE` and report `None` from an already-cancelled token.
 #[derive(Debug)]
 struct CancellationState {
-    cancelled: AtomicBool,
     reason: AtomicU8,
     notify: Notify,
 }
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
补一条测试并说明原因。

我在给 #839 跑变异时发现同一个洞也在这个 PR 上:把 `cancelled()` 改成**从不 await、直接返回 Dispose**,race 那条测试 **3/3 全绿** —— 因为它只要求 `cancelled()` 能 resolve,不要求它在 token 存活时保持 pending。

已补 `cancelled_stays_pending_while_the_token_is_live` 作为对照。四个变异现在各自被杀死:

| 变异 | 失败的测试 |
|---|---|
| `Notified` 建在状态读之后 | race |
| 去掉首次取消的门 | 既有的 reason 测试 |
| `reason()` 恒为 Some | 全部三条 |
| `cancelled()` 从不等待 | 新增的对照 |

21 passed;`fmt --check`、`clippy -D warnings`、`cargo test --workspace` 全过。

TalexDreamSoul:
撤回:重复劳动。PR #1535 / #1536 在我动手期间(04:29)合并,修复实质相同,且用 `tokio::pin!` + `.enable()` 比我的写法更显式。我这条的 pre-flight 是在本轮开头跑的,那时它们还不存在 —— 这是本会话第三次撞上并发合并,pre-flight 只在开头做一次不够。

本 PR 里唯一没有被覆盖到的东西是那条**经过标定的 race 测试**,我会另开一条 test-only 的 PR 提出来,不带任何实现改动。

PR #1537, opened 2026-08-11
fix(screenshot): build the stop Notified before reading the flag

### LINKED ISSUE #839: [audit/rust] MacFrameSource::stop() builds its Notified after the flag check and can hang forever

TalexDreamSoul:
Closes #839

Same defect shape as #840, different file.

## What was wrong

```rust
while !control.stopped.load(Ordering::Acquire) {
    control.stop_notify.notified().await;   // built after the load
}
```

tokio snapshots the `notify_waiters` counter when a `Notified` is **constructed**, and `stop_frames` notifies exactly once (`actor.rs:889`). A stop landing between the load and the construction is therefore already accounted for by the snapshot: the future registers as a waiter and parks, with no wakeup left to come.

`ScreenshotCapability::invoke_stream` awaits this right after `request_stop()`, so a lost notify means the frames operation never emits its terminal frame, its entry stays in `NativeRuntime::streams`, and `finish_dispose` eventually fails with `NATIVE_DISPOSE_TIMEOUT`.

`native-core/src/stream.rs:53` and `backend/mod.rs:314` already build theirs first.

## Tests

| test | covers |
|---|---|
| `a_stop_landing_after_the_flag_read_is_not_lost` | the race — a barrier releases 8 waiters and the actor-side store together |
| `stop_stays_pending_until_the_actor_reports_stopped` | the control — `stop()` must not resolve early |

The second one exists because the race test alone is satisfied by a `stop()` that returns immediately. `request_stop` is idempotent (`stop_requested.swap`), so the eight `stop()` calls queue one command rather than eight.

Calibrated against the defect deliberately restored: **6 of 6 runs** caught it, latest at iteration **2916** against the 20k bound — roughly 7x margin.

Mutations, each applied through a patch asserting it matched exactly once:

| mutation | result |
|---|---|
| `Notified` built after the flag read | race test parks and fails |
| resolve without ever awaiting | control test fails |
| `request_stop` marks it stopped itself | control test fails |

## Platform note

`backend/macos/actor.rs` is `#[cfg(target_os = "macos")]`, so these tests only compile and run on the macOS leg of `native-protocol.yml`. That leg is green and has no `continue-on-error`, so the gate is real — but a Linux/Windows-only run will not exercise this.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` and `cargo test --workspace` all pass.

### DIFF
--- a/packages/tuff-native/native-screenshot/src/backend/macos/actor.rs
+++ b/packages/tuff-native/native-screenshot/src/backend/macos/actor.rs
@@ -164,10 +164,19 @@ impl BackendFrameSource for MacFrameSource {
         self.request_stop();
         let control = Arc::clone(&self.control);
         Box::pin(async move {
-            while !control.stopped.load(Ordering::Acquire) {
-                control.stop_notify.notified().await;
+            loop {
+                // Built before the flag is read, not after. `Notified` snapshots the
+                // notify_waiters counter at construction and the actor notifies exactly
+                // once in stop_frames, so a stop landing between the load and the await
+                // is accounted for by this future rather than lost -- it resolves instead
+                // of parking with no wakeup left to come. Same shape as
+                // native-core/src/stream.rs:53 and backend/mod.rs:314.
+                let notified = control.stop_notify.notified();
+                if control.stopped.load(Ordering::Acquire) {
+                    return Ok(());
+                }
+                notified.await;
             }
-            Ok(())
         })
     }
 }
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
撤回:重复劳动。PR #1535 / #1536 在我动手期间(04:29)合并,修复实质相同,且用 `tokio::pin!` + `.enable()` 比我的写法更显式。我这条的 pre-flight 是在本轮开头跑的,那时它们还不存在 —— 这是本会话第三次撞上并发合并,pre-flight 只在开头做一次不够。

本 PR 里唯一没有被覆盖到的东西是那条**经过标定的 race 测试**,我会另开一条 test-only 的 PR 提出来,不带任何实现改动。

PR #1544, opened 2026-08-11
test(native): make the #839/#840 race tests actually reach the race [#1541]

### LINKED ISSUE #1541: The #839/#840 fixes are correct but untested: restoring either defect leaves the suite green
Both fixes merged (#1535, #1536) are correct. Neither is covered by a test that fails on the defect it fixed.

## Evidence

**#840.** Restore the pre-#1535 ordering in `cancelled()` — read the state, then build the `Notified` — and run the five tests that PR added:

```
run1: test result: ok. 5 passed; 0 failed
run2: test result: ok. 5 passed; 0 failed
run3: test result: ok. 5 passed; 0 failed
```

The five tests cover the paths *around* the window, not the window: a waiter that is already parked, a waiter starting after the cancel, an uncancelled token staying pending, and the reason/flag pair. None of them races a cancel against a waiter's registration, which is the entire defect.

**#839.** `#1536` added 23 lines and **zero** tests, so the same is true by inspection.

## Why it is worth closing

The window is a few instructions wide and is exactly the kind of thing a later refactor reopens by accident — reordering two lines in `cancelled()` restores it silently. Both call sites feed dispose paths that fail with `NATIVE_DISPOSE_TIMEOUT` rather than loudly, so a regression would show up as a leaked stream entry, not a failing test.

## What closes it

A hammered race test per site. Measured on my branch against the defect deliberately restored:

| site | configuration | detection |
|---|---|---|
| `cancelled()` | 8 waiters/iteration, 20k iterations | 6 of 6 runs, latest at iteration 4268 |
| `cancelled()` | 1 waiter/iteration, 20k iterations | 5 of 6 — **missed once** |
| `MacFrameSource::stop()` | 8 waiters/iteration, 20k iterations | 6 of 6 runs, latest at iteration 2916 |

A barrier releases the waiters and the canceller/actor-side store together; a timeout turns a parked waiter into a failure rather than a hung suite. Runtime is ~1.5s per test. The eight-way concurrency is load-bearing, not decorative — one waiter per iteration measurably misses.

Note for whoever picks this up: `MacFrameSource::stop()` also needs a control test asserting it stays pending until the actor reports `stopped`. Without it, a `stop()` mutated to return immediately passes the race test — verified.

## Not in scope

The second half of #840 (the torn `cancelled`/`reason` pair) stays untested on purpose. Its window is two adjacent stores; a tight-spin observer on its own OS thread — 2,000 rounds x 200,000 loads, defect restored — never landed inside it. The single atomic removes the state rather than narrowing it, which is a construction-level guarantee.

TalexDreamSoul:
Fixes #1541. **You were right and I was wrong**: I claimed on both issues that a stress test could not reach these windows and that proving them needed `loom`. The test was under-powered, not impossible.

## Measured, each defect deliberately restored

| site | configuration | detection |
|---|---|---|
| `cancelled()` | 8 waiters × 60k iterations | **6/6** |
| `await_stopped()` | 8 waiters × 60k iterations | **6/6** |

For contrast, on this machine:

| configuration | detection |
|---|---|
| 1 waiter × 2,000 (what I shipped) | 0/3 |
| 8 waiters × 20k, canceller on the test task | 1/3 |
| 8 waiters × 20k, canceller as a task | 2/6 |
| 8 waiters × 60k | **6/6** |

## Two things that turned out to be load-bearing

**The barrier has to be a barrier.** I tried replacing it with a spin gate — an `AtomicBool` every side busy-waits on — expecting a tighter start line. It is **worse: 0/6**. The canceller crosses before the waiters can even enter `cancelled()`, so every waiter sees the reason already set and the window is never entered. The scheduler jitter a `Barrier` introduces is what spreads arrivals *across* the cancel instant.

**The iteration count, more than the concurrency.** 8×20k was 2/6 here; 8×60k is 6/6. Your note that one waiter per iteration measurably misses matches what I saw at the low end.

## #839 needed a seam first

`MacFrameSource` needs a live ScreenCaptureKit session; `FrameControl` does not. The wait is extracted as `await_stopped(Arc<FrameControl>)` and `stop()` is now `self.request_stop(); await_stopped(...)`, so the race is reachable from a unit test.

It also gets the control test you asked for, and it is not theoretical — I verified the exact failure mode you predicted:

```
await_stopped mutated to `return Ok(())` immediately:
  a_stop_racing_registration_is_never_lost ................ ok      ← passes
  await_stopped_stays_pending_until_the_actor_reports_... . FAILED  ← catches it
```

## Not in scope, agreed

The torn `cancelled`/`reason` pair stays untested by construction. One atomic removes the state rather than narrowing it, and your tight-spin observer result matches mine.

## Cost

~5s added to the native suite (2.6s per race test). `cargo fmt --check`, `clippy --workspace -D warnings`, all 7 crates green.

### DIFF
--- a/packages/tuff-native/native-core/src/cancellation.rs
+++ b/packages/tuff-native/native-core/src/cancellation.rs
@@ -132,23 +132,7 @@ mod tests {
 
     /// A cancel delivered to an already-waiting task completes it.
     ///
-    /// **This does not reproduce the race in #840, and is not claimed to.** That window is the few
-    /// instructions between reading the reason and registering the waker; I ran this at 2,000
-    /// iterations against the unfixed order and it passed every time, so shipping it as a
-    /// regression guard would have been a test that cannot fail.
-    ///
-    /// What establishes the fix is structural: `notified()` is created and enabled *before* the
-    /// state is read, so `notify_waiters()` either wakes a future that is already registered, or
-    /// the read that follows observes the reason. Neither order can drop the wakeup. Under the
-    /// previous order a cancel landing in that window was lost, and because `cancel()` fires
-    /// exactly once the waiter parked forever — an idle `frames` stream torn down by
-    /// `begin_dispose` never terminated and `finish_dispose` returned NATIVE_DISPOSE_TIMEOUT.
-    ///
-    /// Proving the absence of that interleaving needs `loom`, which is a dependency decision rather
-    /// than something to add alongside a fix.
-    ///
-    /// What this *does* cover is the ordinary path, with a timeout so a regression that parks the
-    /// waiter fails the suite instead of hanging it.
+    /// The ordinary path, kept as a cheap control alongside the hammered race test below.
     #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
     async fn cancel_wakes_a_task_already_waiting() {
         for _ in 0..256 {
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
**你是对的,但这个 PR 需要先重建基线 —— 它的分支不含已合并的 #1542,直接合会重复或冲突。**

### 我复核了你的核心主张

`await_stopped` 的抽取是行为保持的:`stop()` 变成 `self.request_stop(); await_stopped(control)`,循环体逐字未变,只是把等待搬到自由函数以便单测能构造 `FrameControl`。这个 seam 是对的 —— 我上一轮为了绕开它,在测试里手工组装了 `MacFrameSource` + `ActorInner`,你这个做法更干净。

关于迭代数,我在本机(14 核)重测了 8×20k,**整套跑**而非过滤单跑:

```
run1..run6: 全部命中,最晚第 1167 轮 (6/6)
```

所以我的 6/6 在这台机器上是真的。但你那边同样配置只有 2/6 —— 这恰好说明 **20k 是机器相关的**,而 CI runner 通常 2–4 核,更接近你的环境而不是我的。你抬到 60k 是更稳的选择,我不反对。

### 需要处理的冲突

`test/1541-race-coverage`(#1542)已于 04:58 合并,加了三条测试;你这个分支的基线在那之前,所以:

| 我合并的 | 你这个 PR 的 |
|---|---|
| `a_cancel_racing_a_waiters_registration_is_not_lost` | `a_cancel_racing_registration_is_never_lost` |
| `a_stop_landing_after_the_flag_read_is_not_lost` | `a_stop_racing_registration_is_never_lost` |
| `stop_stays_pending_until_the_actor_reports_stopped` | `await_stopped_stays_pending_until_the_actor_reports_stopped` |

功能等价,两套都留下只会让 native 套件白白多跑一遍。**建议:重建基线到当前 `app-shell-v2`,删掉我那三条,保留你的 60k 版本**(加上 `await_stopped` 的 seam)。理由就是上面那条——60k 对低核数环境更稳。

顺带确认一件我们结论一致的事:torn `cancelled`/`reason` 那半按构造保证、不写测试。我的紧自旋观察者(2000 轮 × 20 万次 load)同样一次没落进去。

Issue #851, opened 2026-08-06
[audit/rust] request_stop silently orphans an SCStream when the macOS actor mailbox is full


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** medium  ·  **Category:** resource-leak  ·  **Domain:** rust

**Location:** `packages/tuff-native/native-screenshot/src/backend/macos/actor.rs`:157

## What

`MacFrameSource::request_stop` posts `Command::StopFrames` with `try_send`. On a full 16-slot mailbox it just marks the control as stopped and returns, so the ActorState entry and its live SCStream session are never torn down.

## Evidence

```
        if self.actor.sender.try_send(command).is_err() {
            self.control.stopped.store(true, Ordering::Release);
            self.control.stop_notify.notify_waiters();
        }
```

## Why it matters / failure scenario

The single actor thread is blocked inside `wait_for_image` (up to SYSTEM_CALLBACK_TIMEOUT = 5 s) while ACTOR_QUEUE_CAPACITY (16) refresh/capture commands queue behind it. A frames consumer disconnects and calls request_stop; try_send returns TrySendError::Full, the caller is told the stream stopped, but `ActorState::streams` still holds the MacStreamSession. ScreenCaptureKit keeps capturing that display or window at the configured fps until the entire actor shuts down.

## Suggested direction

Retry the StopFrames command with a bounded blocking send, or keep a separate stop queue the actor drains, rather than reporting success on a dropped command.

---
<sub>audit-domain: rust · confidence: confirmed</sub>

TalexDreamSoul:
## 前提确认成立(源码级),修法需要一个设计决定

### 缺陷路径

`actor.rs:148-161`:

```rust
if self.actor.sender.try_send(command).is_err() {
    self.control.stopped.store(true, Ordering::Release);
    self.control.stop_notify.notify_waiters();
}
```

`try_send` 失败(邮箱满)时,代码把 `stopped` 置真并唤醒等待者 —— 于是 `stop()`(163-172,循环等 `stopped`)**正常 resolve,调用方认为流已停止**。

但真正拆除原生流的 `state.stop_frames(stream_id, &control)` 只在 actor 循环收到 `Command::StopFrames` 时执行(341-343)。命令没投递进去,**这一步就永远不会发生**。

### 兜底只有一处,而且太晚

[...]

PR #1548, opened 2026-08-11
fix(native): repair ActorInner construction in the #839/#840 race tests

TalexDreamSoul:
**`TalexDreamSoul/app-shell-v2` does not currently compile.** `cargo test --workspace` under `packages/tuff-native` fails with two `E0063`s:

```
error[E0063]: missing field `pending_stops` in initializer of `actor::ActorInner`
    --> native-screenshot/src/backend/macos/actor.rs:1861:34
    --> native-screenshot/src/backend/macos/actor.rs:1918:29
```

Every PR that touches `packages/tuff-native/**` will show a red `Protocol (*)` until this lands.

## How it happened

Neither PR was wrong:

- **#1544** added the two race tests that construct `ActorInner`.
- **#1545** added the `pending_stops` field to `ActorInner`.

Each was green against a base that did not yet contain the other. Git saw no conflict — the edits do not touch the same *lines*, only the same *struct* — so squash-merging both produced a base neither had been tested against.

That is my mistake in the merge, not theirs. I merged both with `--admin` minutes apart, so #1545's checks had already run by the time #1544 landed and nothing re-ran against the combined tree.

## The fix

`pending_stops: Mutex::new(Vec::new())` at both sites — the same initialiser the production constructor and the #851 tests already use.

Verified: `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`, all 7 crates green.

### DIFF
--- a/packages/tuff-native/native-screenshot/src/backend/macos/actor.rs
+++ b/packages/tuff-native/native-screenshot/src/backend/macos/actor.rs
@@ -1861,6 +1861,7 @@ mod tests {
             let actor = Arc::new(ActorInner {
                 sender,
                 closed: AtomicBool::new(false),
+                pending_stops: Mutex::new(Vec::new()),
             });
             let control = Arc::new(FrameControl {
                 slot: LatestFrameSlot::new(),
@@ -1918,6 +1919,7 @@ mod tests {
             actor: Arc::new(ActorInner {
                 sender,
                 closed: AtomicBool::new(false),
+                pending_stops: Mutex::new(Vec::new()),
             }),
             control: Arc::clone(&control),
         };


TalexDreamSoul:
Superseded by #1547, which landed the same repair a better way and is already merged.

Mine added `pending_stops` to the two constructors; #1547 removed those tests entirely, because they were **duplicates** of #1544's — and #1544's are the ones worth keeping, since they reach the wait through the `await_stopped` seam instead of constructing `ActorInner` literally. That is precisely why they did not break when the struct grew a field.

So the right fix was deleting the brittle pair, not patching it. Confirmed the base compiles again: all 7 crates green on `f7c245a8d`.

Closing.

Issue #1552, opened 2026-08-11
[audit/main-process] Global dictation teardown during an in-flight beginCapture() orphans the native session


TalexDreamSoul:
> Found while working #842. This is new as of today — #1549 introduced it while fixing #841, and it is the concrete path that makes #842's leak reachable rather than theoretical.

**Location:** `apps/core-app/src/main/modules/voice/global-dictation.ts`:67

## What

`beginCapture()` became `async` in #1549. `GlobalDictationController.toggle()` now does:

```ts
this.activeSessionId = await voiceService.beginCapture()
```

`unregister()` is the only teardown path, and it cancels through the same field:

```ts
if (this.activeSessionId) {
  voiceService.abortCapture(this.activeSessionId)
  this.activeSessionId = null
}
```

While the `await` is in flight, `this.activeSessionId` is still `null`. A teardown in that window aborts nothing, and the assignment then lands on a controller nobody will toggle again.

## Why it matters

The window is not theoretical: #1549 measured `startCapture` resolving at **162 ms**, and the issue's own scenario (a Bluetooth input reconnecting, or the first-use microphone consent sheet) stretches that to seconds — during which the consent sheet is waiting on a human.

The native session that results is unreachable. Its id existed only in that discarded field, `SESSIONS` is keyed by id with no enumeration or purge API, and every removal path (`stopCapture`, `cancelCapture`) needs the id. The capture thread runs to its max-duration cap and exits, but the `SessionHandle` keeps the interleaved `f32` buffer (15 s × 48 kHz × 2 ch × 4 B ≈ 5.8 MB) plus the snapshot PCM cache for the life of the process.

Before #1549 the same field was assigned synchronously, so this window did not exist.

## Suggested direction

Have `unregister()` mark the controller torn down, and have `toggle()` abort immediately if it resolves into a torn-down controller — the same shape as the `busy` guard already there. The native-side backstop is #842; this is the half that keeps the id from being dropped in the first place.

TalexDreamSoul:
Fixed in #1554. `unregister()` now marks the controller disposed, and a start that resolves into a disposed controller aborts its own session rather than storing it. Both mutations — removing the guard, and leaving `unregister` without the mark — fail the new test. The native backstop is #1553.

TalexDreamSoul:
Fixed in #1554, now merged. `unregister()` marks the controller disposed and a start that resolves into a disposed controller aborts its own session. Native backstop tracked in #842 / #1553.

Issue #845, opened 2026-08-06
[audit/rust] playAudio decodes arbitrary audio synchronously on the JS thread


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** medium  ·  **Category:** perf  ·  **Domain:** rust

**Location:** `packages/tuff-native/native-audio/src/lib.rs`:326

## What

`play_audio` is a synchronous napi function that runs the full symphonia decode loop plus a downmix on the calling (Electron main) thread before returning a playback id. Nothing caps the input size.

## Evidence

```
    let (interleaved, source_rate, source_channels) =
        decode_audio(bytes.to_vec()).map_err(Error::from_reason)?;
    let mono = downmix_to_mono(&interleaved, source_channels);
```

## Why it matters / failure scenario

TTS or a notification sound hands playAudio a multi-minute MP3. `decode_audio` walks every packet and accumulates all samples into a Vec<f32> (10 min stereo 44.1 kHz ~= 212 MB) on the main thread, freezing the UI for the whole decode. The inline comment justifies decoding on the calling thread so bad audio surfaces synchronously, but nothing bounds the input length.

## Suggested direction

Cap input byte length / decoded sample count, and move the decode to the playback thread or an AsyncTask so errors surface via the returned promise.

---
<sub>audit-domain: rust · confidence: confirmed</sub>

TalexDreamSoul:
## 前提属实,但代码**明确写着这是刻意取舍** —— issue 未提及这一点

### 事实核对

`play_audio`(`native-audio/src/lib.rs:321`)确实在 `#[napi]` 同步函数里解码(`325-326`),因此发生在 JS 主线程上。

### 但它上面就有一段说明为什么

```rust
// Decode on the calling thread so bad audio surfaces synchronously; playback
// itself runs on a dedicated thread (cpal `Stream` is not Send) and continues
// after this returns.
```

也就是说:**同步解码是为了让坏音频同步报错**,而播放本身已经在专用线程上(`345-347`)。这与 #869、#866 属于同一类 —— 报告把一个**有记录的设计取舍**当成了疏漏。

### 这个取舍是否仍然站得住:一半站得住

- 「让解码失败同步暴露」是真实收益:调用方不必处理一个异步失败通道;
- 但代价**无上界** —— 解码任意用户音频的耗时不受控,主线程阻塞时长同样不受控。

[...]

PR #1553, opened 2026-08-11
fix(native): reap capture sessions nothing can reach any more [#842]

### LINKED ISSUE #842: [audit/rust] Audio capture sessions are never garbage-collected, leaking threads and PCM buffers

**Severity:** medium  ·  **Category:** resource-leak  ·  **Domain:** rust

**Location:** `packages/tuff-native/native-audio/src/lib.rs`:168

## What

`start_capture` inserts a SessionHandle into a process-global map; entries are removed only by `stop_capture`/`cancel_capture`. There is no TTL or reaper, so any session the JS side abandons keeps its sample Vec and un-joined JoinHandle alive for the life of the process.

## Evidence

```
    let handle = SessionHandle {
        stop_flag,
        cancel_flag,
        samples,
        meta,
        join_handle,
        drain_cursor,
    };
    lock(sessions()).insert(session_id.clone(), handle);
```

The renderer that started dictation is closed or reloaded, or an exception in voice-service's async generator skips the stopCapture call. The capture thread self-stops at the 15 s max-duration cap and exits, but the SessionHandle stays in SESSIONS forever holding the interleaved f32 buffer (15 s x 48 kHz x 2 ch x 4 B ~= 5.8 MB) plus a never-joined thread handle. Repeated dictation attempts accumulate and there is no API to enumerate or purge orphaned sessions.

## Suggested direction

Reap sessions whose `meta.stopped_reason` is Some once a TTL elapses, or have the capture thread drop its own entry when nothing has read it.

---

TalexDreamSoul:
Fixes #842. Concrete reachable path: #1552.

## The premise, restated as what it actually is

A session's **id is the only handle to it**: `SESSIONS` is keyed by id, `stopCapture` and `cancelCapture` both need one, and nothing enumerates the map. So the defect is not really "no TTL" — it is that a lost id makes ~5.8 MB (15 s × 48 kHz × 2 ch × 4 B) plus the snapshot cache permanently unreachable *and* unfreeable.

I could not reach that through any shipped caller when I first looked — every path removed the session on both the success and the failure branch. It **is** reachable now: #1549 made `beginCapture()` async today, and `GlobalDictationController.unregister()` still cancels through a field that is `null` for the whole ~162 ms the start is in flight. Filed as #1552; this is the native backstop, that is the half that stops the id being dropped.

## Fix

Nothing can recover the id, so recover the memory: a session that has **stopped**, and that **nobody has read since**, is not coming back. Sixty seconds after the capture thread finishes, it goes.

Two conditions keep that from becoming a bug of its own:

- **A running session is never touched**, however long it has been going. The caller may simply be recording, and `maxDurationMs` is caller-supplied.
- **Every read resets the clock.** `poll_capture`, `snapshot_capture` and `drain_capture` all mark the session, so the window measures silence from the JS side rather than elapsed time. Without that, every long collection would be a race against the TTL.

## Why at `start_capture`, not on a timer

The map only grows when a session starts, so that is exactly where accumulation happens, and it costs one pass over a map that is normally empty. No reaper thread, no timer to leak.

Dropping the `JoinHandle` detaches rather than joins, which is what we want: `stopped_at` is written on the thread's way out, so there is nothing left to wait for, and a join would be a blocking call inside the sessions lock.

## Not the other suggestion

"Have the capture thread drop its own entry" would break the normal flow: the thread self-stops on trailing silence, and `voice-service` then polls until `active: false` and calls `stopCapture` to collect the audio. Self-removal turns that into `session-not-found` on every silence-terminated dictation.

## Mutations, each caught

| mutation | fails |
|---|---|
| a running session treated as reapable | `a_running_session_is_never_abandoned`, `reaping_removes_only_the_unreachable_sessions` |
| `last_read` ignored | three tests |
| `touch_session` made a no-op | `a_read_moves_a_session_out_of_the_reaper_s_reach` |
| `stopped_at` never recorded | `a_finished_capture_records_when_it_stopped` |
| reaper drops everything | `reaping_removes_only_the_unreachable_sessions` |
| **call dropped from one of the three read paths** | `every_read_path_touches_the_session` |

The last one is the mutation the others cannot see: `touch_session` stays alive for the other two call sites, so there is no dead-code warning under `-D warnings`, and the only symptom is a session read solely through that path being reaped mid-collection. It needs a source-level test, so it has one.

## Verified

`cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`, all 7 native crates green (34 tests in `tuff-native-audio`, up from 27).

### DIFF
--- a/packages/tuff-native/native-audio/src/lib.rs
+++ b/packages/tuff-native/native-audio/src/lib.rs
@@ -180,8 +180,19 @@ fn start_capture_blocking(options: Option<AudioCaptureOptions>) -> Result<String
         meta,
         join_handle,
         drain_cursor,
+        last_read: Mutex::new(Instant::now()),
     };
-    lock(sessions()).insert(session_id.clone(), handle);
+    {
+        // Reap here rather than on a timer: the map only grows when a session is
+        // started, so this is exactly where accumulation would happen, and it costs
+        // one pass over a map that is normally empty.
+        let mut map = lock(sessions());
+        let reaped = reap_abandoned_sessions(&mut map, Instant::now(), ABANDONED_SESSION_TTL);
+        if reaped > 0 {
+            eprintln!("[tuff-native-audio] reaped {reaped} abandoned capture session(s)");
+        }
+        map.insert(session_id.clone(), handle);
+    }
 
     Ok(session_id)
 }
@@ -258,6 +269,7 @@ pub fn poll_capture(session_id: String) -> Result<AudioCaptureState> {
         let handle = guard
             .get(&session_id)
             .ok_or_else(|| Error::from_reason(format!("session-not-found: {session_id}")))?;
+        touch_session(handle);
         (handle.meta.clone(), handle.samples.clone())
     };
 
@@ -288,6 +300,7 @@ pub fn snapshot_capture(session_id: String) -> Result<AudioSnapshot> {
         let handle = guard
             .get(&session_id)
             .ok_or_else(|| Error::from_reason(format!("session-not-found: {session_id}")))?;
+        touch_session(handle);
         (
             handle.meta.clone(),
             handle.samples.clone(),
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Rebased onto `app-shell-v2` now that #1551 (#845) has landed, and re-verified against the **combined** tree rather than each half separately — the two touch the same file, and #1547 was the last time that assumption cost a broken base.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`, all 7 crates green; 37 tests in `tuff-native-audio`.

PR #1558, opened 2026-08-11
fix(tuffex): declare the @vue/reactivity its .d.ts imports

### LINKED ISSUE #1557: tuffex ships .d.ts importing @vue/reactivity, which it does not declare
Eight shipped declaration files import from `@vue/reactivity`. `packages/tuffex/package.json` does not declare it — not in `dependencies`, not in `peerDependencies`, not in `devDependencies`.

```
dist/es/select/src/TxSelect.vue.d.ts              (4 refs)
dist/es/context-menu/src/TxContextMenu.vue.d.ts   (4)
dist/es/search-select/src/TxSearchSelect.vue.d.ts (1)
dist/es/search-input/src/TxSearchInput.vue.d.ts   (1)
+ the four matching index.d.ts
```

Control: `@floating-ui/vue` is imported by a `.d.ts` too, and it **is** declared (`^1.1.11`). So the manifest is not simply missing every type dependency — this one specifically.

## Consumer impact

Reproduced against the tarball `pnpm pack` produces from this tree, in a scratch project with only `typescript` and `vue` installed:

```
node_modules/@talex-touch/tuffex/dist/es/select/src/TxSelect.vue.d.ts(134,141):
  error TS2307: Cannot find module '@vue/reactivity' or its corresponding type declarations.
```

Who hits it:

- **pnpm consumers** — strict `node_modules` means `vue`'s own dependency on `@vue/reactivity` is not reachable from the consumer's tree.
- **with `skipLibCheck: false`** — with the usual `true`, all of these vanish (verified: `TSC_RC=0`).

npm/yarn users are shielded by hoisting, which is why this has been invisible.

## Where it comes from

Not from tuffex's source. The reference sits inside an inlined conditional type:

```ts
(...args: any) => infer R ? (...args: [R, R, import('@vue/reactivity').OnCleanup]) => any : ...
```

That is Vue's own `WatchCallback` shape, emitted by vue-tsc from the module that *declares* `OnCleanup` rather than the one that re-exports it. `defineExpose` in `TxSelect.vue` exposes nothing watch-related — there is no source-level construct to rewrite.

## Fix, verified

Declaring it clears every one of those errors. In the scratch project:

```
$ pnpm add @vue/reactivity@^3.5.33 && tsc --noEmit   # skipLibCheck: false
remaining: @codemirror/state, @floating-ui/vue, marked
```

Those three are tuffex's own **declared** runtime dependencies — absent only because that scratch install was deliberately minimal. A real consumer gets them. **`@vue/reactivity` is the only undeclared one.**

### `peerDependencies` rather than `dependencies`

`vue` is already a peer (`^3.5.27`), and `@vue/reactivity` is versioned in lockstep with it. As a plain dependency, a version drift would install a *second* copy, giving two distinct reactivity type identities — worse than the problem being fixed. As a peer at the same range it stays pinned to whatever `vue` the consumer resolved, and pnpm auto-installs peers by default.

The cost is an unmet-peer warning for anyone running `auto-install-peers=false`.

## How this went unnoticed

`audit:types` exists precisely to catch this — it typechecks the published subpath declarations from an external project with `skipLibCheck: false`. It has been unable to run since `@talex-touch/utils@workspace:^` was added to tuffex (#1555), and no CI job runs it either. The gate that would have caught this was broken and unwatched.

TalexDreamSoul:
Closes #1557

Eight shipped `.d.ts` files import from `@vue/reactivity`. The manifest declared it nowhere — not `dependencies`, not `peerDependencies`, not `devDependencies`.

Control: `@floating-ui/vue` is imported by a `.d.ts` too and **is** declared, so this is one specific omission, not a manifest that lists nothing.

## Correcting my own framing

I filed this as a live breakage and **that was wrong.** A faithful `pnpm install` of the packed tarball typechecks clean:

```
$ tsc --noEmit -p tsconfig.json        # skipLibCheck: false
TSC_RC=0    errors: 0

loaded tuffex files: 421   (including .../select/src/TxSelect.vue.d.ts)
deliberate type error  -> RC=2, error TS2322        <- positive control
installed manifest: dependencies[@vue/reactivity] = None   <- the pre-fix one
```

`--traceResolution` says why:

```
Module name '@vue/reactivity' was successfully resolved to
  node_modules/.pnpm/node_modules/@vue/reactivity/dist/reactivity.d.ts
unresolved: 0
```

pnpm's default `hoist-pattern` puts it in `node_modules/.pnpm/node_modules/` (189 packages there), which sits on the lookup path from tuffex's own `.d.ts`. npm and yarn flatten for the same effect.

My earlier reproduction had extracted the tarball into a plain `node_modules/@talex-touch/tuffex` **by hand**, bypassing every package manager's layout. That is what produced the TS2307, and it is not a layout anyone ships.

**So no consumer is broken today.** The issue is relabelled from `bug` accordingly.

## Why still fix it

An undeclared coupling: what a published type surface imports should be declared, rather than left to whether the consumer's package manager happens to hoist it. Cost is 3 lockfile lines.

## Not fixable at the source

The reference sits inside an inlined conditional type carrying Vue's `WatchCallback` shape:

```ts
(...args: any) => infer R ? (...args: [R, R, import('@vue/reactivity').OnCleanup]) => any : ...
```

vue-tsc emits it from the module that *declares* `OnCleanup`, and `vue` does not re-export it:

```
probe.ts(1,15): error TS2305: Module '"vue"' has no exported member 'OnCleanup'.
```

## `dependencies`, not `peerDependencies`

I wrote it as a peer first. This repo's own `.npmrc` sets `auto-install-peers=false`, so a peer would not be installed for consumers sharing that setting.

The usual argument against a plain dependency — a second copy giving two type identities — does not apply here: **all 32 references are to `OnCleanup`**, which is `(cleanupFn: () => void) => void`, a structural alias assignable across copies. pnpm resolved it to `3.5.39`, the version `vue` already pulls, so nothing is duplicated.

## Verification

| check | result |
|---|---|
| `pnpm typecheck` (tuffex) | 0 errors |
| `pnpm test` (tuffex) | 1114 passed / 145 files |
| `pnpm lint` (tuffex) | clean |
| core-app main-process `tsc` | 0 errors |
| packed manifest | carries `"@vue/reactivity": "^3.5.27"` |

`App suites (nexus)` failed on the first run at `Upload nexus test report` — a 404 from the artifact service. `Run nexus suite` itself passed; re-run requested.

### DIFF
--- a/packages/tuffex/package.json
+++ b/packages/tuffex/package.json
@@ -85,6 +85,7 @@
     "@number-flow/vue": "^0.5.1",
     "@shikijs/engine-javascript": "^4.4.2",
     "@talex-touch/utils": "workspace:^",
+    "@vue/reactivity": "^3.5.27",
     "@vueuse/core": "^14.3.0",
     "dompurify": "^3.4.11",
     "gsap": "^3.15.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1132,6 +1132,9 @@ importers:
       '@talex-touch/utils':
         specifier: workspace:^
         version: link:../utils
+      '@vue/reactivity':
+        specifier: ^3.5.27
+        version: 3.5.39
       '@vueuse/core':
         specifier: ^14.3.0
         version: 14.3.0(vue@3.5.39(typescript@5.9.3))


TalexDreamSoul:
补一条把结论钉死:`hoist=false` 也不会断。

三种布局实测,只有「手工解包到扁平 node_modules」会报 TS2307,而那不是任何包管理器会产生的布局 —— 是我最初复现时自己造出来的。`hoist=false` 下唯一未解析的是 `@babel/types`,报错位置在 `@vue/compiler-core` 自己的 d.ts 里,与 tuffex 无关。

**所以这个 PR 不是缺陷修复,没有消费者受影响。** 剩下的唯一理由是「发布出去的类型引用了什么就该声明」,代价 3 行 lockfile。按这个理由合或不合都合理,不合我也不认为是损失 —— 详见 #1557。

PR #1579, opened 2026-08-11
test(search-engine): declare setAppExecutionRecorder in the app-provider mocks

### LINKED ISSUE #1578: search-engine suite is broken on app-shell-v2: five app-provider mocks are missing setAppExecutionRecorder
`apps/core-app/src/main/modules/box-tool/search-engine` does not run on the current base.

```
Test Files  4 failed | 70 passed (74)
Tests       25 failed | 610 passed (635)

Error: [vitest] No "setAppExecutionRecorder" export is defined on the
"../addon/apps/app-provider" mock. Did you forget to return it from "vi.mock"?
  ❯ src/main/modules/box-tool/search-engine/search-core.ts:2208:1
```

## What happened

`app-provider.ts:94` exports `setAppExecutionRecorder`, and `search-core.ts:39` imports it:

```ts
import { appProvider, setAppExecutionRecorder } from '../addon/apps/app-provider'
```

A factory `vi.mock` replaces the **whole** module, so it has to carry every binding the importer names. Five test files mock that path and **none** of them declares it:

```
search-core.contracts.test.ts          0
app-indexed-source.test.ts             0
search-core.trace.test.ts              0
search-core.gather-ordering.test.ts    0
search-core.regression-baseline.test.ts 0
```

The failure is at import time, before any test body runs — which is why whole suites fail rather than individual assertions.

## Why it landed

`App suites (core-app)` runs with `continue-on-error: ${{ matrix.app != 'nexus' }}`. **A broken core-app suite reports `conclusion: success` and cannot fail a PR.** So an export can be added, its importer updated, and every mock left stale, with nothing in CI objecting.

That is the same blind spot as #1541 (tests that could not fail) and the `SNAP_COUNT`/`APP_COUNT` pair — a control whose failure mode is silence.

## Noticed by accident

I ran this suite alone while verifying a Trellis task. Earlier in the same session I had run it *together with two other directories* and it reported 1155 passed — the base had moved between the two runs. Worth knowing that a green three-suite run is not evidence about a suite you did not isolate.

Fix in the linked PR: declare the binding in all five mocks. Verified by mutation — removing it from one file reproduces the exact error.

TalexDreamSoul:
Closes #1578. **The search-engine suite does not run on this base.**

```
Test Files  4 failed | 70 passed (74)
Tests      25 failed | 610 passed (635)

Error: [vitest] No "setAppExecutionRecorder" export is defined on the
"../addon/apps/app-provider" mock.
  ❯ search-core.ts:2208:1
```

`app-provider.ts:94` exports it, `search-core.ts:39` imports it. A factory `vi.mock` replaces the **whole** module, so it must carry every binding the importer names — and all five files that mock that path declared none. The throw is at import time, before any test body runs, which is why whole files fail rather than assertions.

## After

```
Test Files  74 passed (74)
Tests      646 passed (646)
```

Mutation-checked rather than assumed: removing the declaration from `search-core.gather-ordering.test.ts` reproduces the exact error; restoring it passes.

## Why it could land

`App suites (core-app)` runs with `continue-on-error: ${{ matrix.app != 'nexus' }}`. **A broken core-app suite reports `conclusion: success` and cannot fail a PR.** So an export can be added, its importer updated, and every mock left stale, with nothing in CI objecting.

Same shape as #1541 (tests that could not fail) and the `SNAP_COUNT`/`APP_COUNT` pair — a control whose failure mode is silence.

## How it surfaced

I ran this suite alone while verifying a Trellis task for #1125. Earlier in the same session I had run it *together with two other directories* and got 1155 passed — the base moved between the two runs. A green multi-directory run is not evidence about a suite you did not isolate.

### DIFF
--- a/apps/core-app/src/main/modules/box-tool/search-engine/app-indexed-source.test.ts
+++ b/apps/core-app/src/main/modules/box-tool/search-engine/app-indexed-source.test.ts
@@ -15,6 +15,9 @@ const appProviderMock = vi.hoisted(() => ({
 }))
 
 vi.mock('../addon/apps/app-provider', () => ({
+  // search-core.ts imports this alongside appProvider; a factory mock has to carry every
+  // binding the importer names or vitest throws at import time, before any test runs.
+  setAppExecutionRecorder: vi.fn(),
   appProvider: appProviderMock
 }))
 
--- a/apps/core-app/src/main/modules/box-tool/search-engine/search-core.contracts.test.ts
+++ b/apps/core-app/src/main/modules/box-tool/search-engine/search-core.contracts.test.ts
@@ -150,6 +150,9 @@ vi.mock('../../storage', () => ({
   }
 }))
 vi.mock('../addon/apps/app-provider', () => ({
+  // search-core.ts imports this alongside appProvider; a factory mock has to carry every
+  // binding the importer names or vitest throws at import time, before any test runs.
+  setAppExecutionRecorder: vi.fn(),
   appProvider: {
     id: 'app-provider',
     onSearch: vi.fn(),
--- a/apps/core-app/src/main/modules/box-tool/search-engine/search-core.gather-ordering.test.ts
+++ b/apps/core-app/src/main/modules/box-tool/search-engine/search-core.gather-ordering.test.ts
@@ -145,6 +145,9 @@ vi.mock('../../storage', () => ({
 }))
 
 vi.mock('../addon/apps/app-provider', () => ({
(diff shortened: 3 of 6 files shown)


TalexDreamSoul:
见 #1578 的撤回说明。**前提为假:干净 origin 上套件没有坏**,我把并发 agent 的在途改动当成了基线。

这个分支**不删**:它保存着他们未提交的四个文件(`app-provider.ts`、`app-provider.test.ts`、`app-provider-test-harness.ts`、`search-core.ts`),需要时可从这里取回。

Issue #959, opened 2026-08-06
[audit] Remediation tracker — rolling progress (automated loop)


TalexDreamSoul:
> Rolling progress tracker for the automated **audit remediation loop** working through the 454 governance-audit findings (#484–#958, label `audit`). This is the single coordination issue; per-issue progress is posted as comments on each issue.

## How the loop works
- Fires every ~10 min (durable cron), auto-expires in 7 days; resumes remaining work by querying `gh issue list --label audit --state open`.
- **Isolation:** every fix is made in a dedicated `git worktree` checked out clean from `origin/TalexDreamSoul/app-shell-v2` — never touching the maintainer's in-progress working tree or concurrent sessions.
- **Per issue:** branch `audit-fix/<n>-<slug>` → minimal verified fix → commit → push → PR (`Fixes #<n>`, base `TalexDreamSoul/app-shell-v2`) → progress comment on the issue.
- **Close policy:** low-risk / mechanical issues (docs, i18n keys, dead config, small correctness) are fixed, PR'd, and the issue closed with a pointer to the PR. **No auto-merge** to the shared branch.
- **High-risk issues** (RCE chain, DB migration, native async, auth) get an **investigation + concrete remediation-plan comment and stay open** for human review — they are not auto-closed.

## Progress
Updated each iteration below.

- Loop armed (cron `adeda32a`). Starting with low-risk, high-confidence issues.

TalexDreamSoul:
**Iteration 1**
- ✅ #604 + #605 — README license claim MIT → MPL-2.0 (EN + zh-CN). PR #960 (base `app-shell-v2`), both issues closed.

Method validated end-to-end: isolated worktree off `origin` → minimal verified diff (2 lines) → branch → PR → per-issue comments → close. Continuing with the next low-risk batch.

TalexDreamSoul:
**Iteration 2**
- ✅ #606 + #625 — CLAUDE.md: fixed 7 non-existent `core:build*` command docs → real `build*` scripts; utils version 1.0.23 → 1.0.50. PR #961, both closed.
Running total: 4 issues resolved across 2 PRs (#960, #961).

TalexDreamSoul:
**Iteration 2 (cont.)**
- ✅ #623 — nexus README dev port 3000 → 3200. PR #962, closed.
Running total: **5 issues resolved** across 3 PRs (#960, #961, #962).

TalexDreamSoul:
**Iteration 3**
[...]

PR #1603, opened 2026-08-11
fix(core-box): stub the renderer URL instead of assigning it

TalexDreamSoul:
**`TalexDreamSoul/app-shell-v2` does not typecheck right now.** Every PR opened against it inherits this.

```
meta-overlay.test.ts(137,12) error TS2704: The operand of a 'delete' operator cannot be a read-only property
meta-overlay.test.ts(139,17) error TS2540: Cannot assign to 'ELECTRON_RENDERER_URL' because it is read-only
meta-overlay.test.ts(144,15) error TS2540: same
```

## How it landed

`electron-vite` declares `ELECTRON_RENDERER_URL` readonly, and every other reader in the tree only *reads* it. #1600 was the first to write one, to give `init()` a URL to resolve.

**The tests pass either way** — which is exactly why it got through: `App suites (core-app)` is `continue-on-error`, so nothing in that PR reported it, and `Typecheck (workspace)` only sees the tree after the merge.

## The fix

`vi.stubEnv` — already the repo's idiom for this (`renderer-override-verification.test.ts`, `screenshot-service.test.ts`). It typechecks, and `vi.unstubAllEnvs` restores the value, which is what the hand-rolled save/restore around the assignment existed for, so that goes too.

Typecheck clean, the six meta-overlay tests pass, eslint clean.

## Note

Not my change and not a criticism of it — #1600 fixed two genuinely broken tests. This is the same class as #1547 earlier today: a PR that is green on its own breaks the base, because the job that would have caught it cannot fail a PR.

### DIFF
--- a/apps/core-app/src/main/modules/box-tool/core-box/meta-overlay.test.ts
+++ b/apps/core-app/src/main/modules/box-tool/core-box/meta-overlay.test.ts
@@ -130,18 +130,16 @@ const item = {
 // init() resolves the CoreBox renderer URL, and electron.app is mocked as unpackaged above, so it
 // takes the development branch and throws without one. Introduced by #1465, which gave app-profile
 // views the navigation guards plugin views have; this test was not updated with it.
-const previousRendererUrl = process.env.ELECTRON_RENDERER_URL
-
+// Set through vi.stubEnv rather than by assignment: electron-vite declares ELECTRON_RENDERER_URL
+// readonly, so writing it fails the main-process typecheck (TS2540/TS2704) even though the test
+// itself passes. vi.unstubAllEnvs also restores it, which the hand-rolled save/restore was for.
 afterAll(() => {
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
更正我在这个 PR 描述里的一句判断。

我写的是「`App suites (core-app)` 是 continue-on-error,所以那个 PR 上没人报告」。**查证后不成立** —— #1600 合并时 `Typecheck (workspace)` 就是 **FAILURE**,闸门准确报了红,是**红着被合并**的。

数了最近 12 个已合并 PR(排除全仓皆红的 `Documentation Quality`):#1600 与 #1602 带着 `PR Quality` + `Typecheck (workspace)` 失败合入,其余 10 个干净。

所以"把 App suites 改成阻塞"这条建议我收回 —— 它解决不了红着合并。详细记在 #959。

PR #1606, opened 2026-08-11
ci(core-app): let the suite fail the job it runs in

### LINKED ISSUE #1596: core-app suite has 10 real failures that CI reports as success
`App suites (core-app)` runs with `continue-on-error: ${{ matrix.app != 'nexus' }}`, so **the step reports success whatever the tests do.** The job status carries no information about whether the suite passed.

Read from the JUnit artifact of run `31493661439` on the current base:

```
tests=5840  failures=10  errors=0
```

| file | test |
|---|---|
| `roadmap-task-links.test.ts` | ROADMAP active-task table agrees with the count in its own heading |
| `recommendation-exposure-schema.test.ts` | `recommendation_exposure_daily` applies cleanly on the full migration chain |
| `retention-migration.test.ts` | privacy retention indexes extend the real chain without deleting existing |
| `meta-overlay.test.ts` ×2 | routes renderer item actions back to the parent / prefers a non-overlay sender |
| `plugin-sqlite-worker.integration.test.ts` | worker artifact is built, or skipping was asked for explicitly |
| `ActivatedProviders.test.ts` ×2 | monochrome provider icons / explicitly colourful icon assets |

(The artifact lists 8 `<failure>` elements against a summary count of 10 — two are likely suite-level rather than case-level. Worth checking rather than assuming.)

## One of them is mine, and that is the point

`roadmap-task-links.test.ts` has been failing since **#1567**, which is my commit. It archived a task and dropped the ROADMAP row following #1155's precedent, but left the heading claiming 19 when 18 remain:

```
before  rows=19  heading=19
#1567   rows=18  heading=19   <- the failure
```

**Nothing told me.** Not the PR checks, not the merge. I have been citing this blind spot on #915, #1541 and #1578 all session, and my own change still fell into it — because the only way to see it is to download an artifact and parse XML.

## Relation to #323

#323 is *Restore the full CoreApp Vitest suite* and quotes 6 failing files / 18 failing tests from `v2.4.13`. The current measured state is **10 failures across 7 files**, so its numbers are stale rather than wrong. Posting the current list there too.

## What this issue is for

The distinction worth keeping separate:

- **#323** — fix the failing tests.
- **this** — even at zero failures, the step cannot fail a PR. The `Integration suite (packages/test)` job went through exactly this: it was continue-on-error until its baseline reached zero, then `continue-on-error is gone and this job blocks (#1255)`. The same path exists here; it needs the count at zero first.

TalexDreamSoul:
Closes #1596.

`App suites (core-app)` ran with `continue-on-error`, so **the step reported success whatever the 5840 tests did**. #1596 measured what that was hiding — 10 real failures, readable only from the JUnit artifact, because the job status carried no information.

Those ten are fixed. This removes the escape hatch.

## Verified at zero first, under this job's own environment

```
TUFF_SKIP_SQLITE_WORKER_TESTS=1 vitest run

700 files passed, 4 skipped | 5830 tests passed, 10 skipped | exit 0
```

Without that variable the SQLite worker guard fails locally, which is correct — it refuses to skip silently, and the job *declares* the skip rather than discovering it (#1604).

## The stale comment is part of the evidence

The block said:

> core-app stays non-blocking at **2 failing files**, down from 20 (#323)

and named which two. The suite was at **10**, and nothing reported it — the count in the comment had been wrong long enough to be worth pointing at. Updated rather than deleted: the history of why the hatch existed is exactly the reason to be careful about adding another.

Two other comments that reasoned from `continue-on-error` are refreshed too — the permission-mapping guard's placement (#915) and the JUnit upload's `if: always()`.

## The cost, plainly

**This makes a flake block PRs.** I saw one candidate this session: `search-core.contracts > releases first-result metrics even when telemetry is disabled` timed out at 10s once under full-directory load, and passed alone and on a re-run.

Against that: a suite that cannot fail let ten failures sit unreported, and this session a merge broke the base typecheck while its own PR showed `Typecheck (workspace)` red. The trade looks clearly worth it — but it is a trade, and if flakes start blocking, the answer is to quarantine the flaky test by name, not to restore a step that reports success unconditionally.

Only one non-comment line changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,8 +124,10 @@ jobs:
       - name: Verify search-index writer ownership
         run: pnpm -C apps/core-app check:search-index-writers && pnpm -C apps/core-app check:search-index-writers:self-test
 
-      # Runs here rather than as a core-app vitest file: the App suites job is
-      # continue-on-error, so a test there cannot fail a PR (#915).
+      # Runs here rather than as a core-app vitest file: when this was written the App suites
+      # job was continue-on-error, so a test there could not fail a PR (#915). That is no longer
+      # true (#1596), but a guard over a permission mapping belongs with the other source-level
+      # guards either way, and moving it now would only churn.
       - name: Verify permission API mapping coverage
         run: pnpm -C apps/core-app check:permission-api-mappings && pnpm -C apps/core-app check:permission-api-mappings:self-test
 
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
被 #1605 取代 —— 它在我这条合并之前落地,做了完全相同的改动:摘掉 `continue-on-error`,并同样更新了那三处从它推理出来的过期注释(#915 守卫摆放、"2 failing files" 计数、JUnit 上传的 `if: always()`)。措辞比我的更好。

**这是本会话第五次重复劳动。** 前四次是 #812/#815、#1509/#1435、#839/#840、#1541/#1544。我每轮动手前都会查一次开着的 PR,但这个 loop 的分支往往在我开工之后才出现 —— 单次 pre-flight 挡不住。

我这条唯一多做的是在提交信息里写明了代价(会让偶发失败挡 PR,并点名了我实际观察到的那个候选:`search-core.contracts > releases first-result metrics even when telemetry is disabled`,整目录并跑时超时一次、单独跑通过)。这条信息我补在 #1596 的评论里,不需要为它再开一个 PR。

Issue #594, opened 2026-08-06
[audit/compat] electron-builder macOS target is `dir` on arm64 only — no installable artifact and no Intel build


TalexDreamSoul:
> Filed by the full-repo governance audit sweep (2026-08-05). Tracked via the `audit` label.

**Severity:** medium  ·  **Category:** build  ·  **Domain:** compat

**Location:** `apps/core-app/electron-builder.yml`:120

## What

The mac target produces an unpacked directory for arm64 only, while a dmg config block and a GitHub publish provider are configured and the release workflow searches for .dmg/.zip artifacts.

## Evidence

```
apps/core-app/electron-builder.yml:120-122 `target:\n    - target: dir\n      arch: arm64` with :128 an empty `dmg:` block and :163-166 `publish: provider: github`. .github/workflows/build-and-release.yml:563 `DMG_COUNT=$(find apps/core-app/dist -name "*.dmg" -type f ...)` and :565 `ZIP_COUNT=$(find apps/core-app/dist -name "*.zip" ...)`. apps/core-app/scripts/build-target.js:501 `const defaultArch = normalizedTarget === 'mac' ? 'arm64' : 'x64'` — no target override is injected.
```

## Why it matters / failure scenario

`pnpm core:build:release:mac` (and the macOS release job) emit only `dist/mac-arm64/tuff.app`. DMG_COUNT and ZIP_COUNT are both 0, so the release publishes no macOS installer, and Intel (x64) Macs get no build at all. The `dmg:` config block and the `publish: github` provider are inert.

## Suggested direction

Set the mac target to `dmg`/`zip` and add an x64 (or universal) arch entry, or delete the dmg block and document that macOS is arm64-unpacked-only.

---
<sub>audit-domain: compat · confidence: confirmed</sub>

TalexDreamSoul:
Investigated and partially delivered in #1344 — **not closing**, because both directions you offer decide what users actually receive.

## Confirmed, but the mechanism differs from the evidence

Your conclusion is right: a release ships no macOS installer. The route there is not the one described.

The artifact **check** does *not* fail on a `dir` target. `build-and-release.yml:585` also counts `.app` directories, so `APP_COUNT > 0` sets `ARTIFACTS_FOUND=true` and the step passes. What actually drops macOS is the two steps after it:

- **Upload Artifacts** globs only `*.dmg` / `*.zip` / `*.exe` / `*.AppImage` / `*.deb` / `*.snap`
[...]

PR #1609, opened 2026-08-11
ci(release): stop counting an unpackaged .app as a macOS artifact

TalexDreamSoul:
`APP_COUNT` counts `.app` **directories** and set `ARTIFACTS_FOUND=true` on them. A directory is not something the release step can upload — it collects with `find artifacts -type f` over a fixed extension list, and nothing in this workflow repackages a `.app`: `upload-artifact` compresses for transport and `download-artifact` extracts it straight back to a directory.

## Correction to the original rationale

This PR first said the macOS leg "contributed zero release assets", following #1608. **That is wrong, and I have measured it.** The `v2.4.14-beta.2` release run reports:

```
Found 0 .dmg file(s), 5 .app directory/directories, and 1 .zip file(s)
```

That one zip is real and shippable: `apps/core-app/scripts/build-target/postprocess-mac.js:233` creates it after verifying the Developer ID signature —

```
=== Verifying Developer ID signatures before archive ===
  ✓ Verified Developer ID signature (2L5YC85FQ7): .../dist/mac-arm64/tuff.app
=== Creating zip file ===
  ✓ Created zip: .../dist/tuff-2.4.14-beta.2-arm64.app.zip
```

— and it becomes `macos-latest-beta-tuff-2.4.14-beta.2-arm64.app.zip` on the release, 245 MB, signed and notarized.

## What is still wrong, and why this change still stands

`APP_COUNT` was **5**. There is one application; the other four are helper `.app` bundles inside the framework. So the counter was reporting build intermediates as artifacts — it just was not the *only* signal, which is what made the error invisible.

The change keeps the diagnostic (renamed `APP_DIR_COUNT`) and stops it contributing to `ARTIFACTS_FOUND`. The new message prints only when nothing else was found, so it surfaces a genuinely empty macOS build instead of masking one.

## Behaviour on the current pipeline: unchanged

`ZIP_COUNT=1` already sets `ARTIFACTS_FOUND=true`, so the leg stays green and the new diagnostic stays quiet. This does not gate releases — my earlier note that it would was based on #1608's premise, which the run log above disproves.

Refs #1608 #594

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -584,22 +584,28 @@ jobs:
               ARTIFACTS_FOUND=true
             fi
           elif [ "$RUNNER_OS" = "macOS" ]; then
-            # Check for .dmg files, .app directories, and .zip files (zip contains .app)
             DMG_COUNT=$(find apps/core-app/dist -name "*.dmg" -type f 2>/dev/null | wc -l || echo "0")
-            APP_COUNT=$(find apps/core-app/dist -name "*.app" -type d -path "*/mac-*/*" 2>/dev/null | wc -l || echo "0")
             ZIP_COUNT=$(find apps/core-app/dist -name "*.zip" -type f 2>/dev/null | wc -l || echo "0")
-            echo "Found $DMG_COUNT .dmg file(s), $APP_COUNT .app directory/directories, and $ZIP_COUNT .zip file(s)"
+            # A bare .app is reported but does not count. The release step collects with
+            # `find artifacts -type f` over a fixed extension list, and nothing in this workflow
+            # repackages a .app -- upload-artifact compresses for transport and download-artifact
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
补一条合并前的具体后果,已同步到 #1608:

**这会让整个发布停下,不只是 macOS。** `create-release` 的条件是 `needs.build-and-release.result == 'success'`,而矩阵聚合结果要求所有腿都成功。mac 腿失败时,Windows/Linux 的构建仍会跑完并上传 artifact(`fail-fast: false`),但 `create-release` 不运行,任何平台都不会发布。

现成的绕行是同一个 `if` 的第二分支:`workflow_dispatch` + `sync_tag` + `recover_run_id` 可从既有 run 恢复发布。

**更自然的顺序**:先定 #594 的目标格式并落地,再合这个 PR —— 那样它从第一天起就是绿的,只在未来回退时才发声。先合也行,代价是中间发布要走手动恢复。

PR #1610, opened 2026-08-11
ci(core-app): build the main process so the SQLite sandbox tests run

TalexDreamSoul:
Closes #1607.

Six plugin SQLite sandbox cases had never run in CI. They need `out/main/plugin-sqlite-worker.js`; only `electron-vite build` produces it; this job did not run one. #1604 declared the skip rather than leaving it to be discovered — which cost nothing and gained nothing.

## The cost was the open question. Measured.

Cold, with any existing `out/` moved aside first:

```
electron-vite build                 22s
out/main/plugin-sqlite-worker.js    12,686 bytes
```

Against **~90s** for the suite itself and **~34s** for the install in this same job. And with the artifact present:

```
Tests  7 passed (7)     ← was 6 skipped + 1 guard
```

## One trap, recorded in the comment because I fell into it

**The whole `out/main` is required, not the one file.** The worker is code-split and loads siblings from `out/main/chunks` — 50 of them. Copying only `plugin-sqlite-worker.js` fails all six with

```
Cannot find module './chunks/plugin-sqlite-worker-protocol-DOuRtMyk.js'
```

which reads exactly like six real defects. I reported that to myself first and had to spend the next step ruling it out. Anyone who later caches this artifact between jobs, or trims it, will hit the same thing.

## Note on the measurement environment

Built in the main checkout, not this worktree — `electron-vite build` cannot run here (`@iconify-json/carbon` unresolvable, 33 top-level packages against 1684). That is an incomplete install in the worktree, **not** a missing declaration: the import has been there since 2025-11-12 and the 2026-08-03 release built fine. Same class as #1564; written up on #1607 so the next person does not re-diagnose it as a phantom dependency.

The build wrote only gitignored `out/`; the main checkout's dirty file count was 43 before and 43 after.

PR #1611, opened 2026-08-11
ci: run the plugin SQLite sandbox tests when that code changes

### LINKED ISSUE #1607: Build the plugin SQLite worker in CI so its six sandbox tests actually run
Split out of #1596 so that issue could close on its actual subject.

`plugin-sqlite-worker.integration.test.ts` needs `out/main/plugin-sqlite-worker.js`. CI does not build it, so as of #1604 the suite declares the skip:

```
without TUFF_SKIP_SQLITE_WORKER_TESTS   1 failed | 6 skipped
with it                                 1 passed | 6 skipped
```

**Six plugin SQLite sandbox tests have never run in CI.** Declaring the skip cost nothing — they were skipped either way — but it also did not gain anything. Building the worker would.

## The cost is the question

Only `electron-vite build` produces it: the worker is a rollup input in the main-process build (`electron.vite.config.ts:209`), and the CLI offers `--rendererOnly` but **no `--mainOnly`**. So the options are a full build (main + preload + renderer) on every PR, or a hand-rolled parallel build path that would drift from the real config.

I do not have a measured build time. No workflow runs `electron-vite build` directly, and the only way to time it is to build in the shared worktree and occupy it for minutes — which this session has already lost work to more than once.

## Worth noting about the guard

The test refuses to skip silently and fails instead, with both remedies named in its error message. That design is why this was visible at all once someone parsed the JUnit report. It is a good pattern — the skip is now explicit, and this issue exists because explicit is still not the same as running.

TalexDreamSoul:
Closes #1607.

Six cases in `plugin-sqlite-worker.integration.test.ts` need `out/main/plugin-sqlite-worker.js`, which only `electron-vite build` produces. The core-app suite does not build, so **they have never run in CI**. #1604 made that skip deliberate instead of a failure, which fixed the reporting and changed nothing about the coverage.

## They are not incidental

| |
|---|
| worker-side SQL policy and bounded CRUD |
| browser-history copies staying query-only, rejecting every non-fixed SQL shape |
| stable row and result byte limit errors |
| **a database symlink inserted after host path validation** |
| a timed-out worker terminating and recovering without a late write |
| the worker-owned database cap mapping to a stable quota error |

## Cost, measured not guessed

From the release workflow's own log — the span from `building client environment for production` to the last `built in`:

```
ubuntu-latest   182s
macos-latest    181s
windows-2022    197s
```

`App suites (core-app)` takes **71s** in total and **blocks a PR** as of #1605. Adding a build there would more than triple the feedback loop for every change, including ones that cannot affect this.

So this runs as its own **path-filtered** workflow — the shape  already uses. Paid when the sandbox, its worker, or the config that builds it changes; free otherwise.

## Deliberately no skip env var

The guard failing on a missing artifact is what makes the job meaningful. If the build stops emitting the worker, this reports it rather than quietly skipping six security cases — which is exactly the state #1607 describes.

## Validated

`check:action-pins` and `check:workflow-injection` both pass over 20 workflows with this file in place.

**This PR's own run is the test of whether those six pass.** They have never executed in CI, so if any of them fails on a Linux runner, it will show here before merge rather than after.

### DIFF
--- a/.github/workflows/plugin-sqlite-sandbox.yml
+++ b/.github/workflows/plugin-sqlite-sandbox.yml
@@ -1,0 +1,80 @@
+name: Plugin SQLite Sandbox
+
+# The six sandbox cases in plugin-sqlite-worker.integration.test.ts need
+# out/main/plugin-sqlite-worker.js, which only `electron-vite build` produces. The core-app
+# suite does not build, so those cases have never run in CI -- they report `skipped` there and,
+# since #1604, say so deliberately rather than failing (#1607).
+#
+# They are not incidental coverage: worker-side SQL policy, browser-history copies staying
+# query-only, row and byte limits, a database symlink inserted after host path validation,
+# timed-out workers terminating without a late write, and the per-plugin database cap.
+#
+# Run here rather than inside App suites (core-app) because the build costs ~182s on this runner
+# against that job's 71s total. Paying that on every PR would more than triple the feedback loop
+# for changes that cannot affect this. Path-filtered, so it is paid only when the sandbox, its
+# worker, or the toolchain that builds it actually changes.
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/plugin-sqlite-sandbox.yml'
+      - 'apps/core-app/src/main/modules/plugin/runtime/**'
+      - 'apps/core-app/electron.vite.config.ts'
+      - 'pnpm-lock.yaml'
+      - 'pnpm-workspace.yaml'
+  push:
+    branches:
+      - master
+    paths:
+      - '.github/workflows/plugin-sqlite-sandbox.yml'
+      - 'apps/core-app/src/main/modules/plugin/runtime/**'
+      - 'apps/core-app/electron.vite.config.ts'
+      - 'pnpm-lock.yaml'
+      - 'pnpm-workspace.yaml'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: plugin-sqlite-sandbox-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  sandbox:
+    name: Plugin SQLite sandbox
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
## #1607 已由 #1610 合并关闭了 —— 这条现在会造成重复运行

先说事实,不是要抢:**#1610 已合并进 `55b38c18b`,也就是这条 PR 的 base**。所以 `ci.yml` 里已经有

```yaml
- name: Build core-app main process
  if: ${{ matrix.app == 'core-app' }}
  run: pnpm -C apps/core-app exec electron-vite build -c electron.vite.ci-main.config.ts
```

那六条用例现在每个 PR 都在 `App suites (core-app)` 里跑:`✓ plugin-sqlite-worker.integration.test.ts (7 tests)`。

如果这条也合入,同一批测试会在两个地方各跑一遍。

### 两种做法的实测代价

| | 做法 | 代价 |
|---|---|---|
| #1610(已合并) | 在现有 job 里加一步**仅主进程**构建 | **+11s**,每个 PR |
| 本条 | 独立 workflow + 路径过滤 | 不相关 PR 上 0;触发时要自带 checkout + install(该 job 实测约 34s)+ 构建 |

我最初也想过完整构建,CI 实测是 **87s**(tuffex 29 + build 58)—— 那时确实值得考虑独立 job。但把构建缩到只建主进程之后是 11s:渲染层才是需要 tuffex、吃掉大部分时间的部分,而这里用不到它。11s 之下,独立 workflow 省下的那点很可能被它自己的 install 抵消掉。

关键那步是 `electron.vite.ci-main.config.ts` —— 它 `import` 真实配置并取出 `main`,不复写任何东西,所以你在本条正文里提的"另起一套构建会漂移"这个顾虑不成立。

### 建议

关掉这条,或者改成**替换** #1610 那一步而不是叠加。如果你认为路径过滤更好,那我那 11s 的步骤应该同时删掉 —— 两者不该并存。

我不去动你的分支,判断留给你。

Issue #1303, opened 2026-08-08
[audit/build-ci] drizzle snapshot baseline stops at 0014 while migrations reach 0036, so db:generate's first output is unusable


TalexDreamSoul:
Found while fixing #592 (the missing `drizzle-kit` dependency), and left deliberately unfixed there because it changes files that run against users' databases.

## What

`drizzle-kit generate` now runs, but its first output cannot be committed.

| | |
| --- | --- |
| last drizzle snapshot | `apps/core-app/resources/db/migrations/meta/0014_snapshot.json`, written **2025-12-10** |
| latest migration | `0036_recommendation_exposure_daily.sql`, dated **2026-08-05** |

The 22 migrations in between were written by hand — which is what happens when the generator cannot run — so the snapshot baseline never advanced past 0014.

## Evidence

Running `db:generate` against the current `src/main/db/schema.ts` emits a **669-line** migration:

```
 45  CREATE TABLE
 81  CREATE INDEX
  3  ALTER TABLE
  1  DROP TABLE          (DROP TABLE `scan_progress`, part of a rename)
  1  CREATE UNIQUE INDEX
```

Those 45 tables are not new. They are everything migrations 0015–0036 already created, re-proposed because the diff is taken against the 0014 picture.

## Why it matters

- **The generated SQL would fail on any existing installation.** Zero of the 45 `CREATE TABLE` statements carry `IF NOT EXISTS`.
- A developer who follows CLAUDE.md, changes `schema.ts`, runs `db:generate` and commits the result ships a migration that breaks every upgrading user. The command gives no indication that its baseline is stale.
- Until this is fixed, the documented workflow is installed and runnable but must not be used — arguably a worse state than the command being absent, because nothing signals it.

## Suggested direction

Rebuild the snapshot baseline so the generator diffs against reality. Two options, both of which need a decision rather than an obvious default:

1. **Replay** — regenerate snapshots for 0015–0036 so the history is internally consistent.
2. **Rebaseline** — declare a fresh snapshot at the current schema and treat 0037 onward as the generated era.

Option 2 is far less work and loses the ability to regenerate intermediate states; option 1 preserves the history but requires reconstructing 22 hand-written steps.

Whichever is chosen, a guard is worth adding: a check that the highest `meta/NNNN_snapshot.json` matches the highest migration number, so the two cannot silently drift apart again.

---
<sub>Follow-up to #592. Filed while working the `audit` backlog.</sub>

[...]

PR #1647, opened 2026-08-12
ci: stop the drizzle snapshot baseline drifting further

TalexDreamSoul:
Adds the guard #1303 asks for.

> Whichever is chosen, a guard is worth adding: a check that the highest `meta/NNNN_snapshot.json` matches the highest migration number, so the two cannot silently drift apart again.

This is it, with one change: **it ratchets rather than demanding equality.** Equality fails on arrival, and a gate that is red the day it lands gets a tolerance bolted on within a week — which is the failure mode this repo has spent the day removing.

## The gap has already widened since the issue was written

```
issue recorded    snapshot 0014   journal 0036
today             snapshot 0014   journal 0037
```

Nothing noticed. That is the point.

## What the gap costs

From the issue, re-verified: `drizzle-kit generate` diffs `schema.ts` against the highest snapshot, so its first output re-proposes every table migrations 0015+ already created — **45 `CREATE TABLE`, none carrying `IF NOT EXISTS`**, which would fail on any existing installation. A developer following CLAUDE.md gets that file with nothing indicating the baseline is stale.

## What this does not do

Rebuilding the baseline is #1303's decision — replay 0015+, or rebaseline at the current schema — and it touches files that run against users' databases. This makes neither choice. It only stops the gap growing while the decision is pending.

## It fails in both directions

| | |
|---|---|
| gap **widens** | a migration landed without a snapshot; the generator now diffs against an even older picture |
| gap **narrows** | someone rebaselined and `EXPECTED_GAP` is stale. Left alone it becomes a ceiling nobody measures against — how a recorded tolerance turns into a permanent exemption |

Negative-controlled live, both directions:

```
add a journal entry, no snapshot  ->  gap 24 > recorded 23,                    exit 1
add a snapshot                    ->  "the gap shrank to 22 … lower EXPECTED_GAP", exit 1
both reverted                     ->  exit 0
```

`--self-test` covers 11 cases, including the two that matter most: an empty journal and an empty snapshot set read as `null` rather than `0`, so a moved directory or a partial checkout **fails** instead of reporting no drift.

Refs #1303

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,6 +178,15 @@ jobs:
       - name: macOS file-icon guard
         run: pnpm check:file-icon-guard && pnpm check:file-icon-guard:self-test
 
+      # drizzle-kit generate diffs schema.ts against the highest meta/NNNN_snapshot.json. That
+      # stopped at 0014 while the journal reached 0037, so its first output re-proposes every
+      # table 0015+ already created -- 45 CREATE TABLE, none with IF NOT EXISTS, which fails on
+      # any existing install. Rebuilding the baseline is #1303's decision and touches files that
+      # run against users' databases; this only stops the gap growing, and fails in both
+      # directions so the recorded number cannot quietly become a ceiling nobody measures.
+      - name: drizzle snapshot drift
+        run: pnpm check:drizzle-snapshot-drift && pnpm check:drizzle-snapshot-drift:self-test
+
       # `pnpm lint` reaches plugins either as workspaces or through a hand-maintained brace
       # list, and touch-dictation was in neither (#562). The list is the kind that drifts
       # silently, so this asserts the two paths between them still cover every directory.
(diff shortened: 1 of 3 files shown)


TalexDreamSoul:
关闭 —— 这个 PR 覆盖了一个已经存在且更好的实现。

`scripts/check-drizzle-snapshot-drift.mjs` 在 `app-shell-v2` 上**已经有了**,导出 `KNOWN_MISSING_SNAPSHOTS = 25` 和 `snapshotGap()`,还配了 `check-drizzle-snapshot-drift.test.mjs`。我从 origin 建树时没先查这个文件是否存在,直接写了同名文件,把导出面换成了 `highestJournalIndex` / `highestSnapshotIndex` / `evaluate` —— 已有测试 import 不到它要的两个符号,4 条全红。CI 报的就是这个,不是我的新步骤有问题。

更要紧的是,**已有的度量比我的正确**。他们数的是「没有快照的 journal 条目数」,能抓到中间的空洞;我用的是 `max(journal) − max(snapshot)`,只看尾部截断。他们的测试里有一条正是 `reports a gap in the middle, not just a truncated tail` —— 我的实现会让那条通过一个它本该抓住的情况。所以合并这个 PR 是净损失。

我在查 origin 是否已有该文件时还踩了一次 zsh 的坑:`git cat-file -e "$B:scripts/..."` 里的 `:s` 被当成参数修饰符,返回了「不存在」,让我一度以为没覆盖任何东西。用单引号包住整个 ref 才看到真相。

分支一并删除。#1303 上我会把状态更正过去。

PR #1670, opened 2026-08-12
test(plugins): run the vitest plugin suites too

TalexDreamSoul:
Toward #330.

`touch-translation` and `clipboard-history` are real workspace packages that declare a vitest `test` script. Between them they carry **8 test files and 39 passing cases, and no CI job ran a single one** — this runner only knew about `index.test.cjs`, and nothing else runs `pnpm -r test`.

Same hole as #1629, one layer over, **in the area #1629 was written to close.** 17 suites / 114 cases → **19 suites / 153 cases**. `pnpm test:plugins` is already wired into CI, so nothing in the workflow changes.

## Two halves, two emptiness checks

Discovery is by declared script rather than by file pattern, and the halves check emptiness separately. Folding them into one check would let either half disappear silently as long as the other still found something — which is the exact shape that hid these suites to begin with.

`pnpm -C <dir> exec vitest run` rather than the package's own script: `touch-translation`'s `test` is a bare `vitest`, which would sit in watch mode and never return.

## The exit-code check needs two plants at once to see

| plant | result |
|---|---|
| break a test in `touch-translation` | `1/19 plugin suites failed` |
| ignore the exit code | `All 19 plugin suites passed — 153 cases` |
| **both together** | **`All 19 plugin suites passed — 152 cases`** |

Vitest still prints `Tests  1 failed \| 18 passed`, so a count-only parser reads 18 and reports green. **The only trace of the real failure is a case total one lower than yesterday's**, and nobody watches a count. `status === null` gets its own branch for the same reason — a missing `pnpm` would otherwise fall through to whatever the count check decided.

Self-test 8 → 19 cases.

## One finding not fixed here

Running `touch-translation`'s suite **rewrites a tracked file**: `plugins/touch-translation/components.d.ts` loses its `import { GlobalComponents } from 'vue'` line, reproducibly. So the committed file is out of step with what its generator currently emits, and after this change CI will leave a dirty tree there. No job checks tree cleanliness, so nothing breaks today.

I did not commit the regenerated version, because I cannot tell from here whether CI's resolved `unplugin-vue-components` emits the same thing — "fixing" it could be a regression that only shows up on the runner. It belongs in #330's own open bucket: declaring, per suite, whether it consumes canonical source, a generated bundle, or a packaged runtime.

Refs #330

### DIFF
--- a/scripts/test-plugins.mjs
+++ b/scripts/test-plugins.mjs
@@ -55,7 +55,7 @@ export function discoveryFoundNothing(suites) {
  * failure mode that put #330 on the board.
  */
 export function classifySuiteRun({ status, stdout }) {
-  const text = String(stdout ?? '')
+  const text = stripAnsi(stdout)
   const read = (label) => {
     const match = text.match(new RegExp(`(?:^|\\s)${label}\\s+(\\d+)`, 'm'))
     return match ? Number(match[1]) : null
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
## The first CI run failed both vitest suites, each for its own reason, and both were invisible locally

Pushed a fix commit. Worth recording because one of them is the exact thing this PR exists to prevent.

**`clipboard-history` passed and was reported as failing.** Vitest colours its summary on CI, so the line arrives as `Tests \x1B[22m \x1B[1m\x1B[32m20 passed` and the parser read nothing. Locally the same command through the same `spawnSync` emitted no escapes at all. The guard behaved correctly — it refused to pass a run it could not read rather than reporting green — but the run it could not read was a green one. Both classifiers now strip ANSI first.

**`touch-translation` could not load its vite config.** It imports `@talex-touch/unplugin-export-plugin/vite`, whose `dist/` is gitignored, so a fresh checkout has nothing to import. My checkout had it built from some earlier run — **that is the entire reason all 19 suites were green here and 17 were green there.** The build now happens inside the runner rather than in the workflow step, so a direct `node scripts/test-plugins.mjs` behaves the same way as the pipeline.

Both were reproduced locally before fixing, by deleting the dist and forcing colour:

```
previous commit, dist removed, FORCE_COLOR=1   2/19 plugin suites failed
this commit, same conditions                   all 19 passed — 153 cases
```

Self-test 19 → 22 cases, including the exact coloured line CI produced.

Issue #318, opened 2026-07-28
Indexing: re-validate and enforce large-directory memory bounds


TalexDreamSoul:
## Priority
P1 — verify an operational memory bound before declaring the prior R3 risk closed.

## Problem
The search/cross-platform audit reported that scan and reconciliation could retain multiple full copies of a large filesystem list. Current `master` now contains `scanDirectoryBatches`, a batch/ack worker protocol, and paged reconciliation (`RECONCILIATION_PAGE_SIZE`), so the old implementation claim must be revalidated rather than repeated as fact.

## Required investigation
Measure the current pipeline on a large synthetic tree and determine whether memory remains bounded across:
- worker traversal and batch acknowledgement;
- client pending-batch retention;
- persistence/index flush backpressure;
- reconciliation database paging;
- cancellation and shutdown.

## Proposed solution if the budget is exceeded
- Keep worker traversal streaming and allow only a bounded number of unacknowledged batches.
- Persist/index each batch before acknowledging additional work.
- Page database reads and deletes; never materialize a provider-wide DB result set.
- Release per-root seen-path state incrementally or move it to a temporary SQLite structure when an in-memory set exceeds the budget.
- Propagate cancellation through traversal, persistence, and reconciliation.

## Acceptance criteria
- [ ] A reproducible synthetic-tree benchmark records file count, batch size, peak RSS/heap, duration, and cancellation behavior.
- [ ] Peak memory is bounded by configuration/batch size rather than total file count, or the remaining total-count dependency is documented and fixed.
- [ ] Reconciliation does not hold complete filesystem and DB lists concurrently.
- [ ] Cancellation releases worker/client buffers and idle shutdown still works.
- [ ] Focused tests cover backpressure, bounded pending batches, pagination, and cancellation.
- [ ] The living audit backlog is updated: close R3 if the current implementation already meets the budget; otherwise retain only the verified residual risk.

## Non-goals
- Broad FileProvider refactoring unrelated to measurable memory ownership.

TalexDreamSoul:
## 已加入可复现基准并完成 traversal 侧实测（**PR #1105**，commit `26db01c46`），**本 issue 不关**

新增 `apps/core-app/scripts/file-scan-memory-benchmark.mjs`：构建嵌套合成树，驱动 `scanDirectoryBatches`（scan worker 实际调用的那个 traversal），**在批回调内部**采样 RSS/heap——也就是一批数据正活着的那一刻。

### 实测结果

```
[...]

PR #1683, opened 2026-08-12
test(indexing): hold the scan backpressure bound

TalexDreamSoul:
Toward #318.

The prior investigation on that issue concluded the memory bounds are structurally met, and named the one thing worth fixing before closing R3:

> **`queued.length >= 2` has no test.** That constant *is* the memory bound this issue exists to protect. Someone can change it to `1000` — or delete the `while` and keep the `push` — and nothing goes red.

This is that test.

## Why the constant is the whole risk

`enqueue` parks the producer while two batches sit unconsumed, so peak retention is **batch size × 2 regardless of how many files the tree holds**. Remove it and a million-file root goes back to being unbounded — while every structural argument in the code review still reads as true.

## Three plants, each caught

| plant | result |
|---|---|
| raise the bound from `2` to `1000` | fails |
| delete the park entirely, keep the `push` | fails |
| stop waking the producer after a consume | fails, `after 2 consumed: expected 3 to be 4` |

**The second case is asserted as a constant offset rather than a fixed number** — after N consumed the producer has accepted N + 2. That says *the retention does not scale with the tree*, which is the actual claim, rather than checking one number at one point.

The abort case matters for a different reason: the park sits inside a loop that re-checks the signal, so a consumer stopping early releases the producer instead of leaving it waiting for a pull that never comes. **Without that check the test hangs rather than fails**, so it asserts the producer settled at all.

## What this does not do

No benchmark. The prior comment argued a test is worth more here than one, and I agree with its reasoning: a benchmark tells you what today's number is, a test tells you when it changes. The synthetic-tree benchmark criterion stays open.

`search-engine` 75 files / 666 tests pass. `typecheck:node` clean, lint clean.

Refs #318

### DIFF
--- a/apps/core-app/src/main/modules/box-tool/search-engine/file-indexed-source.test.ts
+++ b/apps/core-app/src/main/modules/box-tool/search-engine/file-indexed-source.test.ts
@@ -332,6 +332,120 @@ describe('fileIndexedSource', () => {
     expect(producerSignal?.aborted).toBe(true)
   })
 
+  /**
+   * The backpressure bound (#318).
+   *
+   * `enqueue` parks the producer while two batches sit unconsumed, and that constant is the
+   * memory bound the whole R3 risk rests on: peak retention is batch size x 2 regardless of how
+   * many files the tree holds. Nothing tested it. Someone could raise it to 1000, or delete the
+   * `while` and keep the `push`, and every structural argument in the code review would still
+   * read as true while a million-file root went back to being unbounded.
+   */
+  describe('scan backpressure', () => {
+    function trackedProducer(total: number): {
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
`App suites (core-app)` failed on the first run and **passed on a rerun of the same commit** — `plugin-sqlite-worker.integration.test.ts > maps the worker-owned database cap to a stable quota error`, `Test timed out in 15000ms`. Nothing else in that run failed (705 files, 5962 tests).

This change touches one test file in `search-engine` and no source, so it cannot reach a sqlite worker integration test. Recorded as a flake with the rerun as the evidence rather than asserted as one.

PR #1705, opened 2026-08-12
fix(deps): move three dependencies onto their patched lines

TalexDreamSoul:
Toward #483.

That issue's baseline was **51 alerts (3 critical, 20 high, 25 moderate, 3 low)** on 2026-08-04. Measured today against `master` after the app-shell-v2 merge: **27 open (1 critical, 8 high, 14 medium, 4 low)**.

Three of the 27 were one patch position away, and the manifests now say so rather than leaving it to resolution:

| | | |
|---|---|---|
| `markdown-it` | `14.1.0` → `14.2.0` | root devDependency, previously pinned exactly |
| `dompurify` | `^3.4.11` → `^3.4.13` | `packages/tuffex` |
| `mermaid` | `^11.16.0` → `^11.16.1` | workspace catalog |

**The last two already permitted the patched version** — `^3.4.11` admits 3.4.13 — but pnpm keeps a satisfying resolution rather than moving to the newest, so the lockfile sat on 3.4.12 and 11.16.0. Refreshing alone would have fixed it silently and let the next resolution drift back; raising the declared floor records the security requirement where someone reading the manifest can see it.

Lockfile now carries exactly one version of each. Production audit gate unchanged at 8 Critical/High, 8 allowlisted.

The full 27-alert triage goes on the issue.

Refs #483

## Summary by CodeRabbit

* **Chores**
  * Updated Markdown rendering, HTML sanitization, and Mermaid diagram components to newer patch versions.
  * Includes routine dependency updates with no changes to public APIs or user-facing functionality.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,7 +98,7 @@
     "eslint": "^9.39.4",
     "husky": "^9.1.7",
     "lint-staged": "^16.4.0",
-    "markdown-it": "14.1.0",
+    "markdown-it": "14.2.0",
     "markdownlint": "0.39.0",
     "marked": "catalog:",
     "remark-mdc": "3.11.1",
--- a/packages/tuffex/package.json
+++ b/packages/tuffex/package.json
@@ -91,7 +91,7 @@
     "@talex-touch/utils": "workspace:^",
     "@vue/reactivity": "^3.5.27",
     "@vueuse/core": "^14.3.0",
-    "dompurify": "^3.4.11",
+    "dompurify": "^3.4.13",
     "gsap": "^3.15.0",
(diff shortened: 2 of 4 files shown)


TalexDreamSoul:
`App suites (nexus)` 第一次挂在 `admin-emergency-token.test.ts > 篡改 token 后校验失败`,**同一 commit 重跑通过**,本机也 8/8 通过。

本 PR 只动了 markdown-it / dompurify / mermaid 三个版本号与 lockfile,够不到 token 校验。按偶发记录,证据是那次重跑,不是断言。

PR #1708, opened 2026-08-12
docs(security): nanoid does move with the nuxt closure after all

TalexDreamSoul:
**CodeRabbit caught an error in my reasoning on this PR's first commit, and it was right.** The PR now says the opposite of what it opened with.

## What the advisory actually covers

`GHSA-28wg-ghj8-5hjv` has vulnerable range **`>= 4.0.0, < 5.1.16`**. The lockfile has two copies:

```
nanoid@3.3.16   via postcss@8.5.23     ← NOT covered; 3.3.16 is the patched 3.x line
nanoid@5.1.6    via @vue/devtools-core → @nuxt/devtools@2.7.0 → nuxt   ← the affected one
```

So the exposure is entirely inside the nuxt closure, and **the original allowlist reason — "moves when that closure moves" — was correct.**

## Three revisions of the same field, and the first was right

| | |
|---|---|
| original | *"Transitive under the same @sentry/nuxt → nuxt closure. Moves when that closure moves."* — **correct** |
| #1691 | reattributed it to `vue-router → … → postcss` as an independent root — **wrong version** |
| this PR, first commit | doubled down: "blocked by postcss, cannot be bumped" — **same wrong version** |
| now | names the version, the range, and both bad corrections |

## What I got wrong twice

I traced the **package name** to a production root without checking **which installed version the advisory covers**. `pnpm why nanoid --prod` prints both copies; I read the tree and not the range. Two corrections built on the same omission.

The reason text now carries the version, the range and the history, so the next reader can see why it contradicts the previous two commits.

Gate unchanged: 8 Critical/High, 8 allowlisted, self-test 9 cases.

Refs #483, #1098

## Summary by CodeRabbit

* **Documentation**
  * Clarified production security audit notes to distinguish affected and unaffected dependency versions.
  * Updated exception explanations to accurately reflect advisory coverage, remediation status, and dependency relationships.
  * Corrected outdated attribution details to improve the accuracy and transparency of audit records.

### DIFF
--- a/.github/prod-audit-allowlist.json
+++ b/.github/prod-audit-allowlist.json
@@ -20,7 +20,7 @@
       "id": "GHSA-28wg-ghj8-5hjv",
       "module": "nanoid",
       "severity": "high",
-      "reason": "Reached from production by many roots, and blocked upstream rather than by the nuxt closure. `pnpm why nanoid -r --prod` puts nanoid@3.3.16 under postcss@8.5.23, and postcss is reached directly by @talex-touch/core-app, tuff-cli and tuff-cli-core through @vue/compiler-sfc, as well as by @nuxt/vite-builder and by vue-router. So it does NOT clear when #1098 moves the nuxt closure. It also cannot be bumped here: postcss@8.5.23 requires nanoid ^3.3.16 and the patched line is 5.1.16, which is ESM-only, so the postcss 8.x line will not adopt it. Moves when postcss does.",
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
@coderabbitai 你是对的,已按此更正。

advisory `GHSA-28wg-ghj8-5hjv` 的范围是 `>= 4.0.0, < 5.1.16`,所以 postcss 拉的 `nanoid@3.3.16` **根本不在覆盖范围内**(3.3.16 已是 3.x 线的补丁版本)。真正受影响的只有 `nanoid@5.1.6`,它来自 `@vue/devtools-core → @nuxt/devtools@2.7.0 → nuxt`。

也就是说最初的豁免理由「moves when that closure moves」是对的,我在 #1691 的「更正」是错的,本 PR 第一个提交又加倍。三次修改同一个字段,第一次才对。

我两次错在同一处:**只把包名追到生产根,没有核对 advisory 实际覆盖哪个已安装版本。**`pnpm why nanoid --prod` 两个副本都打印了,我读了树没读范围。

Issue #346, opened 2026-07-28
Search: measure the query cache and remove it if benefit is negligible


TalexDreamSoul:
## Priority
P2 — evidence-based performance cleanup.

## Problem
SearchCore retains a small query cache (5-second TTL / bounded LRU), but the key includes activation/context state and the current hit rate/latency benefit is unknown. A cache with low hit rate still adds invalidation, session, and stale-result complexity.

## Proposed solution
- Add low-sensitive counters for lookup, hit, miss, invalidation reason, age, and saved provider latency; never record query text.
- Measure representative CoreBox, AI/background, activation-change, and repeated-query workloads.
- Verify cache hits still create fresh request-scoped sessions and cannot leak sender/activation state.
- Keep the cache only if it has a measurable benefit over its complexity; otherwise remove it cleanly.

## Acceptance criteria
- [ ] A benchmark/evidence report records hit rate, p50/p95 saved latency, invalidation distribution, and memory bound without query content.
- [ ] Cache hits preserve fresh session ID, caller, activation snapshot, cancellation, and sink ownership.
- [ ] Stale provider/config/permission/index revisions cannot be served.
- [ ] A documented keep/remove threshold is applied.
- [ ] If removed, all cache-only state/invalidation code and tests are deleted without compatibility shims.
- [ ] If retained, diagnostics and regression tests defend the measured contract.
- [ ] Audit C3 is closed or updated with evidence.

## Non-goals
- Increasing TTL/size before measuring correctness and benefit.

TalexDreamSoul:
## 现状核查（2026-08-06）：缓存仍在，但 issue 假设的问题面比实际大——真正缺的只有测量

分支 `fix/346-query-cache`（基于 origin tip）。先把静态能确定的都确定下来，免得下一个接手的人重跑一遍。

### 缓存位置与参数

`search-engine/search-core.ts`：`SEARCH_CACHE_TTL_MS = 5_000`、`SEARCH_CACHE_MAX_SIZE = 100`、`SEARCH_CACHE_ITEM_LIMIT = 200`（L105-112），`private searchCache = new Map<...>`（L186）。唯一读取点在 **L887**，唯一写入点 `cacheSearchResult` L198。

失效点四处：LRU 逐出（L221）· `completePrivacyRetentionCleanup()`（L414）· `handleSearchIndexCommit()`（L431）· pin 切换（L2065）。

### 已经成立的验收条款（不用再做）

[...]

PR #1712, opened 2026-08-12
test(search): read the query cache counters end to end

TalexDreamSoul:
Progress on #346, and it surfaced why that issue cannot be answered as written.

## The counters were unreadable

`getSearchCacheTelemetry()` has **exactly one reference in the repository — its own declaration**. Nothing calls it: no IPC channel, no dashboard, no test. (Positive control: the same search finds `recordSearchMetrics` in four files, so it does work.)

So while `search-cache-telemetry.test.ts` covers the counter object thoroughly — 16 cases including percentile corruption and sample bounding — every `recordHit` / `recordMiss` / `recordInvalidation` call site in `search-core.ts` was unverified. Counters incrementing into a snapshot nobody obtains read exactly like counters that were never wired.

This adds the first end-to-end read: a cold query, a warm repeat, an index commit, a third query, then the snapshot.

## And it pins the thing #346 turns on

The miss taxonomy is the whole point of the instrumentation — the comment at the classify site says so: *"The reason is the point, not the hit rate: `revision-mismatch` means an index commit denied the entry, which is a different story from a query nobody repeated."*

**`revision-mismatch` is unreachable.** `markCommitted` bumps the revision and notifies listeners synchronously; this facade's listener clears the entire cache before returning. By the time any lookup compares revisions, the entry is already gone, so the miss is `absent`.

The test asserts that directly (`misses['revision-mismatch']` is 0, `misses.absent` is 2), and the causal plant is removing the `searchCache.clear()` on index commit — which turns the test red, because `revision-mismatch` then becomes reachable. The classifier is fine; the clear is what makes the distinction unobservable.

This is not cosmetic. A repeat denied by an index commit lands under the same reason as a query nobody repeated. A report built on that distribution reads "nobody repeats queries, remove the cache" when the truth might be "repeats happen and the index keeps denying them". #346's keep/remove decision cannot rest on it as it stands.

## Verification

| plant | result |
| --- | --- |
| remove `recordHit` | 1 failed |
| remove `recordMiss` | 1 failed |
| remove `searchCache.clear()` on index commit | 1 failed |

`src/main/modules/box-tool/search-engine/` passes: 75 files, 667 tests. `search-core.ts` is restored byte-for-byte after the plants — `git diff` on it is empty, and this PR touches only the test file.

## What remains on #346

The report it asks for still cannot be produced. Two blockers, in order: the snapshot has no reader outside a test, and the reason distribution mis-attributes commit-denied repeats. Both are recorded on the issue; neither is fixed here, because how to surface the snapshot is a product decision and changing the invalidation order is a behaviour change that wants its own justification.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for search-cache telemetry across cold, warm, and post-index-commit searches.
  * Verifies cache lookups, hits, hit rates, and invalidation behavior.
  * Confirms searches after an index update are correctly recorded as cache misses.

### DIFF
--- a/apps/core-app/src/main/modules/box-tool/search-engine/search-core.contracts.test.ts
+++ b/apps/core-app/src/main/modules/box-tool/search-engine/search-core.contracts.test.ts
@@ -840,6 +840,64 @@ describe('SearchEngineCore facade contracts', () => {
     expect(search).toHaveBeenCalledTimes(2)
   })
 
+  /**
+   * The first end-to-end read of the cache counters (#346).
+   *
+   * `search-cache-telemetry.test.ts` covers the counter object thoroughly, and
+   * `getSearchCacheTelemetry` had exactly one reference in the repository -- its own declaration.
+   * So every `recordHit` / `recordMiss` / `recordInvalidation` call site in search-core was
+   * unverified: the counters incremented into a snapshot nobody obtained, which reads the same as
+   * counters that were never wired at all.
+   *
+   * The reason distribution is the part #346 turns on. A miss attributed to `absent` says "nobody
+   * repeated this query", which argues for removing the cache; one attributed to
+   * `revision-mismatch` says "the index denied a repeat that would otherwise have hit", which
+   * argues the opposite. This records which reason an index commit actually produces.
+   */
+  it('attributes a repeat denied by an index commit, and pins which reason it used', async () => {
+    const search = vi.fn(
+      async () =>
+        ({ items: [buildItem('telemetry-item', 'telemetry-provider', 'Telemetry')] }) as never
+    )
+    const query = { inputs: [], text: 'cache telemetry contract' } as TuffQuery
+    core.registerProvider(buildProvider('telemetry-provider', search) as never)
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Taken — you were right, and the plant confirms it.

With one cached entry, `invalidations['index-commit'] > 0` passes on an implementation that records one invalidation per commit *call* rather than one per *entry dropped*. Those two readings are one argument apart at the record site (`this.searchCache.size` versus `1`), and `search-cache-telemetry.test.ts` distinguishes them in its own unit test while this end-to-end one did not.

`e713f0a14` caches two distinct queries before the commit and asserts exactly: `invalidations['index-commit'] === 2` and `entriesDropped === 2`. Replacing `this.searchCache.size` with `1` at the record site now turns it red — the exact faulty implementation you described.

Three plants pass against the new shape: per-call counting fails, removing `recordMiss` fails, and removing the `searchCache.clear()` on index commit fails. 75 files / 667 tests in `search-engine/` still green.

Issue #339, opened 2026-07-28
Architecture: split Plugin and QuickOps modules by lifecycle ownership


TalexDreamSoul:
## Priority
P2 — maintainability and isolation after security/release gates are green.

## Problem
Two main-process modules remain multi-owner coordination hotspots:

- `apps/core-app/src/main/modules/plugin/plugin.ts`: ~3,060 lines;
- `apps/core-app/src/main/modules/quick-ops/index.ts`: ~2,999 lines.

The concern is not line count by itself. Each file combines lifecycle, permission/policy, execution dispatch, state, transport, diagnostics, and feature-specific behavior, making isolated review and teardown correctness difficult.

### Plugin
- activation/lifecycle state;
- permission-bound capability dispatch;
- feature/result projection;
- transport registration/disposal;
- diagnostics/audit.

### QuickOps
- query/command parsing;
- capability policy and confirmation;
- runtime/session ownership;
- platform service adapters;
- result projection and diagnostics.

Extract existing behavior behind narrow interfaces; do not add compatibility aliases or a second state source.

## Acceptance criteria
- [ ] Each extracted service has one lifecycle owner and idempotent disposal.
- [ ] Transport registration remains typed and is not duplicated.
- [ ] Permission/confirmation checks happen before side effects in every extracted path.
- [ ] Existing Plugin and QuickOps full focused suites pass without broad test rewrites.
- [ ] Reload/unload/shutdown tests prove timers, listeners, sessions, and workers are released.
- [ ] The original files become orchestration/composition roots rather than re-export facades or arbitrary fragments.
- [ ] No extraction is justified solely by a target line count.

## Non-goals
- Plugin SDK redesign.
- New QuickOps features or UI polish.
- Mechanical one-class-per-file splitting.

TalexDreamSoul:
## 成分勘察已完成并合入（**PR #1111**，commit `52e28de9c`），**本 issue 不关**

清单：`docs/engineering/plugin-quickops-ownership-survey.md`。本 issue 自述「问题不是行数本身」，所以这次测的是**成分**。

### 两个数字都要更新，其中一个变化很大

| 文件 | issue 记录 | 实测 | |
|---|--:|--:|---|
| `modules/plugin/plugin.ts` | ~3,060 | **4,013** | **+31%**，issue 挂起期间仍在增长 |
| `modules/quick-ops/index.ts` | ~2,999 | 2,999 | 未变 |

### 两个文件形态不同，一套拆法不适用于两者

[...]

PR #1722, opened 2026-08-12
ci: ratchet the two files #339 is about so the split stops losing

TalexDreamSoul:
Progress on #339, and a measurement that changes what "progress" has meant on it.

## Three PRs of real progress, net backwards

`plugin.ts` was **~3,060 lines when #339 was filed** (2026-07-28). Three extraction PRs landed — #1678, #1680, #1681 — and each reported a true number:

```
4,069  →  3,947 (#1678)  →  3,872 (#1680)  →  3,844 (#1681)
```

Every one of those is measured against the **previous PR**. Against the issue:

| | plugin.ts | quick-ops/index.ts |
| --- | --- | --- |
| when #339 was filed | ~3,060 | ~2,999 |
| now | **3,844** | 2,875 |
| | **+784** | −124 |

Extraction removed 225 lines. Accretion added 1,009 over the same period. Meanwhile the plugin directory went from **61 source files to 107** — the extractions are real, they are just slower than what they are racing, and nothing said so because no measurement spanned more than one PR.

## What this adds, and what it deliberately is not

A ceiling that may fall and may not rise, on the two files #339 names and nothing else.

**It does not measure what #339 is about, and #339 says so first**: *"the concern is not line count by itself."* That is correct — no line count can see that one file owns lifecycle, permission policy, dispatch, state, transport and diagnostics simultaneously. Line count is here for one reason: it is the only thing that would have caught the last thousand lines.

A ceiling left more than 50 lines above its file **also fails**, asking to be lowered. A ratchet nobody tightens is a comment. 50 is smaller than any extraction so far — the smallest was 27 — so a real extraction is asked to lower the ceiling and a one-line deletion is not.

## Verification

Proved against the real files, not only the self-test:

| plant | result |
| --- | --- |
| append one line to `plugin.ts` | fails: `is 3845 lines, ceiling 3844 (#339)` |
| delete 60 lines | fails: `lower the ceiling to 3785` |
| delete 40 lines (inside the slack) | passes |
| an empty `RATCHET` | fails — it cannot be a check that cannot fail |

`--self-test` covers 9 cases including the two thresholds, a missing file failing rather than counting as zero lines, `wc -l` semantics with and without a trailing newline, and the real entries resolving.

`mise run workflows:lint` exits 0 with the new step.

## What is still not done

The split itself. This buys it time; it does not do it. The ownership boundaries #339 lists — lifecycle, permission-bound dispatch, feature projection, transport, diagnostics — remain the work, and the next extraction now has to lower a number that is written down.

## Summary by CodeRabbit

* **Chores**
  * Added automated checks to monitor module size and prevent configured limits from being exceeded.
  * Added commands for running the module-size check and its self-test.

* **Tests**
  * Added coverage for line counting, threshold handling, missing files, and configured paths.
  * Integrated module-size validation into the pull request quality workflow.

### DIFF
--- a/.github/module-size-ratchet.json
+++ b/.github/module-size-ratchet.json
@@ -1,0 +1,22 @@
+{
+  "$comment": [
+    "Ceilings for scripts/check-module-size-ratchet.mjs. A ceiling may fall and may not rise:",
+    "the check compares each value here against the same file in the PR base revision, so lowering",
+    "one is part of the change that shrinks the file and raising one fails the build.",
+    "Only the two files #339 names. This is not a general file-size policy."
+  ],
+  "files": [
+    {
+      "file": "apps/core-app/src/main/modules/plugin/plugin.ts",
+      "ceiling": 3844,
+      "issue": "#339",
+      "note": "~3,060 when #339 was filed; peaked at 4,069 while three extraction PRs were landing."
+    },
+    {
+      "file": "apps/core-app/src/main/modules/quick-ops/index.ts",
+      "ceiling": 2875,
+      "issue": "#339",
+      "note": "~2,999 when #339 was filed. Down 124 since, and this holds that."
+    }
+  ]
+}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,6 +182,19 @@ jobs:
       - name: Check maintenance-audit report claims
         run: pnpm check:audit-report-claims && pnpm check:audit-report-claims:self-test
 
+      # plugin.ts was ~3,060 lines when #339 asked for it to be split. Three extraction PRs landed
+      # and each reported real progress -- 4,069 to 3,947 to 3,872 to 3,844 -- against a baseline
+      # that had already grown a thousand lines since the issue was written. Extraction removed 225
+      # lines while accretion added 1,009, and nothing said so, because each PR compared with the
+      # previous PR.
+      #
+      # This does not measure what #339 is about; that issue says "the concern is not line count by
+      # itself" and is right. It measures whether the split is losing, which is the question three
+      # PRs in a row could not answer. The ceiling falls and never rises, and a ceiling left too far
+      # above the file also fails -- a ratchet nobody tightens is a comment.
+      - name: Check module size ratchet
+        run: pnpm check:module-size-ratchet && pnpm check:module-size-ratchet:self-test
+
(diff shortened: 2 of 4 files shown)


TalexDreamSoul:
All three taken. The first one was a real hole and the fix is proved against a real git base, not a fixture.

## 1 — ceiling increases from the base revision (Major)

Correct, and it made the ratchet decorative: raise `plugin.ts` from 3844 to 3895, add one line, and every check passed. A ratchet that unscrews.

The ceilings moved out of the script into `.github/module-size-ratchet.json`, so the base copy is a `git show` plus a `JSON.parse` rather than a regex over source, and `findRaisedCeilings` compares each current ceiling with the same file's ceiling at `merge-base(HEAD, origin/$GITHUB_BASE_REF)`.

Proved by doing exactly what you described, with the pushed branch as the base:

```
✗ apps/core-app/src/main/modules/plugin/plugin.ts ceiling raised 3844 → 3895.
  A ceiling may fall, not rise (#339).
```

Exit 1. Restoring the ceiling and keeping the extra line still exits 1 on the original rule; restoring both exits 0.

A file absent from the base is **not** a raise — that is how a file joins the ratchet. An unreachable base returns `null` and says so on stdout rather than being read as "no raises found", because on CI the base is always reachable (the PR Quality job checks out with `fetch-depth: 0`), so that line appearing in a CI log means the raise check silently did not run.

## 2 — `wc -l` semantics (Minor)

You are right that `countLines` is not `wc -l`: `wc` counts newline characters, so `'a\nb'` is 1 there and 2 here.

Kept the behaviour, corrected the claim. A file whose last line has no trailing newline would otherwise be measured one short, and its ceiling would sit one tighter than the number anyone reading the file arrives at. The doc comment now says what it does and why it is deliberately not `wc -l`, instead of asserting the opposite of what the two tests below it check.

## 3 — the slack rationale (Minor)

Also right, and it was self-contradictory: "50 is smaller than any of the three extractions (122, 75, 27)" — 50 is larger than 27. Now: 50 lets the smallest extraction so far land without a ceiling edit and forces one after a real reduction.

The `3,845` / `224` in the workflow comment are corrected to `3,844` / `225` to match the configured ceiling and the script.

Self-test 9 → 15 cases, adding raise / lower / unchanged / new-file for `findRaisedCeilings`, and the unreachable-base case.

Issue #343, opened 2026-07-28
Architecture: simplify App/File Provider service and indexing ownership


TalexDreamSoul:
## Priority
P2 — simplify indexing ownership after the P0/P1 correctness and memory gates are green.

## Problem
`app-provider.ts` (~4,223 lines) and `file-provider.ts` (~3,604 lines) remain large composition roots. FileProvider also wires a high number of thin services, while search indexing has thin re-export layers beside the real shared implementations. This makes ownership, search, and review difficult and can create a second convention instead of reducing complexity.

The goal is not arbitrary line-count reduction. The goal is one discoverable owner for scan/watch/reconcile, persistence, search projection, asset/icon work, and lifecycle disposal.

## Proposed solution
- Draw the current dependency/ownership map before moving code.
- Keep `FileProvider`/`AppProvider` as orchestration roots with explicit ports for persistence, indexing, native scan, and result projection.
- Merge/remove one-method adapters that add no policy, test seam, or ownership value.
- Remove misleading local re-export layers; import the canonical shared implementation directly.
- Keep workers and SQLite writers behind typed ports with one lifecycle owner.
- Preserve the provider control-plane work in #334 as a prerequisite, not a competing abstraction.

## Acceptance criteria
- [ ] Each retained service has an explicit policy/ownership/test-seam justification.
- [ ] Duplicate thin re-exports/adapters are removed without aliases.
- [ ] Provider composition shows scan/watch/reconcile/persist/index/result/disposal ownership in one place.
- [ ] No new service exists only to forward one call without policy.
- [ ] Focused app/file provider, worker, persistence, search, and shutdown tests pass.
- [ ] Startup/search/indexing performance and memory do not regress.
- [ ] Documentation/import search leads to the canonical implementation, not shadow files.

## Non-goals
- One-class-per-file refactoring.
- Changing search ranking, database schema, or user-facing behavior.

TalexDreamSoul:
## 归属图已完成并合入（**PR #1113**，commit `1c205ba2b`），**本 issue 不关**

文档：`docs/engineering/file-provider-service-ownership-map.md`。本 issue 要求「移动代码前先画图」且明确「目标不是任意削减行数」，所以判定依据是**是否承载策略**，不是大小。

### 三个数字要更新

| | issue | 实测 |
|---|--:|--:|
| `apps/app-provider.ts` | ~4,223 | **4,579**（+8%） |
[...]

PR #1723, opened 2026-08-12
ci: extend the size ratchet to the two files #343 is about

TalexDreamSoul:
Same measurement as #339 on a different pair of files, and the same result.

## #343's own numbers were right, and both files have grown since

| file | in #343 | at the commit before it was filed | today |
| --- | ---: | ---: | ---: |
| `app-provider.ts` | ~4,223 | **4,223** | **4,600** (+377) |
| `file-provider.ts` | ~3,604 | **3,604** | **4,298** (+694) |

**+1,071 lines together**, over the two weeks the issue has been asking for them to be simplified. The issue's figures match git exactly at its own baseline, so this is not a stale premise — it is an accurate one that has got worse.

## Same framing as #339, for the same reason

#343 says *"The goal is not arbitrary line-count reduction. The goal is one discoverable owner for scan/watch/reconcile, persistence, search projection, asset/icon work, and lifecycle disposal."* That is right, and no line count can see any of it.

The ratchet does not measure that. It measures whether the simplification is losing, which is a question nothing was answering — and the answer for two weeks was yes.

## Verification

`file-provider.ts` with one line appended:

```
✗ .../file-provider.ts is 4299 lines, ceiling 4298 (#343).
```

Exit 1; restoring exits 0. The raise-a-ceiling guard from #1722 applies to these entries too — a PR that lifts 4298 to 4400 fails against the base revision.

Four files in the ratchet now, self-test 15 cases green.

## Summary by CodeRabbit

* **Documentation**
  * Updated module-size tracking documentation with growth details for provider modules.
  * Added size limits associated with issue `#343`.

### DIFF
--- a/.github/module-size-ratchet.json
+++ b/.github/module-size-ratchet.json
@@ -3,7 +3,11 @@
     "Ceilings for scripts/check-module-size-ratchet.mjs. A ceiling may fall and may not rise:",
     "the check compares each value here against the same file in the PR base revision, so lowering",
     "one is part of the change that shrinks the file and raising one fails the build.",
-    "Only the two files #339 names. This is not a general file-size policy."
+    "Only the two files #339 names. This is not a general file-size policy.",
+    "2026-08-12: extended to #343 for the same reason as #339. Both issues say line count is not",
+    "the point and both are right; both also grew while asking to be split -- plugin.ts +784,",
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Taken — the `$comment` still said "only the two files #339 names" after two more went in for #343. It is four files across two issues, and the line now says so. `266a9ef44`.

PR #1724, opened 2026-08-12
fix(ci): the size ratchet was a check against writing things down

TalexDreamSoul:
A defect in the ratchet I added in #1722/#1723, found by discharging a caveat I had attached to my own #343 comment.

## The problem

It counted **raw lines**. On the four files it guards, comments and blanks are 10-15%:

| file | raw | code | comment + blank |
| --- | ---: | ---: | ---: |
| `plugin.ts` | 3,844 | 3,249 | 15% |
| `quick-ops/index.ts` | 2,875 | 2,594 | 10% |
| `app-provider.ts` | 4,600 | 3,950 | 14% |
| `file-provider.ts` | 4,298 | 3,703 | 14% |

And on the *growth* that put them under the ratchet it is much worse. `app-provider.ts` is +377 since #343 was filed, and **+131 of that is comments** — 35% of the increase.

So the ceiling would have failed the next person who wrote down why a decision was made, in files whose stated problem is that their decisions are unexplained. That is the opposite of what both issues want.

Ceilings are code lines now, lowered to match: 3844→3249, 2875→2594, 4600→3950, 4298→3703. Lowering is what a ratchet permits, so the base-revision check from #1722 is satisfied by the same edit.

## How it was found

Not by review. My #343 comment carried a caveat: the +1,071 might be legitimate streaming work from #318's `scanDirectoryBatches` / paged reconciliation, in which case "simplification is losing" would not hold. Discharging it:

| | added code | mentions batching / paging / ack |
| --- | ---: | ---: |
| `file-provider.ts` | 612 | 58 (9.5%) |
| `app-provider.ts` | 385 | 40 (10.4%) |

So the hedge was mostly wrong — a tenth, not the bulk. But the same pass produced the comment counts, and those were a defect in my own check rather than in the codebase.

## Verification

Seven new self-test cases for the counter: a line comment, a block comment across lines, a one-line block comment, blank lines, code resuming after a block closes, a trailing comment not removing its line, and an all-comment file counting zero. 15 → **22 cases**.

And on the real file rather than a fixture:

| appended to `app-provider.ts` | exit |
| --- | --- |
| six comment lines (two `//`, a four-line block) | **0** |
| one line of code | **1** — `is 3951 code lines, ceiling 3950` |

The stale header — "the two coordination hotspots in #339" — now names four files and both issues, matching the fix CodeRabbit asked for in the JSON on #1723.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved module size tracking by counting nonblank, non-comment code lines.
  * Added support for correctly ignoring block comments and blank lines.
  * Updated size thresholds and messages to reflect code-line counts.
* **Tests**
  * Expanded validation coverage for comments, blank lines, and block comments.

### DIFF
--- a/.github/module-size-ratchet.json
+++ b/.github/module-size-ratchet.json
@@ -7,30 +7,35 @@
     "2026-08-12: extended to #343 for the same reason as #339. Both issues say line count is not",
     "the point and both are right; both also grew while asking to be split -- plugin.ts +784,",
     "app-provider.ts +377, file-provider.ts +694. Nothing measured that, because every report",
-    "compared against the previous PR rather than against the issue."
+    "compared against the previous PR rather than against the issue.",
+    "2026-08-12 (later): ceilings are CODE lines now -- comments and blanks excluded -- and were",
+    "lowered accordingly. Counting raw lines made this a check against writing things down: of",
+    "app-provider.ts's +377 since #343 was filed, +131 was comments. A ceiling that fails the next",
+    "person who explains a decision, in files whose problem is unexplained decisions, is worse than",
+    "no ceiling."
   ],
   "files": [
     {
       "file": "apps/core-app/src/main/modules/plugin/plugin.ts",
-      "ceiling": 3844,
+      "ceiling": 3249,
       "issue": "#339",
       "note": "~3,060 when #339 was filed; peaked at 4,069 while three extraction PRs were landing."
     },
     {
       "file": "apps/core-app/src/main/modules/quick-ops/index.ts",
-      "ceiling": 2875,
+      "ceiling": 2594,
       "issue": "#339",
       "note": "~2,999 when #339 was filed. Down 124 since, and this holds that."
     },
     {
       "file": "apps/core-app/src/main/modules/box-tool/addon/apps/app-provider.ts",
-      "ceiling": 4600,
+      "ceiling": 3950,
       "issue": "#343",
       "note": "~4,223 when #343 was filed, which is exactly the number that issue cites. +377 since."
     },
     {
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
Taken. All four shapes were real, and one of them bit me while I was fixing it.

`countLines` is a character scan now, tracking block-comment state and quotes. `0995e146e`.

## The four, and one you did not list

| shape | old | new |
| --- | --- | --- |
| one-line block comment followed by code | lost the code | counted |
| block opened after code, mid-line | never entered | entered |
| generator method `*gen()` | read as comment continuation | code |
| wrapped expression beginning with `*` | read as comment continuation | code |
| `/*` inside a string literal | would have opened a comment | not a comment |

The last one was not in your list and follows from the same rewrite; there is a case for it, and one for an escaped quote not ending the string.

## They change nothing today, and that is worth saying

Run against all four guarded files, the character scan and the line version agree **exactly**:

```
plugin.ts 3249   quick-ops/index.ts 2594   app-provider.ts 3950   file-provider.ts 3703
```

So none of these shapes appears in the guarded code and **no ceiling moved**. Fixed anyway — they would have bitten the first time one appeared, and a counter that is right by luck is not right.

## The irony, recorded in the file

Writing the first shape literally in the doc comment **closed the comment and broke the module** — `Cannot find module`, from a stray `*/` inside a block that was explaining stray `*/`. The prose now describes the shapes and the self-test carries them as strings.

22 → **28 self-test cases**. Live control still holds on the real file: six comment lines appended to `app-provider.ts` exit 0, one line of code exits 1.

Issue #334, opened 2026-07-28
Search: unify provider lifecycle, executable registry, and disposal ownership


TalexDreamSoul:
## Priority
P2 — prerequisite architecture work after P0 release gates and active OTA/evidence work are closed.

## Problem
Search provider identity and lifecycle are split across multiple control planes:

- Settings constructs descriptors through `search-provider-registry`.
- Search execution uses a separate provider map in `search-core`.
- Load failures can leave an entry presented as active while execution is degraded.
- `ISearchProvider` does not own a complete async lifecycle/disposal contract.
- provider timers, workers, listeners, streams, and background tasks have inconsistent teardown ownership.

This makes provider availability, Settings diagnostics, execution eligibility, reload, and shutdown describe different states.

## Proposed solution
- Define one async idempotent lifecycle: `load -> start -> stop -> destroy`.
- Introduce explicit registered/loading/ready/degraded/stopped/destroyed states.
- Build one executable registry entry that owns descriptor, implementation/indexed-source adapter, policy, health, configuration, and disposable scope.
- Execute only `ready` entries and expose degraded reasons through the same registry consumed by Settings.
- Add a generic FTS-by-source adapter for eligible indexed sources instead of new SearchCore branches.
- Make `SearchEngineCore.destroy()` await sessions, provider scopes, workers, and transport listener disposal.

## Acceptance criteria
- [ ] A failed load enters degraded state, is excluded from search, and is visible in Settings diagnostics.
- [ ] Repeated load/start/stop/destroy calls are idempotent and do not duplicate handlers.
- [ ] Unload/reload tests prove timers, pollers, listeners, workers, and streams are released/recreated exactly once.
- [ ] Provider/source ID mismatch and collisions fail registration with a typed diagnostic.
- [ ] A generic source adapter queries an eligible indexed source without a hard-coded SearchCore branch.
- [ ] Shutdown awaits session/provider drain and leaves no singleton controller/listener.
- [ ] Existing provider ordering and push-provider behavior remain compatible.

## Dependencies
Do not start until request-scoped sessions, onboarding/storage admission, and single-writer persistence ownership are confirmed complete in the Trellis dependency record.

## Non-goals
- Arbitrary line-count splitting or UI redesign.

TalexDreamSoul:
## 复核依赖闸门：卡住 #334 的是**记账**，不是工作（**PR #1110**，commit `402707e59`），**本 issue 不关**

[...]

PR #1725, opened 2026-08-12
fix(search): await provider teardown instead of firing it

TalexDreamSoul:
A real defect behind #334's fourth and fifth bullets, found while verifying that issue's premises.

## What happens

`SearchProviderRegistry.destroy()` is `async` and does await in-flight loads:

```ts
const activeLoads = [...this.ensureLoadedOperations]
if (activeLoads.length > 0) await Promise.allSettled(activeLoads)
for (const provider of this.providers.values()) {
  const destroyableProvider = provider as DestroyableSearchProvider
  if (destroyableProvider.onDestroy) destroyableProvider.onDestroy()   // ← not awaited
  else provider.onDeactivate?.()
}
```

`AppProvider`'s implementation is **`async onDestroy(): Promise<void>`** and awaits `prepareForSearchIndexShutdown()`. So `destroy()` resolved while the search index was still shutting down.

**TypeScript could not see it.** `ISearchProvider` has no disposal member — #334's fourth bullet, verified: the interface is `id`, `type`, `name?`, `icon?`, `supportedInputTypes?`, `priority?`, `onSearch`, `onActivate?`, `onDeactivate?`, `onExecute?`, `onLoad?`, `info` and nothing else. The registry compensates with a local structural type that declared `onDestroy?: () => void`, so an async implementation looked settled at the call site and there was no floating promise to flag.

Implementations already return both shapes: `AppProvider` async, `EverythingProvider` and `FileSystemWatcher` synchronous.

## The change

`onDestroy?: () => void | Promise<void>` on the local type, and `await` at the call. Nothing else — adding a disposal member to `ISearchProvider` is the interface work #334 is actually about, and it belongs there rather than smuggled into a bug fix.

The loop stays sequential with the `catch` inside it: teardown touches the shared index and the writer, and one provider rejecting must not skip the rest.

## Verification

| | result |
| --- | --- |
| baseline | 9 passed |
| remove the `await` (i.e. today's master) | **1 failed** |
| restored | 9 passed |

Two tests: one holds a teardown open and asserts `destroy()` has not resolved while it is pending, then that it resolves after; one has the first provider reject and asserts the second is still torn down.

`search-engine/` passes: 75 files, 669 tests.

## What is still open on #334

The interface. Also worth recording from the same pass, because it is *good* news the issue does not know: its second bullet — "search execution uses a separate provider map in `search-core`" — **is no longer true**. `SearchCore.registerProvider` delegates straight to `this.providerRegistry.register(...)`, and the only container is the registry's own `providers` Map. The two control planes have been unified since it was filed.

### DIFF
--- a/apps/core-app/src/main/modules/box-tool/search-engine/search-provider-registry.test.ts
+++ b/apps/core-app/src/main/modules/box-tool/search-engine/search-provider-registry.test.ts
@@ -165,6 +165,85 @@ describe('search provider registry', () => {
     vi.useRealTimers()
   })
 
+  /**
+   * `destroy()` fired provider teardown without awaiting it (#334).
+   *
+   * `ISearchProvider` has no disposal member at all, so the registry reaches for `onDestroy`
+   * structurally through a local type that declared it `() => void`. `AppProvider`'s is
+   * `async onDestroy(): Promise<void>` and awaits `prepareForSearchIndexShutdown()` -- so the
+   * registry resolved while the search index was still shutting down, and the `void` declaration
+   * meant TypeScript saw no floating promise to complain about.
+   */
+  it('waits for an async provider teardown before resolving', async () => {
+    let released!: () => void
+    const teardownFinished = vi.fn()
+    const provider = {
+      id: 'slow-teardown',
+      type: 'app' as never,
+      onSearch: vi.fn(async () => ({ items: [] })) as never,
+      onDestroy: vi.fn(async () => {
+        await new Promise<void>((resolve) => {
+          released = resolve
+        })
+        teardownFinished()
+      })
+    }
+
+    const registry = new SearchProviderRegistry({
+      beforeProvidersLoad: async () => undefined,
+      onProvidersReady: async () => undefined,
+      getSearchIndexService: () => null,
+      getTouchApp: () => null,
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
**Correcting this PR's description after the fact.** It claims `destroy()` "resolved while the search index was still shutting down". I have now read the shutdown path and that is not true.

`AppProvider.prepareForSearchIndexShutdown()` is memoised behind `if (!this.shutdownPreparation)`, and `SearchCore.destroy()` awaits it at line 2171 with explicit failure recording — **35 lines before** `await this.providerRegistry.destroy()` at 2206. By the time the registry fires `onDestroy()`, that promise is already settled and the call returns the cached one. Nothing was truncated.

The change is still right, for smaller reasons:

- `onDestroy?: () => void` on a structural type whose known implementation is `async onDestroy(): Promise<void>` hides a floating promise from the compiler.
- `destroy()` is `async` and awaits in-flight loads, then did not await the teardown it triggers. Harmless **only because** `SearchCore` drains `AppProvider` separately and the other two implementations are synchronous — a future async provider without that separate drain would have been fire-and-forget with nothing to notice it.

Latent-hazard and type-correctness, not data loss. Recording it here rather than quietly, because the merged commit message carries the same overstatement.

Issue #1727, opened 2026-08-12
Any folder named Library, Applications, System, private, var, tmp, dev, cache, logs… is silently skipped, at any depth


TalexDreamSoul:
Found while measuring #318. Different problem, worse symptom: **files that should be indexed are silently not indexed, with no error and no diagnostic** — the user just cannot find them.

## What happens

`file-filter-service.ts:171-180`:

```ts
const segments = pathSegments(path)
const directoryName = segments.at(-1) ?? ''
const lowerDirectoryName = directoryName.toLowerCase()
…
if (LOWERCASE_SYSTEM_DIRS.has(lowerDirectoryName)) return 'system-path'
```

The check is on the directory's **own name only**. Where it sits is not considered. `SYSTEM_BLACKLISTED_DIRS` exists to identify OS directories at the filesystem root, and it contains ordinary English words:

```
Library  Applications  System  private  usr  bin  sbin  var  tmp  opt  etc  dev
proc  sys  boot  lib  mnt  media  srv  temp  temporary  cache  .cache  logs  .logs
Windows  ProgramData  AppData  Recovery  Boot  EFI
```

So a folder a user creates inside an indexed root is excluded because of its name.

## Measured against the real default roots

Default roots are `documents · downloads · desktop · music · pictures · videos` (`file-provider.ts:350`). Plausible folders inside them:

| path | result |
| --- | --- |
| `~/Documents/build` | **skipped** — development-path |
| `~/Documents/dev` | **skipped** — system-path |
| `~/Documents/target` | **skipped** — development-path |
| `~/Pictures/private` | **skipped** — system-path |
| `~/Documents/bin` | **skipped** — development-path |
| `~/Music/out` | **skipped** — development-path |
| `~/Documents/logs` | **skipped** — system-path |
| `~/Movies/tmp` | **skipped** — system-path |
| `~/Documents/etc` · `usr` · `var` · `opt` | **skipped** — system-path |
| `~/Pictures/dist` · `cache` | **skipped** |
| `~/Documents/2026 年度报告` | indexed |
| `~/Pictures/旅行` | indexed |

**14 of 16.** And `media`, `Library`, `Applications`, `System` are on the list too — `~/Pictures/media` and `~/Documents/Library` are not unusual names.

## Matching is whole-segment at any depth

Not a substring bug — that part is fine:

| path | |
| --- | --- |
| `/private/var/folders/x` | system-path ✅ correct |
| `/Users/x/private` | system-path ❌ |
[...]

PR #1728, opened 2026-08-12
fix(utils): anchor the system-directory exclusion to the filesystem root

TalexDreamSoul:
Fixes the `system-path` half of #1727.

## The defect

`getTraversalExclusionReason` matched `SYSTEM_BLACKLISTED_DIRS` on the directory's **own name**, at any depth. That list names OS locations — on macOS `Library`, `Application Support`, `Applications`, `System`, `private`, `usr`, `bin`, `sbin`, `var`, `tmp`, `opt`, `etc`, `dev` — and every one is also a word someone may name a folder.

**Silently**: the check runs before `readdir`, so nothing errors, `errorCount` is not incremented, and the file never appears in search — indistinguishable from a file that does not exist.

Measured on one default install: 545 directories under the six default roots, **six excluded this way**, five of them real user content.

## Why anchoring rather than deleting

The correct behaviour was already implemented and already gated. `PATH_PATTERNS.SYSTEM_PATHS` holds root-anchored regexes (`/^\/System\//`, `/^\/private\//`, `/^\/Users\/[^/]+\/Library\//`, the Windows equivalents), and those already catch `/private/var/folders/x`, `/System/Library/Fonts`, `/usr/local/bin`, `~/Library/Caches` — while correctly not catching `~/Documents/tmp`.

Deleting the name check outright would lose the Windows root-only names that have no pattern (`Windows.old`, `Recovery`, `Boot`, `EFI`). Anchoring keeps them.

## Deliberately narrow

Two sibling lists are matched the same way and are **not** touched: `TEMP_BLACKLISTED_DIRS` (`tmp`, `cache`, `logs`) and `DEV_BLACKLISTED_DIRS` (`build`, `out`, `bin`, `target`, `dist`, `node_modules`). The observed Codex `tmp` folders are still excluded, now as `cache-path`. Whether an ordinary word should exclude a folder under `~/Documents` is a policy question, and a test pins the current answer rather than changing it.

## Verification

| plant | result |
| --- | --- |
| remove the anchor (today's master) | **3 failed** |
| make the anchor reject the filesystem root itself | **1 failed** |

`pnpm -C packages/utils` — the three affected files are 12 tests green. Running vitest from the repo root instead makes two unrelated suites fail on `ENOENT: plugin/sdk/channel-client.ts`; they resolve source paths relative to cwd.

`SYSTEM_BLACKLISTED_DIRS` is platform-scoped — on macOS only `MACOS_SYSTEM_DIRS` applies, which is worth knowing before writing tests against these lists.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filesystem filtering so system-directory exclusions apply only to directories directly under a filesystem root.
  * Prevented similarly named folders within user directories from being incorrectly excluded.
  * Preserved existing exclusions for temporary, development, and other recognized system paths.
* **Tests**
  * Added coverage for POSIX and Windows root paths, nested directories, and exclusion behavior.

### DIFF
--- a/packages/utils/__tests__/file-filter-traversal-anchor.test.ts
+++ b/packages/utils/__tests__/file-filter-traversal-anchor.test.ts
@@ -42,6 +42,20 @@ describe('#1727 system-dir anchoring', () => {
     }
   })
 
+  /**
+   * `pathSegments` drops separators, so `usr` and `/usr` are both one segment. Without the root
+   * marker a caller passing a relative path — `scanDirectory` does not require an absolute one —
+   * would have it classified as a filesystem root. Found by CodeRabbit on #1728.
+   */
+  it('does not treat a relative path as the filesystem root', () => {
+    for (const name of systemNames) {
+      expect(reason(name), name).not.toBe('system-path')
+      expect(reason(`./${name}`), `./${name}`).not.toBe('system-path')
+      expect(reason(`a/${name}`), `a/${name}`).not.toBe('system-path')
+      expect(reason(`C:${name}`), `C:${name}`).not.toBe('system-path')
+    }
+  })
+
   it('keeps the root-anchored patterns doing their job below the root', () => {
     // From PATH_PATTERNS.SYSTEM_PATHS, which is where root-level system paths were always handled;
     // the name list was only ever adding the unanchored half. Platform-split for the same reason
--- a/packages/utils/__tests__/file-filter-traversal-anchor.test.ts
+++ b/packages/utils/__tests__/file-filter-traversal-anchor.test.ts
@@ -1,50 +1,73 @@
-import { fileFilterService } from '../common/file-filter-service'
 import { describe, expect, it } from 'vitest'
-const r = (p: string): string => String(fileFilterService.getTraversalExclusionReason(p, undefined as never) ?? '—')
+import { fileFilterService } from '../common/file-filter-service'
+import { SYSTEM_BLACKLISTED_DIRS } from '../common/file-scan-constants'
+
+const reason = (p: string): string =>
+  String(fileFilterService.getTraversalExclusionReason(p, undefined as never) ?? '—')
+
+/**
+ * The names are read from the runtime set rather than written down.
+ *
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
CI caught a platform assumption in my own tests — pushed `0b1c350dd`.

The first version asserted `/private` and `C:/Windows` literally, passed on macOS, and failed on Linux: `SYSTEM_BLACKLISTED_DIRS` is assembled per platform, and `private` is in `MACOS_SYSTEM_DIRS` only. I noted that caveat in the PR description after hitting it once on `C:/Windows`, then left the rest of the file asserting macOS names — the note did not become a fix.

The test now reads the runtime set and checks the **property** rather than the membership: every listed name is still excluded at `/name`, and none of them reports `system-path` at `/Users/someone/Documents/name`. That is what the change actually does, and it holds on any platform. Below-root patterns are still asserted, split by platform because they genuinely are platform-specific.

A vacuity guard was added too — the derived list must be non-empty, otherwise both loops would pass by iterating nothing.

Removing the anchor fails two of the six. The production change is unchanged.

TalexDreamSoul:
Taken, and it was reachable — `0c027b74c`.

`pathSegments` drops the separators, so `/usr` and a relative `usr` are both **one segment**, and my check accepted either. Measured before the fix:

| path | before | after |
| --- | --- | --- |
| `usr` | **system-path** | — |
| `Library` | **system-path** | — |
| `etc` | **system-path** | — |
| `./usr` | — | — |
| `a/usr` | — | — |
| `C:Library` | — | — |
| `/usr` | system-path | system-path |

So three of the seven were misclassified. `./usr` and `a/usr` were already safe (they normalise to more than one segment), and `C:Library` is drive-relative — it stays a single segment with no `/`, so it is correctly not a root either before or after.

The original path now carries the marker: a leading `/` for the one-segment case, the existing drive-letter segment for the two-segment case. A regression case loops every name in the runtime set across all four shapes.

Worth noting for the record: on master this is **not** a new exposure — an unanchored name check excluded relative `usr` too. The anchor simply had to be as strict as the thing it replaced.

Removing the marker requirement fails the new case. `packages/utils` is green.

PR #1730, opened 2026-08-13
fix(search): let a commit-denied cache entry report why it was denied

TalexDreamSoul:
Removes #346's second and last blocker. With #1729 the counters became readable; this makes them mean something.

## The defect

`classifySearchCacheLookup` has always had a `revision-mismatch` outcome — "the entry was there but the index moved on". The comment at the classify site says it is the point of the whole taxonomy:

> The reason is the point, not the hit rate: `revision-mismatch` means an index commit denied the entry, which is a different story from a query nobody repeated.

**It was unreachable.** `handleSearchIndexCommit` cleared the entire cache synchronously, so by the time any lookup compared revisions the entry was gone and the miss was filed as `absent` — the same bucket as a query nobody repeated.

That is the distinction #346 exists to measure, and its only evidence read as its own opposite: *"nobody repeats queries, remove the cache"* when the truth may be *"repeats happen and the index keeps denying them"*.

## The change

The commit no longer clears. Entries are rejected on revision at lookup and dropped there, so the reason is counted where it occurs.

| | before | after |
| --- | --- | --- |
| `misses.absent` | 3 | **2** |
| `misses['revision-mismatch']` | **0** | **1** |
| `invalidations['index-commit']` | 2 | 1 |

**Nothing stale is served.** The outcome is computed before the hit branch and that path is unchanged — a `revision-mismatch` never reaches `materializeCachedSearchResult`. The cost is that a superseded entry holds its slot until it is looked up, expires at the 5s TTL, or is evicted by the existing LRU bound.

## Verification

| plant | result |
| --- | --- |
| restore `searchCache.clear()` in the commit handler | **1 failed** |

`search-engine/`: 75 files, 670 tests.

The test also asserts that a second lookup after the mismatch **hits**, not mismatches again — one `revision-mismatch` per commit per query, not one per lookup.

## One thing deliberately not asserted, and it is in the test

The `searchCache.delete(cacheKey)` at the lookup is **not** observable through the telemetry: the miss re-runs the search and re-caches under the same key, so the entry is replaced either way. A plant that removed the delete passed, which is recorded in the test rather than papered over. The delete is kept for the paths that do **not** re-cache — an aborted or failed search — where the superseded entry would otherwise hold its slot until the TTL.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search cache handling when indexed data changes.
  * Cache telemetry now distinguishes stale entries, expired entries, and entries that were never cached.
  * Rejected stale results are removed during lookup, and repeated queries can be cached again for improved hit rates.

### DIFF
--- a/apps/core-app/src/main/modules/box-tool/search-engine/search-core.contracts.test.ts
+++ b/apps/core-app/src/main/modules/box-tool/search-engine/search-core.contracts.test.ts
@@ -887,21 +887,40 @@ describe('SearchEngineCore facade contracts', () => {
 
     // Exactly two, because two entries were in the cache. Counted by entries dropped rather than
     // by calls -- a per-call implementation reports 1 here.
-    expect(snapshot.invalidations['index-commit']).toBe(2)
-    expect(snapshot.entriesDropped).toBe(2)
+    // One entry was superseded and dropped at its own lookup; the other was never asked for.
+    expect(snapshot.entriesDropped).toBe(1)
 
     /*
-     * And this is the finding. `revision-mismatch` is unreachable from here: `markCommitted` bumps
-     * the revision and notifies listeners synchronously, and this facade's listener clears the
-     * whole cache before returning. By the time any lookup compares revisions the entry is already
-     * gone, so the miss is `absent`.
+     * The attribution #346 turns on, and the reason this test exists.
      *
-     * That matters for the report #346 asks for. A repeat denied by a commit is filed under the
-     * same reason as a query nobody repeated, so the distribution understates the cache's value
-     * and cannot carry the keep/remove decision as it stands.
+     * `revision-mismatch` used to be unreachable: the commit listener cleared the whole cache
+     * before returning, so by the time a lookup compared revisions the entry was gone and the miss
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
Measured the one real cost of this change rather than leaving it as prose. `1ba8b6f`-range pushed.

I wrote "a superseded entry holds its slot until it is looked up, expires at the 5s TTL, or is evicted by the existing LRU bound". That is true and it is also exactly the kind of sentence that hides a worst case — a cache filling with entries that can only ever miss.

It cannot happen:

- `SEARCH_CACHE_MAX_SIZE = 100`, a hard cap enforced on every insert.
- Eviction picks the **oldest timestamp**, and after a commit the oldest entries are the superseded ones — so it drains **stale-first**, one per insert.
- No stale entry is ever served: the outcome is computed before the hit branch, unchanged by this PR.

Test: 120 queries to fill past the cap, one commit to supersede everything, then 40 fresh keys. Zero hits, zero revision mismatches on the fresh keys, and `lru-evict > 0` — which is only possible if the cache stayed at its cap throughout.

Raising the cap to 100,000 fails it, so the test holds the bound rather than the arithmetic happening to work out.

`search-engine/`: 75 files, **671** tests.

PR #1763, opened 2026-08-20
fix(utils): require a project marker before ordinary-word dirs exclude

### LINKED ISSUE #1727: Any folder named Library, Applications, System, private, var, tmp, dev, cache, logs… is silently skipped, at any depth

TalexDreamSoul:
Closes the last two acceptance criteria of #1727. The first two — root-anchoring `SYSTEM_BLACKLISTED_DIRS`, and keeping `~/Library` / `%APPDATA%` excluded — landed in #1728 and #1731.

## The decision #1727 asked for

> A decision recorded on whether development names apply under user document roots.

**Not by name alone.** `node_modules` is nobody's document folder and keeps excluding wherever it appears. `build`, `dist`, `out`, `bin`, `target`, `coverage`, `logs`, `tmp`, `temp` and `cache` are ordinary English words, and they now exclude only where a project marker (`package.json`, `Cargo.toml`, `go.mod`, `Makefile`, `.git`, `*.csproj`, …) sits beside the directory.

That is the signal ripgrep, fd and VS Code use for the same question, and it is free here: `scanDirectoryInto` has already called `readdir` on the parent before it descends, so the sibling list is in hand. One `map`, no extra syscall.

| path | before | after |
| --- | --- | --- |
| `~/Documents/build` | skipped, silently | indexed |
| `~/Documents/build/2026` | skipped, silently | indexed |
| `~/Movies/tmp`, `~/Pictures/cache`, `~/Documents/logs` | skipped | indexed |
| `~/Projects/app/dist` (next to `package.json`) | skipped | skipped |
| `~/Documents/x/node_modules` | skipped | skipped |
| `/usr/bin`, `/var/tmp` | skipped | skipped |

## The half a leaf-name-only fix would have missed

`PATH_PATTERNS.DEV_PATHS` carries `/build\//` and `CACHE_PATHS` carries `/\/tmp\//` — unanchored, so they match an **ancestor** segment. Relaxing the leaf name alone would have made `~/Documents/build` indexable while everything under it stayed excluded: a fix that passes a shallow test and still loses the user's files.

Those two pattern sets are therefore skipped for callers that supply a sibling list. Such a caller is walking top-down and has already had the per-level rule applied at every ancestor, and every name the patterns carry is either in `DEV_BLACKLISTED_DIRS` / `TEMP_BLACKLISTED_DIRS` or starts with a dot, so nothing is lost. `SYSTEM_PATHS` stays unconditional — it is the one of the three that catches what the per-level name rule cannot, `/usr/bin` being exactly that case once `bin` is relaxed.

## Blast radius

`TraversalContext` is optional and only `scanDirectoryInto` passes it. Index upserts, manual adds and the file-level containing-path check read as "cannot tell" and keep the pre-#1727 answer, so they are byte-for-byte unchanged. `undefined` and `[]` are deliberately distinct: never-looked vs. read-and-found-nothing.

One reason string changes: `/usr/bin` was reported as `development-path` and is now `system-path`. No caller branches on the value — every one of them tests `=== null` or truthiness.

## Verification

- `packages/utils` full suite: **189 files, 1378 passed, 1 skipped**
- 14 new tests in `file-filter-project-context.test.ts`, both directions
- `pnpm lint` (the utils CI gate) clean
- **Negative control:** with the new constants kept but the service logic reverted to HEAD, the 4 tests that encode the fix fail and the 10 that guard unchanged behaviour pass. The tests fail on the bug, not just pass on the fix.

Fixes #1727

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved directory scanning to recognize project context using nearby project files and metadata.
  - Reduced false exclusions for ordinary directories that share names with development or cache folders.
  - Preserved strict exclusions for system paths, hidden entries, bundles, dependency folders, and configured blacklists.
  - Maintained consistent filtering when project context is unavailable, including nested paths and case variations.

- **Tests**
  - Added comprehensive coverage for project-aware filtering and traversal scenarios.

### DIFF
--- a/packages/utils/__tests__/file-filter-project-context.test.ts
+++ b/packages/utils/__tests__/file-filter-project-context.test.ts
@@ -1,0 +1,155 @@
+import { describe, expect, it } from 'vitest'
+import {
+  CONTEXT_DEPENDENT_BLACKLISTED_DIRS,
+  DEV_BLACKLISTED_DIRS,
+  TEMP_BLACKLISTED_DIRS,
+} from '../common/file-scan-constants'
+import { fileFilterService } from '../common/file-filter-service'
+
+/**
+ * #1727's last acceptance criterion: a decision on whether development names apply under a user's
+ * document roots.
+ *
+ * The decision is *not by name alone*. `node_modules` is nobody's document folder and keeps
+ * excluding everywhere; `build`, `dist`, `out`, `bin`, `target`, `coverage`, `logs`, `tmp`, `temp`
+ * and `cache` are ordinary English words, and they exclude only where a project marker sits beside
+ * them — the same signal ripgrep, fd and VS Code use, and free here because the traversal has
+ * already read the parent before it descends.
+ *
+ * Both directions are asserted below, because only one of them is the silent one: a wrongly-kept
+ * exclusion produces no error, no `errorCount`, and a search result identical to "file not there".
+ */
+
+/** A directory whose siblings were read and contain no project marker. */
+const withoutProject = (p: string, siblings: readonly string[] = ['notes.txt']): string =>
+  String(fileFilterService.getTraversalExclusionReason(p, undefined, { siblingNames: siblings }) ?? '—')
+
+/** The same, next to a `package.json`. */
+const withProject = (p: string, marker = 'package.json'): string =>
+  String(
+    fileFilterService.getTraversalExclusionReason(p, undefined, {
+      siblingNames: [marker, 'src', 'README.md'],
+    }) ?? '—',
+  )
+
+/** No context at all — every caller other than the traversal. */
+const withoutContext = (p: string): string =>
+  String(fileFilterService.getTraversalExclusionReason(p) ?? '—')
+
+const ORDINARY_WORDS = [...CONTEXT_DEPENDENT_BLACKLISTED_DIRS]
+
+describe('#1727 ordinary-word dirs under a user root', () => {
+  it('has names to check, so the loops below are not vacuous', () => {
+    expect(ORDINARY_WORDS.length).toBeGreaterThan(0)
+  })
+
+  /**
+   * The set only ever weakens where an existing list applies. A name here that is on neither list
+   * would be inventing policy in a constant nobody reads.
+   */
+  it('only relaxes names the two blacklists already own', () => {
+    for (const name of ORDINARY_WORDS) {
+      expect(
+        DEV_BLACKLISTED_DIRS.has(name) || TEMP_BLACKLISTED_DIRS.has(name),
+        name,
+      ).toBe(true)
+    }
+  })
+
+  it('indexes them when nothing beside them says "project"', () => {
+    for (const name of ORDINARY_WORDS) {
+      const p = `/Users/someone/Documents/${name}`
+      expect(withoutProject(p), p).toBe('—')
+    }
+  })
+
+  it('still excludes them beside a project marker', () => {
+    for (const name of ORDINARY_WORDS) {
+      const p = `/Users/someone/Projects/app/${name}`
+      expect(withProject(p), p).not.toBe('—')
+    }
+  })
+
+  /**
+   * The half a leaf-name-only fix would have missed. `PATH_PATTERNS.DEV_PATHS` carries `/build\//`
+   * unanchored, so it matches an *ancestor* segment: `~/Documents/build` would come back indexed
+   * while everything under it stayed excluded.
+   */
+  it('indexes what is below them too, not just the folder itself', () => {
+    const p = '/Users/someone/Documents/build/2026'
+    expect(withoutProject(p, ['2026', 'notes.txt']), p).toBe('—')
+    expect(withoutProject('/Users/someone/Documents/tmp/receipts'), 'tmp/receipts').toBe('—')
+  })
+
+  it('recognises project markers other than package.json', () => {
+    for (const marker of ['Cargo.toml', 'go.mod', 'Makefile', 'pom.xml', '.git'])
+      expect(withProject('/Users/someone/Documents/target', marker), marker).not.toBe('—')
+  })
+
+  /** `App.csproj` is named per project, so the marker list matches it by suffix. */
+  it('recognises per-project marker names by suffix', () => {
+    expect(withProject('/Users/someone/Documents/bin', 'App.csproj')).not.toBe('—')
+  })
+
+  it('matches markers case-insensitively', () => {
+    expect(withProject('/Users/someone/Documents/build', 'makefile')).not.toBe('—')
+  })
+})
+
+describe('#1727 what the relaxation must not touch', () => {
+  it('keeps node_modules excluded with or without a project beside it', () => {
+    const p = '/Users/someone/Documents/node_modules'
+    expect(withoutProject(p)).toBe('development-path')
+    expect(withProject(p)).toBe('development-path')
+    expect(withoutContext(p)).toBe('development-path')
+  })
+
+  /**
+   * `bin` is on the macOS system list *and* the dev list. Relaxing the dev half must not hand back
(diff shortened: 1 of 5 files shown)


TalexDreamSoul:
## Wiring verification

I said in review that this PR had no end-to-end verification, and that the fixture experiment I tried did not run — the control file came back unindexed, which proved the experiment was broken rather than the filter aggressive. Here is what I did instead, and what it does and does not establish.

The risk worth checking is not "is the rule right" — the unit tests cover that. It is **"is the new context actually supplied by the real scanner, at the right level, in the shape the filter reads"**. A context that no production caller populates would leave `walksEveryLevel` permanently false and this PR would be inert, while every unit test still passed. That is a static property, so I traced it statically rather than staging another live scan.

**The capability is derived, not passed as a flag:**

```
file-filter-service.ts:340   const walksEveryLevel = context?.siblingNames !== undefined;
```

**The recursion supplies it from entries it already has** — this is the line that makes the PR live, and the one absent on master:

```
file-scan-utils.ts:457   const entryNames = entries.map((entry) => entry.name);
file-scan-utils.ts:459     await scanDirectoryInto(subDir, opts, excludePaths, depth + 1, out, stats, sink, entryNames)
```

so every directory below a scan root is judged against its own sibling list, and a scan root — which has no parent that was read — passes `undefined` and keeps the stricter pre-existing answer. That asymmetry is the right way round: the widening never applies at a level where the evidence for it was never gathered.

**And the production index path is that same code**, not a parallel implementation:

```
file-provider.ts:841,922        scanDirectory: (rootPath, excludePaths, runOptions) => …
addon/files/utils.ts:42-47      → globalScanDirectory   (a field remap, no traversal logic of its own)
file-scan-utils.ts              → scanDirectoryInto
```

consumed by both `file-provider-full-scan-run-service.ts:76` and `file-provider-reconciliation-run-service.ts:193`, so full scans and reconciliation runs both go through it.

Worked through the fixture case by hand against that chain: `Documents/build` is judged against `['build','notes']`, no project marker, indexed; `Projects/app/dist` is judged against `['package.json','dist']`, marker present, excluded. That is the intended semantics.

**What this does not establish.** No live scan was run and no rows were read out of `search-index.db` — this is a trace of the call graph, not an observation of the outcome. It rules out the inert-context failure, which was the one I actually doubted; it would not catch something that goes wrong only against a real filesystem, such as a permission-denied directory or a symlink loop changing which entries `readdir` returns.

TalexDreamSoul:
Disclosure against #318, which scopes the remaining indexing-memory risk to exactly the function this PR changes.

That issue's residual is that `fs.readdir(dirPath, { withFileTypes: true })` is unbounded for a single directory with huge fan-out — measured there at 95.7 MiB and climbing for 100k entries in one directory. **This PR adds an allocation on that path**:

```js
const entryNames = entries.map((entry) => entry.name);
```

retained for the duration of each directory's recursion, alongside `entries`. Measured:

| entries | `entries` | added `entryNames` | overhead |
| ---: | ---: | ---: | ---: |
| 100,000 | 19.08 MiB | 0.76 MiB | **4.0%** |
| 200,000 | 41.20 MiB | 1.53 MiB | **3.7%** |

The strings are shared with the Dirent objects, so this is the pointer array only. A Dirent stand-in was used, so the ratio is the reliable part, not the absolute MiB.

Not proposing a change here. The comment at that line already says the map "costs one map, no syscall", which is true about syscalls and silent about memory — worth saying out loud on an issue that is specifically about this function's retention. It is also compatible with the `opendir` streaming fix #318 is weighing: names-only retention is ~4% of the current per-directory cost, so collecting names while streaming would keep this PR's filter working and still take the big win.

Also noted on #318: that issue's tree-level numbers were measured on the assumption the dev/temp filter half stays as-is, which is what this PR changes. They should be re-measured after this lands rather than carried forward.

PR #1769, opened 2026-08-20
test(nexus): stop check-worker-bundle demanding icons this app cannot draw

### LINKED ISSUE #1768: 8 of 9 nexus icon shortcuts emit no CSS — they alias onto @iconify-json/ri, which nexus does not install
Found while checking why `App suites (nexus)` went red on master. The test failure is a symptom; this is the cause, and it is shipping.

## What is broken

`apps/nexus/uno.config.ts` `shortcuts` rewrites nine icon names. Eight of them target `i-ri-*`, and **nexus does not install `@iconify-json/ri`**, so UnoCSS expands the shortcut, fails to match the target, and emits nothing. The class renders no icon at all.

Measured by generating CSS with nexus's own config, not inferred:

```
  BLANK   i-carbon-settings              -> i-ri-settings-line
  BLANK   i-carbon-fire                  -> i-ri-fire-line
  BLANK   i-ri-settings-3-line           -> i-ri-settings-line
  BLANK   i-ri-settings-3-fill           -> i-ri-settings-fill
  BLANK   i-carbon-data-vis-1            -> i-ri-bar-chart-line
  BLANK   i-carbon-deployment-pattern    -> i-ri-route-line
  BLANK   i-carbon-queued                -> i-ri-time-line
  BLANK   i-carbon-ai-status             -> i-ri-robot-2-line
  ok      i-simple-icons-linux           -> i-carbon-linux-alt

  8 of 9 icon shortcuts emit no CSS
```

UnoCSS says so itself: `[unocss] unmatched utility "i-ri-settings-line" in shortcut "i-carbon-settings"`.

The one that works is the one aliasing onto `carbon`, which *is* installed.

## Reachability

`i-carbon-settings` appears in **12 files** under `apps/nexus/app`, `i-carbon-ai-status` in 4, and `i-carbon-fire` / `i-carbon-data-vis-1` / `i-carbon-deployment-pattern` / `i-carbon-queued` in one each. Roughly 20 call sites render a blank where an icon is intended. `i-ri-settings-3-line` and `-3-fill` have no call sites, so those two aliases are simply dead.

## Root cause

From `apps/nexus`, with `cib` / `carbon` / `logos` / `twemoji` as positive controls so the absence is not a broken probe:

```
  @iconify-json/cib            RESOLVES
  @iconify-json/carbon         RESOLVES
  @iconify-json/logos          RESOLVES
  @iconify-json/twemoji        RESOLVES
  @iconify-json/ri             unresolvable
  @iconify-json/simple-icons   unresolvable
```

`@iconify-json/ri` and `@iconify-json/simple-icons` are in the lockfile, but declared by `apps/core-app` and `plugins/touch-image` — pnpm isolates them there. `apps/nexus/.npmrc` sets `shamefully-hoist=true`, which flattens *nexus's own* tree and does not reach another workspace package's dependencies.

So the shortcut comment — "Keep common glyphs compact because Uno emits every referenced icon into the shared entry CSS" — is aiming at a real problem (entry-CSS size) but redirecting to a collection that was never available here. The size saving is real only because the icon disappears.

## The test enforces the breakage

`apps/nexus/build/check-worker-bundle.test.ts:94` asserts each of these nine aliases is present:

```js
expect(unoSource).toContain("['i-carbon-settings', 'i-ri-settings-line']")
expect(unoSource).toContain("['i-carbon-ai-status', 'i-ri-robot-2-line']")
// …seven more
```

and separately requires `i-simple-icons-apple`, `i-simple-icons-linux`, `i-ri-safari-line`, `i-ri-edge-line`, `i-ri-fingerprint-2-line`, `i-ri-figma-fill` to appear in route sources — six more names from the two unresolvable collections — while forbidding `i-cib-apple`, `i-cib-linux`, `i-cib-safari`, `i-cib-microsoft-edge`, which are the ones that *do* resolve (`i-cib-apple` generates 1,629 bytes of real icon CSS).

That is why `fbb19a780` and `535be553e` turned `App suites (nexus)` red: those commits moved the code onto icons that render, and the test still demands the ones that do not. **The code is right and the test is wrong**, which is worth stating explicitly because the failure reads the other way round.

## Suggested direction

1. Point the eight shortcuts at `carbon` equivalents (installed, and the collection the source names already use), or delete them so the original `i-carbon-*` name resolves directly. Deleting is the smaller change and costs only the entry-CSS saving that was never actually being realised.
2. Drop the two dead aliases for `i-ri-settings-3-line` / `-3-fill`.
3. Update `check-worker-bundle.test.ts` to assert on collections nexus installs. Its `sharedEntryCssBudget` assertions are the part worth keeping; the icon-name lists encode a policy that no longer holds.
4. If `ri` / `simple-icons` are genuinely wanted, the fix is to declare them in `apps/nexus/package.json` — but that adds two collections to the bundle, which is the opposite of what the shortcut block was trying to do.

## Acceptance

- [ ] Generating CSS for every `i-*` shortcut in `uno.config.ts` emits a non-empty rule.
- [ ] The guard test asserts icon names against the collections nexus actually installs, and a name from an uninstalled collection fails it.
- [ ] `App suites (nexus)` is green on master.

TalexDreamSoul:
Fixes #1768. **Rebased onto `9237a8282`** — that commit deleted the icon shortcuts and extended `check-icon-collections.mjs` to scan `uno.config.ts`, which is the same conclusion this branch had reached. Its `uno.config.ts` change is therefore dropped; only the test half remains, and master is red because of it.

## master currently contradicts itself

`9237a8282` removed the icon shortcuts. `check-worker-bundle.test.ts` still requires them:

```ts
expect(unoSource).toContain("['i-simple-icons-linux', 'i-carbon-linux-alt']")
expect(unoSource).toContain("['i-carbon-settings', 'i-ri-settings-line']")
// …seven more
```

Reproduced by putting master's own copy of the file back on top of master:

```
AssertionError: expected 'import {\n  defineConfig,\n  presetAt…'
  to contain '['i-simple-icons-linux', 'i-carbon…'
Tests  1 failed | 18 passed (19)
```

The same test separately requires ten route files to contain `i-simple-icons-apple`, `i-ri-safari-line`, `i-ri-edge-line`, `i-ri-fingerprint-2-line` and `i-ri-figma-fill`, while forbidding `i-cib-*`. That is backwards — `i-cib-apple` emits 1,629 bytes of real icon CSS and the `ri`/`simple-icons` names emit none — and it contradicts `check-icon-collections.mjs` in the same directory: **a file satisfying the old "must contain" list is an offender to the new guard.** Both tests cannot pass.

## What changes

The name lists are removed rather than corrected, because the question they were trying to answer now has an owner that answers it from `package.json`.

In their place, an assertion that every `i-*` shortcut generates non-empty CSS. This is complementary, not duplicated:

- `check-icon-collections.mjs` reads the class **as written**, so it catches a shortcut pointing at a missing *collection* — which is the case that actually happened.
- Generating CSS also catches a shortcut pointing at a name that does not exist *inside an installed collection*, which a collection scan reads as fine.

It carries its own positive control, since a loop over zero shortcuts and a broken generator read identically:

```ts
const installed = await uno.generate('i-carbon-settings', ...)
expect(installed.css.length).toBeGreaterThan(0)
const uninstalled = await uno.generate('i-ri-settings-line', ...)
expect(uninstalled.css.length).toBe(0)
```

The `sharedEntryCssBudget` assertions are untouched.

## Verification

| check | result |
| --- | --- |
| nexus suite on top of `9237a8282` | **202 files / 1120 tests pass** |
| control — master's own copy of this file, on master | 1 failed / 18 passed, the contradiction above |
| `node build/check-icon-collections.mjs` | `ok — installed: carbon, cib, logos, twemoji`, exit 0 |
| eslint | clean |

Earlier revisions of this branch also fixed `uno.config.ts` and confirmed on CI that restoring the six previously-blank icons does **not** trip `sharedEntryCssBudget` — `Nexus (build + checks)` passed with the real build. That data point still stands for `9237a8282`, which makes the same change.

Diff is now one file, −49/+39.

### DIFF
--- a/apps/nexus/build/check-worker-bundle.test.ts
+++ b/apps/nexus/build/check-worker-bundle.test.ts
@@ -1,8 +1,10 @@
 import { describe, expect, it } from 'vitest'
 import { existsSync, readFileSync, statSync } from 'node:fs'
 import { join } from 'node:path'
+import { createGenerator } from 'unocss'
 import enLocale from '../i18n/locales/en'
 import zhLocale from '../i18n/locales/zh'
+import unoConfig from '../uno.config'
 
 const nexusRoot = join(import.meta.dirname, '..')
 const shotsRoot = join(nexusRoot, 'public/shots')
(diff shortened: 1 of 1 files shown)


TalexDreamSoul:
Superseded by `b3d751a3d`, which repoints this test at a contract that can hold, together with `9237a8282` (shortcuts deleted) and `0f0ff5768` (`check-icon-collections.mjs`). Master's `App suites (nexus)` contradiction is gone — the assertion count for the old `i-simple-icons-apple` / `i-ri-safari-line` / `['i-carbon-settings', 'i-ri-…]` set is now 0. Closing rather than rebasing: the remaining delta is too small to justify a third mechanism.

Recording the one gap the new contract does not cover, so it is a decision rather than an oversight.

`not.toContain('i-ri-')` plus the collection scan answer "is the collection installed". Neither answers "does this icon exist inside an installed collection". A typo like `['i-carbon-settings', 'i-carbon-setttings']` passes both — the collection is `carbon`, which *is* installed — and renders the same blank that started #1768.

The assertion that closes it is three lines and needs no allowlist:

```ts
const uno = await createGenerator(unoConfig)
for (const [from] of (unoConfig.shortcuts ?? []).filter(e => Array.isArray(e) && e[0].startsWith('i-')))
  expect((await uno.generate(from, { preflights: false })).css.length, from).toBeGreaterThan(0)
```

It needs its own positive control, because with zero icon shortcuts the loop and a broken generator read identically:

```ts
expect((await uno.generate('i-carbon-settings', { preflights: false })).css.length).toBeGreaterThan(0)
expect((await uno.generate('i-ri-settings-line', { preflights: false })).css.length).toBe(0)
```

Worth having whenever a shortcut comes back; not worth a PR while the block is empty. #1768 can close on `b3d751a3d` as far as the reported defect goes.

One measurement from this branch that still applies to the shipped fix, since it was the only real risk in it: restoring the six previously-blank icons adds ~6.8 KiB raw of icon CSS, and `Nexus (build + checks)` **passed** with the real build — `sharedEntryCssBudget` (278 KB / 50 KB gzip) was not tripped.

Issue #1748, opened 2026-08-16
[release/database] Prove default-on search-index split in an isolated CoreApp profile


TalexDreamSoul:
## Priority
P0 — release evidence gate

## Observed gap
`TUFF_DB_SEARCH_SPLIT_ENABLED` defaults on; `=0` is the emergency rollback. The active task `07-28-migrate-search-index-split-write-paths` still has no direct isolated-profile evidence for every remaining writer, startup ordering, or rollback. Source/type checks and historical runs cannot establish current application behavior after the remaining migration work.

The incomplete-migration failure mode is silent data divergence: a provider writes `database.db` while readers consume `search-index.db`.

## Required work
1. Reconcile every 2d/2e, provider, embedding, first-launch reindex, and orphan-cleanup write against the worker-owned connection under the default-on topology.
2. Run CoreApp against a disposable profile with split enabled; record first-launch rebuild, populated `search-index.db`, matching application/file counts and representative results, startup ordering, and absence of `SQLITE_BUSY`/WAL storms.
3. Exercise `TUFF_DB_SEARCH_SPLIT_ENABLED=0` only after quiescing/reconciling writes, then prove post-restart result/count parity on the shared-file topology.

## Acceptance criteria
- [ ] Every inventoried moved-table writer executes through the worker under default split; no pre-ready main-thread fallback exists.
- [ ] Disposable-profile evidence maps first-launch reindex, app/file result parity, populated `search-index.db`, healthy queue/WAL behavior, and representative queries to one reproducible run.
- [ ] The `=0` rollback has a documented quiesce/reconcile/restart procedure and proves post-restart parity.
- [ ] Release review treats this issue as blocking until the direct evidence is attached; typecheck alone cannot close it.

## Evidence and relationships
- Task contract: `.trellis/tasks/07-28-migrate-search-index-split-write-paths/prd.md`
- Global ordering: `docs/plan-prd/TODO.md`
- Current audit: `docs/engineering/reports/maintenance-audit-2026-08-16.md`
- #1745 fixed the stale default-off documentation only; it did not supply runtime proof.
- Coordinate with #351 on durable writer ownership; do not add a second writer.

TalexDreamSoul:
Working the part of this that does not need a run. The static half turned up a live instance of exactly the failure mode in the issue body — a writer on `database.db` while readers consume `search-index.db` — filed separately as #1770.

## What the static pass settles

The core routing mechanism is sound and **fails closed**, which is the important part:

[...]

PR #1771, opened 2026-08-20
fix(core-app): clean the file index on the right database, and stop it deleting apps

### LINKED ISSUE #1770: cleanupFileIndex() writes to the primary DB under the default-on search split: it deletes the app catalog and never clears the file index
Found while working the static half of #1748, whose stated failure mode is *"a provider writes `database.db` while readers consume `search-index.db`"*. This is that, in a path outside the migration's own 2d/2e writer inventory.

## What is wrong

`apps/core-app/src/main/service/storage-maintenance.ts` never consults the split. Its only accessor is:

```ts
async function getDb(): Promise<LibSQLDatabase<...> | null> {
  return databaseModule.getDb()   // primary, unconditionally
}
```

Grepping the whole file for `getSearchDb`, `isSearchSplitEnabled`, `searchIndexWriter`, `execWrite` or `searchDb` returns **0**. `DatabaseModule` exposes both `isSearchSplitEnabled()` (`modules/database/index.ts:130`) and `getSearchDb()` (`:145`) — they are simply not used here.

`cleanupFileIndex()` then does, with **no `where` clause**:

```ts
const removeAll = async (table: DeleteTable) => {
  const rows = await db.select({ count: sql<number>`count(*)` }).from(table)
  removedCount.push(rows[0]?.count ?? 0)
  await db.delete(table)
}

await removeAll(fileIndexProgress)
await removeAll(fileExtensions)
await removeAll(files)
await removeAll(scanProgress)
```

## Two failures at once, and they point opposite ways

Under `TUFF_DB_SEARCH_SPLIT_ENABLED` (default **on**):

**It does not clear what it is named for.** File-provider rows live in `search-index.db`; this deletes from `database.db`. The `removedCount` reported back to the UI is counted from the wrong file, and the live file index the search engine reads is untouched.

**It deletes something it was never meant to touch.** The app catalog lives in the primary `files` / `file_extensions`, deliberately. `app-provider.ts:840-852` says so in its own words:

> The app CATALOG (files/file_extensions rows of type 'app', **including user-authored managed entries**) lives on the PRIMARY db … manual entries are **user data that must not move into the rebuildable search-index.db**.

`removeAll(files)` and `removeAll(fileExtensions)` delete every row of those tables on the primary. Auto-discovered apps return on the next scan; entries the code explicitly classifies as user data that must not be rebuilt do not have a disk source to be rediscovered from. I have **not** traced where manual entries are seeded, so treat "unrecoverable" as unproven — but the delete being unfiltered, and the comment's classification, are both verified.

`cleanupFileIndex({ rebuild: true })` calls `appProvider.rebuildIndex()` and `fileProvider.rebuildIndex({ force: true })` afterwards, which do route correctly. That may converge for rediscoverable rows. It does not make the count right, and it does not help anything the rescan cannot find.

## It is live, not dead code

`channel/common.ts:1632-1634` wires it to `StorageEvents.cleanup.fileIndex` — a renderer-triggered Settings/Storage action.

## The guard cannot catch this, structurally

`scripts/check-search-index-writers.mjs` reports a clean bill of health today:

```
[search-index-writers] 3 worker-owned tables have no writer outside 3 approved owners
```

It cannot see this file. Its matcher requires a literal `schema.` prefix:

```js
const MUTATION = new RegExp(
  String.raw`\.\s*(?:insert|update|delete)\s*\(\s*schema\.(${OWNED_TABLES.join('|')})\s*\)`,
)
```

Tested against both spellings:

```
MATCHED db.delete(schema.files)   // prefixed — the guard can see this
missed  db.delete(files)          // bare — how storage-maintenance imports them
```

And the real code is worse for a regex than a bare identifier: the mutation is `db.delete(table)` where `table` is a **function parameter**, so no call-site pattern can resolve it at all. `OWNED_TABLES` is also only `['files', 'fileExtensions', 'keywordMappings']` — `fileIndexProgress`, `scanProgress` and `embeddings` are outside its scope entirely, and this function deletes all three.

The guard's message should be read as "no writer *matching this regex*", not as the invariant it claims.

## The test locks the behaviour in

`storage-maintenance.test.ts:21-26` mocks the database module with `getDb` and `getAuxDb` only — no `getSearchDb`, no `isSearchSplitEnabled`. The suite can only ever exercise the primary path, so it passes identically with or without the defect.

## Suggested direction

1. Route the file-index half of `cleanupFileIndex()` through the worker-owned connection when the split is on, the way `failed-files-cleanup-task.ts` and `file-provider-runtime-reset-service.ts` already do.
2. Scope the primary-side deletes so a *file* index cleanup cannot remove `type='app'` rows, split or no split. That half looks wrong even with the split off.
3. Make the guard check the connection rather than the call-site spelling, or at minimum fail closed on an unresolvable `db.delete(<variable>)` in a non-approved file. A guard whose clean result depends on how a table happens to be imported is not measuring the invariant.
4. Give `storage-maintenance.test.ts` a split-on case; it currently cannot fail on this.

## Acceptance

- [ ] With the split on, `cleanupFileIndex()` clears `search-index.db` and reports a count from it.
- [ ] No path in this function deletes `type='app'` rows.
- [ ] The writer guard fails on this file if the split check is removed again.
- [ ] A split-on test exists that fails against today's code.

Relates to #1748 — this is a concrete instance of the divergence that gate exists to catch, found statically. The runtime half of #1748 is still open.

TalexDreamSoul:
Fixes #1770. Found while working the static half of #1748, whose stated failure mode is *"a provider writes `database.db` while readers consume `search-index.db`"*.

`cleanupFileIndex()` had **no reference to the search split at all** — grepping the file for `getSearchDb`, `isSearchSplitEnabled`, `searchIndexWriter`, `execWrite` or `searchDb` returns zero. It read and wrote through `getDb()` and deleted with no `where` clause. Two things went wrong at once, pointing opposite ways.

## It cleaned the wrong file

With the split on — default since `cd39bdbf6` — file rows live in `search-index.db`. Deleting from `database.db` left the live index intact and reported a `removedCount` counted from a near-empty table.

`getSearchDb()` already returns the primary connection when the split is off or the search file is not open (`modules/database/index.ts:145`, and its own comment says callers "never need to branch on the flag"), so this is one accessor change, not a flag branch.

## It deleted the app catalog

`files` is shared. `type: 'app'` rows are the app catalog, which `app-provider.ts:840-852` keeps on the primary **deliberately**, in its own words:

> The app CATALOG (files/file_extensions rows of type 'app', **including user-authored managed entries**) lives on the PRIMARY db … manual entries are **user data that must not move into the rebuildable search-index.db**.

Those entries have nowhere else to live — `managedEntry` is not even a column — and they are not rediscoverable: `rebuildIndex()` rescans `appScanner.getWatchPaths()`, while `addAppByPath` resolves with `skipWatchCheck: true`, so an app added from outside the watch paths cannot come back.

**This half predates the split.** The app catalog moved into `files` in `1549ab65b` (2025-08-10); this function was written in `f86d4596c` (2026-01-19). It has been true for about seven months and is independent of the flag.

`files` and `file_extensions` are now scoped to non-app rows. `file_index_progress` and `scan_progress` are file-only and stay unscoped.

## Not widened

`clearSearchIndex` also clears `search_index` and `keyword_mappings`, which carry the app catalog's searchable projection as well as the file index. Same shape of over-broad delete — but it is opt-in and `rebuild` re-projects the catalog, so it is a separate decision. Recorded as a comment at the site rather than changed here.

## Why nothing caught it

`storage-maintenance.test.ts` mocks the database module with `getDb` and `getAuxDb` only, so the suite could only ever exercise the primary path and passed identically with or without the defect.

`scripts/check-search-index-writers.mjs` cannot see this file either. Its matcher requires a literal `schema.` prefix:

```
MATCHED db.delete(schema.files)   ← the guard can see this
missed  db.delete(files)          ← how this file imports them
```

and the real code is `db.delete(table)` with `table` a **function parameter**, which no call-site regex can resolve. `OWNED_TABLES` is also only `['files', 'fileExtensions', 'keywordMappings']`, while this function deletes `fileIndexProgress`, `scanProgress` and `embeddings` too. Its "3 worker-owned tables have no writer outside 3 approved owners" should be read as "no writer *matching this regex*". Improving the guard is left to #1770; this PR does not touch it.

## Verification

| check | result |
| --- | --- |
| `storage-maintenance.test.ts` | **4/4 pass** |
| negative control — new tests against the unfixed source | both fail (`expected spy to be called at least once`, `expected 2 times, got 0`); the two pre-existing cases still pass |
| `tsc --noEmit -p tsconfig.node.json` | exit 0 |
| eslint (core-app's own config) | clean |

The runtime half of #1748 — disposable-profile first-launch reindex, count parity, WAL behaviour, `=0` rollback — is untouched and still open.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file index cleanup by using the appropriate search database connection.
  * Preserved application files during cleanup.
  * Ensured related file-extension records are removed only when associated with eligible files.
  * Maintained existing cleanup behavior for other file-related data.
  * Improved cleanup consistency by applying appropriate scope to eligible file records.

### DIFF
--- a/apps/core-app/src/main/service/storage-maintenance.test.ts
+++ b/apps/core-app/src/main/service/storage-maintenance.test.ts
@@ -5,22 +5,29 @@ const {
   deleteMock,
   fileRebuildMock,
   getDbMock,
+  getSearchDbMock,
   runMock,
   selectFromMock,
-  selectMock
+  selectMock,
+  deleteWhereMock,
+  selectWhereMock
 } = vi.hoisted(() => ({
   appRebuildMock: vi.fn(),
   deleteMock: vi.fn(),
   fileRebuildMock: vi.fn(),
   getDbMock: vi.fn(),
+  getSearchDbMock: vi.fn(),
   runMock: vi.fn(),
   selectFromMock: vi.fn(),
-  selectMock: vi.fn()
+  selectMock: vi.fn(),
+  deleteWhereMock: vi.fn(),
+  selectWhereMock: vi.fn()
 }))
 
 vi.mock('../modules/database', () => ({
   databaseModule: {
     getDb: getDbMock,
+    getSearchDb: getSearchDbMock,
     getAuxDb: vi.fn()
   }
 }))
@@ -60,15 +67,28 @@ import { cleanupFileIndex } from './storage-maintenance'
 
 describe('cleanupFileIndex', () => {
   beforeEach(() => {
-    selectFromMock.mockReset().mockResolvedValue([{ count: 1 }])
+    // `.from()` is awaited directly for unscoped counts and chained with `.where()` for scoped
+    // ones, so it has to be thenable *and* carry a `where`.
+    selectWhereMock.mockReset().mockResolvedValue([{ count: 1 }])
+    selectFromMock.mockReset().mockImplementation(() => {
+      const rows = [{ count: 1 }]
+      return Object.assign(Promise.resolve(rows), { where: selectWhereMock })
+    })
     selectMock.mockReset().mockReturnValue({ from: selectFromMock })
-    deleteMock.mockReset().mockResolvedValue(undefined)
+    deleteWhereMock.mockReset().mockResolvedValue(undefined)
+    deleteMock
+      .mockReset()
+      .mockImplementation(() =>
+        Object.assign(Promise.resolve(undefined), { where: deleteWhereMock })
+      )
     runMock.mockReset().mockResolvedValue(undefined)
-    getDbMock.mockReset().mockResolvedValue({
+    const connection = {
       select: selectMock,
       delete: deleteMock,
       run: runMock
-    })
+    }
+    getDbMock.mockReset().mockResolvedValue(connection)
+    getSearchDbMock.mockReset().mockReturnValue(connection)
     appRebuildMock.mockReset().mockResolvedValue({ success: true })
     fileRebuildMock.mockReset().mockResolvedValue({ success: true })
   })
@@ -83,6 +103,34 @@ describe('cleanupFileIndex', () => {
     expect(appRebuildMock.mock.invocationCallOrder[0]).toBeLessThan(
       fileRebuildMock.mock.invocationCallOrder[0]
     )
+  })
+
+  /**
+   * #1770. File rows live in `search-index.db` under the default-on split; `getDb()` is the primary
+   * connection, which holds only the app catalog. Cleaning up through it counted and deleted from
+   * the wrong file.
+   */
+  it('cleans the search connection, not the primary', async () => {
+    await cleanupFileIndex({})
+
+    expect(getSearchDbMock).toHaveBeenCalled()
+    expect(getDbMock).not.toHaveBeenCalled()
+  })
+
+  /**
+   * The other half of #1770, which predates the split: `files` also holds the app catalog,
+   * including user-authored entries added via `addAppByPath` that exist nowhere else and that
+   * `rebuildIndex()` cannot rediscover outside the watch paths. An unscoped delete removed them.
+   */
+  it('scopes the files and file_extensions deletes so the app catalog survives', async () => {
+    await cleanupFileIndex({})
+
+    // Four tables are cleared: fileIndexProgress and scanProgress are file-only and stay
+    // unscoped, while files and file_extensions must each go through `.where(...)`. Asserting the
+    // delete counts rather than the select counts keeps this independent of how the
+    // file_extensions subquery happens to be built.
+    expect(deleteMock).toHaveBeenCalledTimes(4)
+    expect(deleteWhereMock).toHaveBeenCalledTimes(2)
   })
 
   it('returns rebuild error while still attempting file index rebuild', async () => {
--- a/apps/core-app/src/main/service/storage-maintenance.test.ts
+++ b/apps/core-app/src/main/service/storage-maintenance.test.ts
@@ -144,13 +144,14 @@ describe('cleanupFileIndex', () => {
     const [filesPredicate] = deleteWhereMock.mock.calls[1] as [SQL]
 
     expect(render(filesPredicate)).toEqual(render(ne(files.type, 'app')))
-    // The extensions delete must be scoped by the same non-app file set. Rendering the expected
-    // subquery shape verbatim would over-pin how it is built, so assert the load-bearing parts:
-    // it filters file_extensions.file_id against files rows excluding type = 'app'.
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
Reproduced the loss this fixes on a real database rather than only in unit tests.

Launched CoreApp against a fresh isolated profile, let start-up finish, and read both files of the split:

```
database.db        files=229   {'app': 229}
search-index.db    files=0     {}
```

Then applied both versions of the delete sequence to copies of that real `database.db`:

```
before          : {'app': 229}
after master    : {}            <- entire app catalog removed
after PR #1771  : {'app': 229}  <- intact
```

229 applications on a stock profile, from the action whose confirmation promises only 「将删除文件索引与搜索索引数据」.

This is the same conclusion the unit tests reach — they fail against the unfixed source with `expected 2 times, got 0` — but on real data the size of it is easier to weigh: it is not an edge case, it is every application the app knows about, on any profile, every time that button is used.

TalexDreamSoul:
Took the review point in 80253dc3d rather than resolving it away: the test now renders the predicate each delete received through `SQLiteSyncDialect().sqlToQuery` and asserts the files delete renders identically to `ne(files.type, 'app')`, and the file_extensions delete filters `file_id` against files rows excluding `type = 'app'` (`params === ['app']`). An inverted or dropped predicate now diverges in the rendered SQL and fails; the call-count assertions alone could not catch that. Render shapes were verified against the real drizzle 0.45 dialect before pushing.

TalexDreamSoul:
CI caught my previous commit and it was a real mistake in the test, not a flake: I asserted the extensions predicate would render nested subquery SQL (`"type" <> ?`), but the unit connection is a mock, so drizzle sees an opaque value and renders the whole subquery as a single `?`. Corrected in a8cc0de35 to assert what is actually observable — the rendered shape `"file_extensions"."file_id" in ?` plus the predicate the subquery itself was built with (`selectWhereMock` call 0), which is where the non-app scoping lives.

Verified locally against this branch, with three negative controls rather than a green run alone:

| mutation | result |
| --- | --- |
| none (as pushed) | 4/4 pass |
| `ne(files.type,'app')` → `eq(...)` | fails: `'"files"."type" = ?'` vs `'<> ?'` |
| `removeScoped(files, …)` → `removeAll(files)` (the original #1770 bug) | fails: where-spy called 1×, expected 2× |
| subquery scope drifted to `ne(files.type,'zzz')`, files delete left correct | fails on the params mismatch |

The third is the one the review asked for and the one my first attempt did not actually cover.

PR #1772, opened 2026-08-20
test(core-app): add the search-split topology verifier for #1748's runtime gate

TalexDreamSoul:
Refs #1748. That gate's remaining criteria need a real app run — a disposable-profile first-launch index under the default-on split, then a `TUFF_DB_SEARCH_SPLIT_ENABLED=0` rollback with post-restart parity. This is the half of that which can be automated and verified: the measuring and the judging.

It deliberately does **not** launch CoreApp. Doing that correctly needs a packaged bundle and CDP, and `coreapp-packaged-indexing-diagnostics-probe.ts` already owns that shape. Writing a second launcher I could not run myself would have meant shipping unverified automation.

```bash
# criterion 2 — after a first-launch index with the split on
tsx scripts/search-split-topology-verify.ts --profile /tmp/tuff-split-on --expect-split \
  --log ~/Library/Logs/TalexTouch/main.log --out /tmp/split-on.json

# criterion 3 — relaunch the same profile with TUFF_DB_SEARCH_SPLIT_ENABLED=0, then
tsx scripts/search-split-topology-verify.ts --profile /tmp/tuff-split-on --expect-shared \
  --baseline /tmp/split-on.json
```

Exit code is 0 only when every check passes, so it can gate a release review directly.

## What it checks, and why each one is there

| check | why |
| --- | --- |
| no non-app `files` rows on the primary | the divergence itself — a provider still writing `database.db` |
| `search-index.db` is populated | an empty profile must not pass by having nothing anywhere |
| the app catalog did **not** move | the mirror-image mistake, and a real one: routing the catalog through the split was ship-blocker #3 on 2026-08-05 |
| each worker-owned table unwritten on the primary | catches a stray writer that did not go through the worker |
| parity across the rollback | measured on the **union** of both files, since the rows are supposed to move |
| `SQLITE_BUSY` / "database is locked" | the gate asks for their absence, so absence has to be measured |

Two details that decide whether it can actually fail:

**A table that is absent is skipped, not counted as zero.** Otherwise a missing migration reads as a clean bill of health.

**`--expect-split` / `--expect-shared` is required, not defaulted.** Passing the wrong one turns a real divergence into a green run, so the tool refuses to guess and exits 2.

## Verified end to end, not only in unit tests

Seeded real SQLite profiles and ran the CLI against them:

```
healthy split profile                      -> 5 checks pass, exit 0
300 file rows + 17 keyword_mappings
  stranded on the primary                  -> 2 of 5 FAIL, exit 1, both named
no --expect flag                           -> exit 2
rollback parity                            -> before=500 after=500 files, 220/220 apps
```

The diverged run reports exactly what a real leak would look like:

```
✗ no file rows stranded on the primary: database.db holds 300 non-app file rows — a writer is still on the primary
✗ keyword_mappings is not written on the primary: database.db keyword_mappings=17, search-index.db keyword_mappings=0
```

15 unit tests cover the judging, each written as a pair — the healthy shape passes and the specific damage fails. `tsc --noEmit -p tsconfig.node.json` exit 0, eslint clean.

## Scope

#1748 is still open. This removes the part that is easy to get wrong by hand — knowing which table belongs in which file, and that the app catalog is the one exception — so the remaining work is running the app twice and pasting two outputs. The static half of that gate is recorded in a comment on the issue, along with #1770, which this verifier would have caught.

## Summary by CodeRabbit

* **New Features**
  * Added a command-line verifier for search-index database topology.
  * Supports validation of split and shared configurations, rollback parity, packaged and development layouts, and SQLite contention errors.
  * Detects unsupported non-app file extensions in the primary database.
  * Reports verification results and returns a failure status when checks do not pass.
* **Tests**
  * Expanded coverage for healthy and invalid topologies, layout precedence, missing databases, candidate paths, rollback behavior, and error handling.


TalexDreamSoul:
All four review findings taken in the follow-up commit — each was a fail-open path in a tool whose whole justification is failing closed:

- **SQLITE_LOCKED**: `judgeLog` now carries a third check on the `SQLITE_LOCKED` result-code token, which substring-matches the extended forms (`SQLITE_LOCKED_SHAREDCACHE`, …) the same way the `SQLITE_BUSY` pattern already did. Its message form ("database table is locked") was not caught by the existing text pattern; test added for both spellings.
- **Error discrimination in `readCounts`**: only a confirmed missing table (`/no such table/i`) may map to `null` now; lock, I/O, permission and corruption errors propagate. Verified against the real `@libsql/client` before pushing: a zero-byte db yields `SQLITE_ERROR: no such table: …` (→ absent), a corrupt file yields `SQLITE_NOTADB: file is not a database` (→ rejects). Tests cover the classifier plus both end-to-end paths through real database files.
- **Contradictory expectations**: `--expect-split --expect-shared` together now exits 2 with a refusal instead of silently judging as split.
- **Unreadable `--log`**: the `.catch(() => '')` is gone; a log that cannot be read propagates to main's catch, reports the error, and exits nonzero — absence of evidence no longer reads as evidence of absence.

**Declined: the OpenGrep raw-query-concat flag on the `count(*)` template.** The interpolated identifier comes from the `SEARCH_OWNED` const list in the same file; table names cannot be bound parameters in SQLite, and nothing user-controlled reaches the string. A comment now records that at the call site.

PR #1774, opened 2026-08-20
test(core-app): add the release-notes visual probe #482 has been waiting on

TalexDreamSoul:
Refs #482. Implements the recipe I posted on that issue rather than leaving it as a description.

#482's remaining half is packaged/dev-mode captures of `WhatsChangedDialog` and the Settings → Update page, at desktop and narrow width. Every primitive already exists — seven sibling probes drive the renderer over CDP and capture with `Emulation.setDeviceMetricsOverride` + `Page.captureScreenshot`. What was missing was a probe for these two surfaces.

## Three facts it encodes, each of which costs a session to rediscover

**The dialog is single-shot per profile.** `useReleaseNotesRuntime.evaluateStartup()` opens it only while the version is unacknowledged; `closeDialog()` acknowledges permanently. So it is reproducible **only against a fresh `userDataDir`** — which is exactly what the sibling probes already isolate. Pointed at a real profile it captures nothing, and a first attempt would fail with no hint as to why. So a missing dialog is reported as a *failed check naming that reason*, never as absent evidence:

```
dialog-not-shown: no .whats-changed-dialog. Expected on a fresh userDataDir; against a used
profile the version is already acknowledged and the dialog never opens
```

That distinction is the point — absent evidence reads identical to a pass in any summary, which is the failure mode #482 exists to close.

**"Update-history" does not exist.** #482 and the maintenance audit both use that name; searching all 756 renderer files for `update.?history` returns zero. The router maps category key `update` → `SettingUpdatePage.vue`, category paths are `/setting/<key>` (`DEFAULT_SETTING_PATH` is `/setting/overview`), and history is `createWebHashHistory` — so the address is `#/setting/update`. Worth fixing the wording in the issue and audit too.

**CoreBox and plugin Surfaces are also `page` targets.** Taking the first one is a race that would screenshot a plugin and then report a missing dialog. `selectMainWindowTarget` excludes them and has a test for exactly that ordering.

## One thing I got wrong and caught before pushing

An earlier draft asserted `[data-setting-page="update"], .setting-update-page`. **I invented both** — they exist nowhere in the renderer. Replaced with two signals that do exist and are judged together: `location.hash` and `.SettingsPage` (from `components/settings/SettingsPage.vue`, which every category page wraps itself in). `.SettingsPage` alone only says "some settings page mounted", so the hash is what says which one. Wrong-route and route-rendered-nothing are reported as different failures because they are different bugs.

## Verification

| check | result |
| --- | --- |
| `scripts/coreapp-visible-release-notes-probe.test.ts` | **16/16 pass** |
| `tsc --noEmit -p tsconfig.node.json` | clean |
| eslint | clean |

The test covers the deciding half only, per the split in `coreapp-packaged-ai-ask-probe.test.ts` — including that a half-loaded renderer reports *only* `renderer-not-ready` rather than judging everything else as broken, and that the DOM expression queries the class names `WhatsChangedDialog.vue` actually renders (a selector typo is invisible at runtime; the query just returns nothing and the check fails for the wrong reason).

## What this does not do

**It has not been run against a live app.** The capture sequence is read from source, so treat it as a plan until someone executes it once. Two things a first run should confirm: that the dialog appears before the router is usable (inferred from `AppEntrance.vue`'s `appEntranceMode === 'MainApp'` gate, not observed), and that `#/setting/update` renders with the dialog dismissed rather than behind it.

Launching is deliberately left to the caller, the same split `coreapp-packaged-indexing-diagnostics-probe.ts` uses — start CoreApp with `--remote-debugging-port` and an isolated `userDataDir`, then:

```bash
pnpm release-notes:visual:probe -- --cdp-url http://127.0.0.1:9222 \
  --out-dir docs/engineering/reports/release-notes-<date>
```

Also worth noting for whoever wires this up: **no CI workflow invokes any of these probes today** — 0 hits across all 23 workflow files for `visible:experience`, `packaged-*-probe`, `coreapp-visible`. This one will be hand-run like its siblings until that changes.

## Summary by CodeRabbit

- **New Features**
  - Added visual verification for the Release Notes dialog and Settings → Update page.
  - Captures desktop and narrow-screen screenshots to confirm responsive layouts.
  - Checks release notes content, dismissal controls, loading states, and update navigation.
  - Provides clear validation results and diagnostics when expected screens or evidence are unavailable.

- **Tests**
  - Added automated coverage for missing, incomplete, blank, misrouted, and still-loading page states.
  - Verifies window selection, page content extraction, navigation behavior, viewport handling, and screenshot output.

### DIFF
--- a/apps/core-app/package.json
+++ b/apps/core-app/package.json
@@ -68,6 +68,7 @@
     "clipboard:stress:verify": "pnpm exec tsx \"scripts/clipboard-stress-verify.ts\"",
     "app-index:diagnostic:verify": "pnpm exec tsx \"scripts/app-index-diagnostic-verify.ts\"",
     "everything:diagnostic:verify": "pnpm exec tsx \"scripts/everything-diagnostic-verify.ts\"",
+    "release-notes:visual:probe": "pnpm exec tsx \"scripts/coreapp-visible-release-notes-probe.ts\"",
     "settings:indexing-diagnostics:verify": "pnpm exec tsx \"scripts/settings-indexing-diagnostics-verify.ts\"",
     "search:fts-ownership-simulate": "pnpm exec tsx \"scripts/search-index-fts-ownership-simulation.ts\"",
     "search:index-migration:preflight": "pnpm exec tsx \"scripts/search-index-migration-preflight.ts\"",
--- a/apps/core-app/scripts/coreapp-visible-release-notes-probe.test.ts
+++ b/apps/core-app/scripts/coreapp-visible-release-notes-probe.test.ts
@@ -1,0 +1,167 @@
+import { describe, expect, it } from 'vitest'
+import {
+  buildReleaseNotesEvidenceChecks,
+  navigateToUpdatePageExpression,
+  readReleaseNotesDomExpression,
+  screenshotFileName,
+  selectMainWindowTarget,
+  UPDATE_ROUTE_HASH,
+  VIEWPORTS,
+  type DevToolsTarget,
+  type ReleaseNotesProbeDom
+} from './coreapp-visible-release-notes-probe'
+
+/**
+ * Covers the decision half of the probe, per the split in `coreapp-packaged-ai-ask-probe.test.ts`.
+ * The launch/attach/capture wiring is exercised only by a real run and is not asserted here — a
+ * test that mocked CDP into passing would say nothing about whether the app renders the dialog.
+ */
+
+function dom(overrides: Partial<ReleaseNotesProbeDom> = {}): ReleaseNotesProbeDom {
+  return {
+    href: 'app://index.html#/',
+    readyState: 'complete',
+    hasDialog: true,
+    dialogVersionCount: 2,
+    dialogSummaryCount: 7,
+    hasDialogActions: true,
+    hash: UPDATE_ROUTE_HASH,
+    hasSettingsPage: true,
+    bodyText: '',
+    ...overrides
+  }
+}
+
+const failing = (checks: ReturnType<typeof buildReleaseNotesEvidenceChecks>): string[] =>
+  checks.filter((check) => !check.ok).map((check) => check.tag)
+
+describe('buildReleaseNotesEvidenceChecks', () => {
+  it('passes a fresh-profile capture', () => {
+    expect(failing(buildReleaseNotesEvidenceChecks(dom()))).toEqual([])
+  })
+
+  /**
+   * The case this probe exists for. Against a used profile the version is already acknowledged, so
+   * the dialog never opens — and the run must say so loudly rather than produce no dialog evidence,
+   * which would read the same as a pass in any summary.
+   */
+  it('reports a missing dialog as a failed check, not as silence', () => {
+    const checks = buildReleaseNotesEvidenceChecks(dom({ hasDialog: false }))
+    expect(failing(checks)).toContain('dialog-not-shown')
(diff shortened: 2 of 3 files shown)


TalexDreamSoul:
I attempted a live run rather than leaving the caveat abstract. It did not complete, but it produced three facts worth having, and one of them corrects the prerequisite I wrote in the PR body.

## What worked

Launched with the recipe the sibling probes use, against an isolated profile:

```bash
TUFF_STARTUP_BENCHMARK_ONCE=1 \
TUFF_STARTUP_BENCHMARK_EXIT_DELAY_MS=90000 \
TUFF_STARTUP_BENCHMARK_USER_DATA_DIR=/tmp/tuff-rn-probe \
electron out/main/index.js --remote-debugging-port=9333
```

- **The isolation works exactly as designed.** `[TouchApp] Startup benchmark mode enabled, skip single-instance lock` — so this cannot collide with a running instance, and `precore.ts:160` repoints `userData` at the temp dir. Confirmed in the log, not assumed.
- **CDP comes up in about 5 seconds**, serving one `page` target.

## What did not, and why

The app shut down 4.3 seconds after start, before the probe could attach:

```
[15:04:17.040] [ERROR] [TouchApp] TouchApp runtime initialization failed
                       error=Error: ELECTRON_RENDERER_URL is not set
[15:04:17.632] [ERROR] [CoreBox:Window] Failed to load content in new box window
[15:04:17.635] [DEBUG] [CoreBox:Window] BoxWindow closed
[15:04:20.707] [INFO]  [TouchApp] App quit requested
```

`out/main/index.js` from `electron-vite build` is not self-sufficient — it expects `ELECTRON_RENDERER_URL` pointing at a running Vite dev server. Without it the runtime fails to initialise, the window has nothing to load, and the app cascades to shutdown.

**This corrects the PR body.** I wrote "start CoreApp with `--remote-debugging-port` and an isolated `userDataDir`", which is incomplete. The real prerequisite is one of:

- a **packaged** app — which is why every sibling probe takes an `executablePath` rather than a build directory, a detail I had read past, or
- a running dev server plus `ELECTRON_RENDERER_URL`

I have updated nothing in the branch for this yet; say which shape you want and I will fold it into the header comment, or add the launch half so the probe takes `--app-bundle` like `coreapp-packaged-indexing-diagnostics-probe.ts` does.

## What is still unverified

The two timing assumptions I flagged originally remain untested: that the dialog appears before the router is usable, and that `#/setting/update` renders with the dialog dismissed rather than behind it. Nothing in this attempt got far enough to observe either.

The unit half is unaffected — 16/16 still pass, and none of them mock CDP into passing, which is why they stayed honest about this.

Cleaned up after the attempt: temp profile removed, no stray process.

TalexDreamSoul:
Declining this one with reasons rather than taking it. `dom()` models the snapshot the judge receives — the probe's post-navigation capture — not a navigation starting state, so the golden fixture deliberately is the passing capture. The two properties the suggestion wants protected are already pinned: the judge rejects a non-update route (`wrongRoute` case, `hash: '#/setting/overview'` → `hash` tag fails), and the navigation expression targets the judged route (`navigateToUpdatePageExpression()` contains `UPDATE_ROUTE_HASH === '#/setting/update'`). Defaulting the fixture to `#/` would make every single-dimension case fail two checks at once (its own + route), weakening the isolation the `toContain` assertions rely on, without adding a property the two existing cases don't already hold.

Issue #1776, opened 2026-08-20
Three nexus check:* gates are defined but invoked by nothing — including the one a comment says blocks the #1768 relapse


TalexDreamSoul:
Found while confirming #1768 was really fixed. The fix landed and is good; this is about the guard that was added alongside it, which does not run.

## The finding

`apps/nexus/package.json` defines seven `check:*` / analyze gates. Three are invoked by nothing — no workflow, no aggregate script, no git hook:

| script | in a workflow | anywhere else |
| --- | --- | --- |
| `check:api-routes` | yes | — |
| `check:mdc-fences` | yes | — |
| `check:preview-secrets` | yes | — |
| `check:runtime-evidence` | yes | — |
| **`check:doc-parity`** | **no** | nothing at all |
| **`check:icon-collections`** | **no** | one code comment |
| **`build:analyze-worker`** | **no** | archived July task notes |

`job_nexus` in `ci.yml:375-412` runs exactly three things — `check:api-routes`, `check:mdc-fences`, `build` — so the rest are hand-run or not run.

Ruled out before filing, because "defined but unused" is easy to get wrong: not in any `.github/workflows` file, not in a `check:all`-style aggregate, not in `.husky/pre-commit` (which only touches `apps/core-app/package.json`) or `.husky/commit-msg` (commitlint only), and not in `lint-staged`. `check:api-routes` was the positive control on every scan and came back wired each time.

## Why `check:icon-collections` is the one that matters

#1768 was that eight of nine icon shortcuts aliased onto `@iconify-json/ri`, which nexus does not install, so they rendered blanks. That is fixed. But the test written with the fix says, at `check-worker-bundle.test.ts:122`:

> …and `pnpm check:icon-collections` now blocks a relapse by comparing every class against the installed @iconify-json deps.

That sentence is the only reference to the script in the repository outside its own definition. Nothing invokes it, so it blocks nothing — a relapse would land exactly the way #1768 did. The guard is written and, as far as I can tell, correct; it is simply not connected.

Still true on master today, which is what makes this reachable rather than theoretical:

```
@iconify-json/carbon         RESOLVES     <- positive control
@iconify-json/cib            RESOLVES     <- positive control
@iconify-json/ri             unresolvable
@iconify-json/simple-icons   unresolvable
@iconify/json  (monolith)    unresolvable
```

## A second, smaller thing this exposes

`build/check-worker-bundle.mjs:48` still budgets two tokens from the unresolvable collection:

```js
[...]

Issue #1764, opened 2026-08-20
ci / CI - tuffex has been red on master and every open PR since 08-18 — border-beam styles.js is 76.7 KiB against a 48 KiB budget


TalexDreamSoul:
Found while checking CI on an unrelated `packages/utils` PR (#1763). The failure is not that PR's — it is on master, and on every open PR I checked.

## The failure

```
[audit-package-size] Size audit failed:
- Component JS packages/tuffex/dist/es/border-beam/src/styles.js is 76.7 KiB; limit is 48.0 KiB
```

`ci / CI - tuffex` — measured, not inferred:

| where | result |
| --- | --- |
| master `c5e6660d9` (08-18) | failure |
| PR #1763 (`packages/utils` only) | fail |
| PR #1762, #1760, #1754 (dependabot) | fail |

`package-tuffex-ci.yml` is path-filtered on `packages/utils/**` as well as `packages/tuffex/**`, which is why PRs that never touch tuffex still carry the red.

## It regressed, and the revert proves the cause

`git show <sha>:packages/tuffex/packages/components/src/border-beam/src/styles.ts | wc -c`, against that commit's tuffex CI result:

| date | commit | styles.ts | tuffex CI |
| --- | --- | --- | --- |
| 08-17 | `2b2c2c033` | 91,076 B | failure |
| 08-17 | `c90fcfb3d` "revert(scope): restore scoped master state" | **0 B (file absent)** | **success** |
| 08-18 | `c5e6660d9` "feat: ship intelligence and component updates" | 90,929 B | failure |

The gate went green exactly when the file was absent and red exactly when it was present, in consecutive commits. That is the whole cause.

This is not stale-budget debt. The budget's own contract, stated in `scripts/audit-package-size.mjs:85-91`, is "today's size plus minimal headroom, growth from here fails" — the last re-baseline was 08-15 for the Beautiful UI port, and it was green then.

## Why it stayed red for two days

master has zero required status checks (#1716). A failing job at the merge boundary is indistinguishable from a passing one, so this landed and nothing stopped it. This is the second concrete instance of that defect after #1712/#1714.

## The size is real content, not a duplicated bundle

[...]

PR #1777, opened 2026-08-20
ci(nexus): run the icon-collection gate that was written but never invoked

TalexDreamSoul:
First of the three items in #1776, which stays open for the other two.

## What was wrong

`check:icon-collections` was written alongside the #1768 fix specifically to stop a relapse, and `check-worker-bundle.test.ts:122` states that it does:

> …and `pnpm check:icon-collections` now blocks a relapse by comparing every class against the installed @iconify-json deps.

That comment was the only reference to the script anywhere in the repository outside its own `package.json` entry. No workflow, no aggregate script, no git hook — `.husky/pre-commit` touches only `apps/core-app/package.json`, `commit-msg` is commitlint, and `lint-staged` does not mention it. So the sentence describes something that was not happening, and a relapse would have landed exactly the way #1768 did.

The condition it guards is still live on master:

```
@iconify-json/carbon         RESOLVES     <- positive control
@iconify-json/cib            RESOLVES     <- positive control
@iconify-json/ri             unresolvable
@iconify-json/simple-icons   unresolvable
@iconify/json  (monolith)    unresolvable
```

## Verification

The two steps this sits next to set a convention in their comment — verified passing on a clean tree before being wired — so I did that, and one thing more.

**It passes on a clean tree**, which is what makes it safe to make blocking today:

```
[check-icon-collections] ok — installed: carbon, cib, logos, twemoji
exit 0
```

**And it can fail**, which matters more here than usual: a gate that only ever passes is the exact shape that let #1768 ship for two days. Planting `class="i-ri-refresh-line"` in a file under `app/components/`:

```
[check-icon-collections] icon classes from collections this app does not install:
  installed: carbon, cib, logos, twemoji

  i-ri-refresh-line
    components/__negctl_icon_probe.vue

These render as nothing. Remap them onto an installed collection, or add the dependency.
exit 1
```

Worth recording how nearly I got that wrong: my first run of the negative control read the exit code through a pipe, so `$?` was `tail`'s status and reported 0 — a passing-looking result for a check that had just printed a complaint. The exit codes above are node's own.

**This PR exercises its own change.** The `nexus` path filter includes `.github/workflows/ci.yml`, so `job_nexus` runs here and the new step executes rather than being skipped.

## Not in scope

The other two items in #1776 are left alone. `build:analyze-worker` cannot simply be wired the same way — `check-worker-bundle.mjs:48` still budgets `i-ri-settings-3-line` and `i-ri-settings-3-fill`, that guard reports a token emitting no CSS as missing rather than passing it, and `ri` cannot resolve, so it would go red on entries #1768 already found had no call sites. Removing them also means moving the two assertions at `check-worker-bundle.test.ts:151-152`. `check:doc-parity` I have not run.

## Summary by CodeRabbit

- **Chores**
  - Enhanced automated quality checks to verify that required icon collections are installed.
  - Releases now receive an additional validation step alongside existing API and content-format checks, helping identify configuration issues before changes are delivered.
  - Updated verification guidance to clarify the icon collection checks and improve consistency across the release process.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -394,13 +394,24 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
-      # Both verified passing on a clean tree before being wired: the route-tree check exits 0
-      # with "server/api route tree verified", the fence check exits 0 silently.
+      # All three verified passing on a clean tree before being wired: the route-tree check exits 0
+      # with "server/api route tree verified", the fence check exits 0 silently, and the icon check
+      # exits 0 with "ok — installed: carbon, cib, logos, twemoji".
       - name: Verify server API route tree
         run: pnpm -C apps/nexus check:api-routes
 
       - name: Verify MDC fences
         run: pnpm -C apps/nexus check:mdc-fences
+
+      # #1768 shipped blank icons for two days: eight of nine shortcuts aliased onto
+      # `@iconify-json/ri`, which this app does not install, and UnoCSS emits nothing for a
+      # collection it cannot resolve. The aliases were fixed and this script was written to block a
+      # relapse — but it was never invoked by anything, so it blocked nothing (#1776).
+      #
+      # Also verified it can fail, since a gate that only ever passes is what let #1768 land: a
+      # planted `class="i-ri-refresh-line"` under app/components/ exits 1 and names the file.
+      - name: Verify icon collections are installed
+        run: pnpm -C apps/nexus check:icon-collections
 
       # Blocking. It was briefly continue-on-error because wiring it up is what discovered
       # the build was broken — nitro could not resolve @panva/hkdf from next-auth against a


TalexDreamSoul:
The `App suites (core-app)` red here is not this change — this PR adds one step to `job_nexus`, which cannot reach core-app's vitest run.

```
FAIL  scripts/search-index-migration-evidence-verify.test.ts
      > passes strict CLI verification with PNG screenshots and DOM artifacts
Error: Test timed out in 5000ms.

Tests  1 failed | 6252 passed | 4 skipped (6257)
```

That test shells out through corepack → pnpm exec → tsx before the verifier starts, and the file sets no timeout, so the 5s default covers the spawn chain. The run took 194s wall with 227s in collect — an oversubscribed runner, which is when a fixed-cost spawn crosses the line. Filed as #1778 and fixed in #1779; failed job re-run here.

Worth saying out loud given what this PR is about: I re-ran a red rather than investigated it away, which is the habit that let #1768 and #1764 ship. The difference is that the cause is identified, reproduced from the log, and has a fix in flight — not assumed to be flaky because it was inconvenient.

The step this PR adds did run and passed: `Nexus (build + checks)` is green, so the path filter reaches it rather than skipping.

Issue #1773, opened 2026-08-20
The negated-closing-keyword guard checks a snapshot, but GitHub acts on the body at merge time


TalexDreamSoul:
The guard added for #213 / #308 works — it caught me on #1772 today, with the exact message and the accepted rephrasings, which is a good check. This is about *when* it runs, not what it does.

## The gap

`ci.yml:192-198` reads the body out of the event payload:

```yaml
- name: Check for negated closing keywords in the PR body
  env:
    PR_BODY: ${{ github.event.pull_request.body }}
```

and `ci.yml:3` triggers on `pull_request` with no `types`, so the default set is `opened`, `synchronize`, `reopened` — **not `edited`**.

So the guard sees the body as it was at the last push. GitHub, by contrast, evaluates closing keywords against the body **at merge time**. Those are different documents whenever anyone edits a description, which on this repo is often.

## Measured, not inferred

From #1772 this evening, the complete run list for that branch:

```
18:49:12  pull_request  47fa603e2  failure     <- push; body said "This does not close #1748"
18:55:32  pull_request  9cd2a91bd  success     <- push; body already rewritten
```

I edited the body between those two, at roughly 18:52, to the phrasing the guard suggests. **No run was created for the edit.** Two runs exist, both tied to pushes. I only got a green check because I happened to have a commit to push; a body-only fix would have left the PR red with nothing to do about it except push something.

## The direction that actually matters

The annoyance above is the harmless half. The other half is that a body which **passed** can gain a negated closing keyword afterwards — during review, when someone adds "note: this does not fix #NNN" to answer a question — and nothing re-checks it. The PR merges green and closes an unfixed issue, which is precisely the outcome #213 and #308 recorded.

That failure mode is invisible in exactly the way the guard exists to prevent: the check is green, and it is green about a body that no longer exists.

## Options

1. Add `types: [opened, synchronize, reopened, edited]` to the `pull_request` trigger. Closes it completely. Cost: the whole `PR Quality` job re-runs on every description edit, which is not free on a repo this active.
2. Split the body check into its own tiny workflow with `types: [opened, edited, synchronize, reopened]`, leaving `ci.yml` alone. Same coverage, seconds per run instead of minutes.
3. Move it to a merge-time check (`merge_group`, or a required check that re-evaluates on `pull_request_target`), which is the only place that matches when GitHub actually reads the keywords.

(2) looks like the best ratio to me, but the trade is yours — this is CI cost, not correctness of the check itself.

[...]

Issue #308, opened 2026-07-28
test(windows): complete packaged CoreBox Everything acceptance


TalexDreamSoul:
## Problem
The Windows Everything productionization work is blocked on interactive acceptance in a packaged CoreBox app. Existing CI evidence covers SDK, CLI, unavailable recovery, native packaging, and performance; it does not exercise packaged UI search modes.

## Required evidence
On a Windows packaged app, capture evidence for:
- [ ] ordinary search;
- [ ] `@file` search;
- [ ] structured filtering;
- [ ] result, empty, and degraded states.

Attach the evidence to the Windows acceptance manifest and run the strict manifest verifier.

## Completion criteria
- [ ] Acceptance manifest is complete and strict verification passes.
- [ ] Evidence is from a packaged Windows CoreBox app, not a Darwin build or a substituted backend probe.
- [ ] The task can move past its documented manual-acceptance blocker.

TalexDreamSoul:
## The gate is built, strict, and provably rejects unfilled evidence. What is missing is only the Windows run.

I checked whether the tooling side of this issue is actually ready, because "blocked on manual acceptance" is often half-true — the manifest turns out to be missing the fields, or the verifier accepts a placeholder.

Neither is the case here.

### Three of the four required items are already named fields

`windows-acceptance-manifest-verifier.ts:196-208`:

```ts
export interface WindowsAcceptanceEverythingSearchManualCheck {
  normalSearchPassed: boolean            // ordinary search
  explicitFileSearchPassed: boolean      // @file search
  structuredFilterSearchPassed: boolean  // structured filtering
  sdkBackendEvidencePath?: string
  cliBackendEvidencePath?: string
  unavailableBackendEvidencePath?: string   // the degraded state
  ...
}
```

and each maps to its own failure message at `:584-586`:

```
'Everything normal CoreBox search was not verified'
'Everything explicit @file search was not verified'
'Everything structured file-filter search was not verified'
```

So the manifest cannot silently omit them. The fourth item — *result, empty, and degraded states* — is partly covered by `unavailableBackendEvidencePath`; "result" and "empty" have no dedicated field, which is the one schema gap worth deciding on before the run rather than after.

### The verifier rejects the ways this would otherwise be faked

`windows-acceptance-evidence-verifier.test.ts` has 16 cases, and six of them are rejections:

```
rejects checked manual evidence templates without filled evidence fields
reports manual evidence completion failures with missing labels
rejects manual evidence with unchecked checklist items
[...]

PR #1780, opened 2026-08-21
ci: re-check the PR body when it is edited, not only when it is pushed

TalexDreamSoul:
#1773, option 2.

## The gap

The guard reads the body out of the `pull_request` event payload, and `ci.yml`'s trigger declares no `types`, so it gets the default `opened, synchronize, reopened` — no `edited`. It therefore saw the description as of the last push. GitHub reads closing keywords from the body **at merge time**. Those are different documents the moment anyone edits a description.

The annoying half is that a body-only fix could not turn the check green — measured on #1772, where the two runs on that branch were both pushes and my edit in between produced no run at all.

The half that matters is the other direction: a body that **passed** can gain "note: this does not fix #N" during review, and nothing re-checks it. The PR merges green and closes an unfixed issue — which is exactly what happened to #213 and #308, an hour apart, and why this guard exists. The check is green, about a body that no longer exists.

## Why a separate workflow

Adding `edited` to `ci.yml` closes the gap too, but PR Quality is an install plus a dozen checks and would re-run in full on every description edit. This is a checkout and one node script. The check itself is untouched — same `scripts/check-closing-keywords.mjs`, same message.

`ci.yml` keeps a comment at the old location pointing at where it went, so the next person looking for it in PR Quality finds it rather than concluding it was dropped.

## Verification

The script's own controls, run directly rather than through the pnpm alias — `package.json:52-53` shows they are the same command, which is worth confirming rather than assuming:

```
self-test                       26 cases passed, exit 0
body "This does not close #1."  exit 1, names #1 and suggests the rephrasings
body "Part of #1."              exit 0
```

Repository checks, both of which count 23 workflows rather than 22, so they are seeing the new file rather than skipping it:

```
[workflow-injection] 23 workflows: no caller-controlled context reaches a run: body
[action-pins] 23 workflows: every third-party action is SHA-pinned (actions/* exempt)
actionlint .github/workflows/pr-body-guard.yml .github/workflows/ci.yml   exit 0
```

`actions/checkout` is in the pin-exempt namespace, per the policy stated at the top of `check-action-pins.mjs`.

The body still goes through the environment rather than shell interpolation, and writes to `$RUNNER_TEMP` rather than `/tmp`.

## What this PR does not yet show

#1773's first two acceptance lines are about behaviour I cannot demonstrate from the source: that editing a description turns the check red, and editing it back turns it green, both without a push. I will exercise that on this PR once the workflow exists on the branch, using an already-closed issue number so that the worst case if someone merges mid-demonstration is closing something already closed. Results will be posted here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,23 +182,11 @@ jobs:
       - name: Check maintenance-audit report claims
         run: pnpm check:audit-report-claims && pnpm check:audit-report-claims:self-test
 
-      # GitHub's closing-keyword parser is a regex over `keyword #N` and ignores the words before
-      # it, so "It does not fix #213" marks #213 for closing exactly as "Fixes #213" does. That is
-      # not hypothetical: on 2026-08-13 it closed two user-facing issues within an hour -- #213 via
-      # "does not fix", #308 via "does not close" -- and both sentences existed specifically to stop
-      # a reader mistaking instrumentation for a fix. The second happened after the first was
-      # already understood, which is why this is a check and not a note. The body is passed through
-      # the environment rather than interpolated into the shell.
-      - name: Check for negated closing keywords in the PR body
-        if: github.event_name == 'pull_request'
-        env:
-          PR_BODY: ${{ github.event.pull_request.body }}
-        run: |
-          printf '%s' "$PR_BODY" > /tmp/pr-body.md
-          node scripts/check-closing-keywords.mjs /tmp/pr-body.md
-
-      - name: Check closing-keyword guard self-test
-        run: pnpm check:closing-keywords:self-test
+      # The negated-closing-keyword guard moved to .github/workflows/pr-body-guard.yml (#1773). It
+      # has to re-run when a description is edited, and this job's trigger has no `edited` type, so
+      # here it only ever saw the body as of the last push -- while GitHub reads closing keywords
+      # from the body at merge time. Adding `edited` to this workflow would re-run every step above
+      # on each description edit; the guard is a checkout and one node script, so it went alone.
 
       # 34 distinct blocked reasons live across plugins/ and 20 of them belong to exactly one
       # plugin, so the vocabulary is open and a closed union would be the wrong contract. Only the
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
## Acceptance 1 and 2, demonstrated on this PR

Both lines are about behaviour that cannot be read off the source, so I exercised them here. Three runs of `pr-body-guard.yml` on this branch, and **only the first came from a push**:

| run | trigger | result |
| --- | --- | --- |
| `32432906632` | push (PR opened) | success |
| `32432978606` | **body edited**, no push | **failure** |
| `32433024390` | **body edited back**, no push | **success** |

The middle run is the one that could not exist before this change.

It also failed for the right reason rather than incidentally, which is the part worth checking — a red for an unrelated cause would demonstrate nothing:

```
A closing keyword is negated. GitHub will close the issue anyway:

  #1768  <-  "not close #1768"

GitHub matches `keyword #N` and ignores the words before it, so this sentence marks the issue
for closing while telling the reader the opposite.

Write `#N is still open`, `see #N`, or `Refs #N` instead.
exit 1
```

I pointed the demonstration sentence at #1768, which is already closed, so the worst case if anyone had merged while it was in the body was closing something already closed. The sentence has been removed — `grep` against the live body returns 0 occurrences.

Acceptance 3 (the self-test still runs and still fails on a crafted negated body) is covered in the description: 26 cases pass, and `This does not close #1.` exits 1.

TalexDreamSoul:
Direct evidence for this PR, produced accidentally on #1782 an hour after opening it.

I wrote **"It does not fix #213"** in that PR's body — the exact sentence from #1736 that auto-closed #213 in the first place, in a PR whose whole subject is that incident. The guard caught it, which is the guard working.

Then the part relevant here. I fixed the body, verified locally that `check-closing-keywords.mjs` passed on the new text, and **re-ran the failed job**:

```
run 32437789805, re-run with the corrected body already saved
  A closing keyword is negated. GitHub will close the issue anyway:
    #213  <-  "not fix #213"
  exit 1
```

Still red. A re-run replays the original event payload, so it re-checks the **old** body — the corrected one is unreachable to it. Only a push cleared it.

That is stronger than what I could show when filing #1773. There I said a body-only fix "would have left the PR red with nothing to do about it except push something", reasoning from the absence of a run. This is the same conclusion measured: the obvious workaround, re-running, does not work either, because the stale body travels with the payload rather than being read at run time.

Worth noting the shape it leaves in the log: two consecutive red runs on the same commit, for a sentence that no longer exists in the PR. Anyone reading that history without knowing the trigger semantics would be looking for a defect in the branch.

Issue #1775, opened 2026-08-20
coreapp-packaged-indexing-diagnostics-probe's --runMaintenanceAction always throws: the preload has no ipcRenderer.invoke


TalexDreamSoul:
Found by trying to use it. `--runMaintenanceAction scan|reconcile|reset` cannot work against the current preload, and fails with the probe's own "unavailable" error rather than anything that points at the cause.

## The mismatch

`coreapp-packaged-indexing-diagnostics-probe.ts:runMaintenanceActionExpression` builds:

```js
const invoke = window.electron?.ipcRenderer?.invoke?.bind(window.electron.ipcRenderer)
if (!invoke) throw new Error('window.electron.ipcRenderer.invoke is unavailable')
return await invoke('app:indexed-source:scan', { sourceId, reason: 'manual-rebuild' })
```

The preload exposes no such method. `apps/core-app/src/preload/index.d.ts:14-18`:

```ts
ipcRenderer: {
  send: (channel: string, ...args: unknown[]) => void
  on: (channel: string, listener: …) => () => void
  removeListener: (channel: string, listener: …) => void
}
```

Confirmed at runtime, not only from the declaration. Attaching over CDP to a running CoreApp and reading the bridge from both windows:

```
{"cls":"MacIntel core-box","electron":["ipcRenderer","process"],"ipcKeys":["send","on","removeListener"]}
{"cls":"MacIntel",         "electron":["ipcRenderer","process"],"ipcKeys":["send","on","removeListener"]}
```

`window.electron` is present; `invoke` is not, on either window.

## Why it drifted

The preload says so itself (`index.ts:166-169`):

> A narrow replacement for @electron-toolkit/preload's electronAPI. That helper bridges send/on/once/**invoke**/sendSync/removeAllListeners for *any* channel name…

So the surface was deliberately narrowed to three methods over an allowlist (`assertBridgedChannel`, `index.ts:161-164`). The probe was written against the wide toolkit surface and never updated. Nothing failed at build time because the expression is a **string** evaluated in the renderer — there is no type checking across that boundary, which is also why a `.d.ts` that never had `invoke` did not catch it.

## Consequences

- The three maintenance actions are unreachable. Anyone using the probe to drive a scan gets `window.electron.ipcRenderer.invoke is unavailable`, which reads like an environment problem rather than a surface mismatch.
- This is on the path #1748 needs: triggering a file-provider scan against a disposable profile is exactly what `--runMaintenanceAction scan` is for.
[...]

PR #1781, opened 2026-08-21
fix(core-app): reach the diagnostics channel with send/on, not the absent invoke

TalexDreamSoul:
#1775, and the scope there was too small — see below.

## What was broken

The probe called `window.electron.ipcRenderer.invoke` in three places. It does not exist. The preload bridges `send` / `on` / `removeListener` over a two-channel allowlist and says so explicitly:

> `invoke` and `sendSync` are deliberately absent rather than allowlisted — nothing calls them, and adding them back should be a decision rather than an inheritance.

Confirmed against the implementation rather than the `.d.ts`, since a declaration can be narrower than the runtime: `src/preload/index.ts:183-197` exposes exactly those three methods, and the `invoke` at line 57 is inside the preload, on the privileged side.

**Wider than the issue recorded.** I filed #1775 as "`--runMaintenanceAction` always throws". In fact `loadDiagnosticsExpression` — the main path at line 1268, behind no flag — used `invoke` too, and worse:

```ts
export function selectSettingsTarget(...) {
  return snapshots.find((entry) =>
    entry.target.type === 'page' && … && entry.snapshot.hasIpcInvoke && …
```

`hasIpcInvoke` is `Boolean(window.electron?.ipcRenderer?.invoke)`, so it is always false and this selector **matched no target at all**. The probe could not choose a window, never mind run an action. Nothing about that was flag-gated.

Its unit test passed throughout, because the fixtures asserted `hasIpcInvoke: true` — a value no window can produce. The test described a renderer that does not exist.

## The fix

`send` + `on` are enough, because they are what the renderer's own client uses. `channelRequestPrelude()` mirrors `renderer/src/modules/channel/channel-core.ts:205-230`: same envelope, same reply match on `header.status === 'reply'` and `sync.id`, listener removed via the remover `on` returns.

That is reachable because `transport.on(AppEvents.indexedSource.*)` ends at `regChannel(BRIDGE_CHANNEL.MAIN, eventName, …)` (`transport/sdk/main-transport.ts:738`) under the plain event name — so these events do not require the MessagePort the transport prefers. `RAW_MAIN_PROCESS_CHANNEL` is in `BRIDGED_IPC_CHANNELS`, which both `send` and `on` check, so the allowlist permits it.

The capability probe now reports `hasChannelBridge` — what this actually needs — instead of a capability that could never be true.

I did not re-add `invoke` to the preload, which was my first instinct on the issue. It would reopen a surface the preload closed on purpose, to serve a diagnostics script. The cost of the direction taken is real and worth naming: this is a second implementation of the channel protocol and will drift if that protocol changes. There is no shared helper to import from a `scripts/` file today.

## Tests

Two cases the old suite could not have failed on:

- a snapshot shaped like the real bridge (`send` and `on` present, nothing else) must still select a target — this is the one that was silently false before;
- a target with no bridge must be rejected, so the field is not merely decorative.

## Verification, and its limits

This worktree cannot run the suite: `vitest.config.ts` needs `@vitejs/plugin-vue` and `tsc -p tsconfig.node.json` reports `Cannot find module` for ~40 workspace imports — missing dependencies, not this change. I did not copy the files into the main checkout to run them there; that tree has concurrent edits.

So locally this is parse-checked only, with the control established rather than assumed — my first two attempts at that control were themselves broken (an extra `(` is still valid JS; a failed `sed` produced an empty file that parsed fine). The one that works removes the final `}` and is rejected with `Unexpected end of file`, while both real files pass. CI is the gate for the tests.

**Not verified: that a real run now succeeds.** That needs a packaged app plus a CDP session, and it is the claim that matters most. The three failure modes above are all provable from the source; that the rebuilt request actually returns a diagnostics snapshot is not, and I am not asserting it.

## Left alone deliberately

Three sibling probes carry the same dependency and are untouched — `coreapp-visible-app-index-workbench-probe.ts` (7 references, including the same `hasIpcInvoke` selector at lines 1027/1037/1066), `coreapp-visible-assistant-floating-ball-probe.ts`, and `coreapp-visible-assistant-image-translate-probe.ts`. Four of the nine CDP probes in that directory, so this is systemic rather than one script's slip. Bundling them would make one unreviewable diff across four tools I cannot run; they are recorded on the issue with locations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostics communication with renderer targets.
  * Added reliable request tracking, response filtering, timeout handling, and listener cleanup.
  * Updated Settings target detection to recognize supported channel bridges.
  * Prevented unsupported targets without a channel bridge from being selected.
  * Improved reliability when communicating with supported targets during diagnostics and maintenance operations.

* **Tests**
  * Added coverage confirming valid bridged targets are selected and unsupported targets are rejected.

### DIFF
--- a/apps/core-app/scripts/coreapp-packaged-indexing-diagnostics-probe.test.ts
+++ b/apps/core-app/scripts/coreapp-packaged-indexing-diagnostics-probe.test.ts
@@ -166,7 +166,7 @@ describe('packaged indexing diagnostics probe helpers', () => {
         target: makeTarget('overlay', 'app://tuff/renderer/index.html#/meta-overlay', 'CoreBox'),
         snapshot: {
           hasRouter: true,
-          hasIpcInvoke: true,
+          hasChannelBridge: true,
           hasSettingsShell: false,
           text: 'CoreBox'
         }
@@ -175,7 +175,7 @@ describe('packaged indexing diagnostics probe helpers', () => {
         target: makeTarget('settings', 'app://tuff/renderer/index.html#/setting', 'Settings'),
         snapshot: {
           hasRouter: true,
-          hasIpcInvoke: true,
+          hasChannelBridge: true,
           hasSettingsShell: true,
           text: 'App Settings'
         }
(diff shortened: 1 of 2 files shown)


TalexDreamSoul:
**Callable-members thread — taken.** `inspectTargetExpression` now reports `hasChannelBridge` via `typeof … === 'function'` on both `send` and `on`, the same gate `channelRequestPrelude` applies, so a target whose bridge members are truthy non-functions is never selected and then refused later. The test mirrors the new expression and adds the rejection case (`{ send: true, on: 'listener' }` → no target selected).

**Typed-transport thread — declined, with reasons.** This probe is black-box packaged-app evidence tooling: it drives the packaged renderer over CDP from outside, where the only reachable surface is what the production preload actually exposes. `AppEvents.indexedSource`'s typed SDK lives inside the renderer bundle — it is not a window global, so a `Runtime.evaluate` string cannot import it; routing through it would mean adding a probe-only preload API to the production app, which changes the surface under test and defeats the point of the probe class. Rebuilding the `RAW_MAIN_PROCESS_CHANNEL` envelope is deliberate: the probe validates the wire contract the renderer really speaks, which is exactly how the absent-`invoke` drift (#1775) was caught. If a sanctioned diagnostics API for probes is wanted, that belongs in the #1789 acceptance-machinery discussion, not this fix.

PR #1782, opened 2026-08-21
ci(linux): record whether the runner had the restriction the smoke test is for

TalexDreamSoul:
Implements the one-line gap named in #213 — and a little more, because one printed line turned out not to be enough to make the verdict safe to read.

## The gap

The Linux smoke step (added by #1736) launches the AppImage under `xvfb` on `ubuntu-24.04`, and three release runs came back `ok`. By the decision rule set in that thread, `ok` is evidence against the userns/AppArmor hypothesis.

It is only evidence **if the runner had the restriction switched on**. `kernel.apparmor_restrict_unprivileged_userns=1` is 24.04's desktop default and is what an AppImage cannot satisfy, but GitHub's hosted images are not guaranteed to match it — `actions/runner-images#10015` proposes disabling AppArmor on them. On a runner without it, the smoke passes because the condition under test is absent. That is a check that cannot fail, and it had already been read as a result.

## What this does

Reads the sysctl before launching and hands it to the classifier, which prints it **next to the verdict**:

```
verdict: ok
The packaged app started and was still running when the window opened.
userns restriction: ON (kernel.apparmor_restrict_unprivileged_userns=1) -- this run exercised the #213 condition.
```

or, when it is not:

```
userns restriction: OFF (kernel.apparmor_restrict_unprivileged_userns=0) -- this runner does not
have the restriction, so a pass says nothing about #213.
```

Next to the verdict rather than as a bare `cat` earlier in the log, because the failure mode is someone reading `ok` three runs later and concluding the hypothesis is dead. Four states are distinguished: `1`, `0`, absent sysctl, and anything else — the last says `UNRECOGNISED` rather than quietly folding into "absent", since a value nobody predicted is not the same as a restriction that is off.

**Nothing is gated on it.** This reports; it does not fail the build. Whether an `ok` on an unrestricted runner should be a failure is a policy call for whoever owns #213, and making it one now would turn every release build red on a question nobody has answered yet.

macOS and Windows pass no env var and get no line.

## Verification

Self-test **40 → 50 cases**, baseline re-measured from `origin/master` rather than remembered.

The part worth reporting is that my first version of these tests was weak, and a negative control caught it. Mutating the `value === '0'` branch to something unreachable left **all 47 cases green** — because `'0'` then fell through to the unrecognised-value fallback, which is also `present: false`, which is all I was asserting. The tests now assert the message too, and each branch mutated separately fails:

```
mutate the '1' branch          -> 2 of 50 cases failed
mutate the '0' branch          -> 1 of 50 cases failed
mutate the ''/'not present'    -> 2 of 50 cases failed
```

Workflow side:

- `actionlint` — no finding anywhere in the changed range; the 4 pre-existing `SC2129` style warnings are 4 on `origin/master` too, i.e. unchanged.
- `check-workflow-injection` and `check-action-pins` both pass.
- The whole Linux `run:` block parses under `bash -n`, with a deliberately broken copy rejected first so the check is known to discriminate.

## What this does not do

This is not a fix — #213 is still open — and it does not tell us today whether the runner has the restriction — that answer arrives with the next release build. If it comes back `ON`, the three existing `ok` runs become real evidence and the hypothesis is genuinely weakened. If `OFF`, those runs said nothing and the smoke needs a runner that reproduces the condition before it can speak to this at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux application launch diagnostics by reporting AppArmor user-namespace restrictions when available.
  * Missing system information is now clearly identified without causing the launch check to fail.
  * Unrecognized or incomplete restriction values are classified and explained more clearly.

* **Tests**
  * Added coverage for recognized, missing, trimmed, and unrecognized system values.

### DIFF
--- a/.github/workflows/build-and-release.yml
+++ b/.github/workflows/build-and-release.yml
@@ -661,6 +661,17 @@ jobs:
           sudo apt-get install -y -qq xvfb libgbm1 libnss3 libasound2t64 libgtk-3-0t64 || sudo apt-get install -y -qq xvfb libgbm1 libnss3 libasound2 libgtk-3-0
           command -v xvfb-run >/dev/null || { echo "xvfb-run missing after install"; exit 1; }
 
+          # Whether this runner has the thing #213 is about. 24.04 ships
+          # kernel.apparmor_restrict_unprivileged_userns=1, which is what an AppImage cannot satisfy
+          # -- but GitHub's hosted images are not guaranteed to match a desktop install
+          # (actions/runner-images#10015 proposes disabling AppArmor on them). Without this the
+          # three `ok` runs on #213 cannot be told apart from "the restriction was never there",
+          # which is a pass produced by the absence of the condition under test.
+          # `|| true` because -e is on and a missing sysctl is a real answer, not a step failure.
+          USERNS_RESTRICT=$(cat /proc/sys/kernel/apparmor_restrict_unprivileged_userns 2>/dev/null || echo "not present")
+          export TUFF_LAUNCH_USERNS_RESTRICT="$USERNS_RESTRICT"
+          echo "kernel.apparmor_restrict_unprivileged_userns=$USERNS_RESTRICT"
+
           APPIMAGE=$(find apps/core-app/dist -name "*.AppImage" -type f | head -1)
           if [ -z "$APPIMAGE" ]; then
             echo "No .AppImage to launch. The verification step above should already have failed."
--- a/scripts/classify-app-launch.mjs
+++ b/scripts/classify-app-launch.mjs
@@ -263,6 +263,11 @@ function selfTest() {
       actual: describeUsernsRestriction('1').present,
       expected: true,
     },
+    {
+      name: 'restriction on is reported as ON, not as an unrecognised value',
+      actual: describeUsernsRestriction('1').text.includes('ON ('),
+      expected: true,
+    },
     /**
      * These assert the message, not just `present`. Asserting `present === false` alone passes
      * against a broken branch, because the unrecognised-value fallback is also `false` -- so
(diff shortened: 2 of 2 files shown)


TalexDreamSoul:
Two things from getting this green, both worth recording because they are the same shape as what the PR is about.

**1. The guard caught me writing the sentence this issue exists because of.** The first version of this description said *"It does not fix #213"* — the exact wording from #1736 that GitHub parsed as `fix #213` and used to close the issue. In a PR about that incident. Fixed to `#213 is still open`; the re-run evidence went to #1780, since it demonstrates the gap that PR closes.

**2. My first version of the tests was the third instance of a pattern this file already documents twice.** Its header says:

> That is the second time in this file a case has asserted a path that does not exist — the first was a gerund closing keyword — and both times the test was green on a branch that could never fire.

Mine asserted `describeUsernsRestriction('0').present === false`, which passes against a deleted `'0'` branch because the unrecognised-value fallback returns `false` too. Caught by mutating the branch and watching all 47 cases stay green — not by reading it back.

The tests now assert the message as well, and each branch mutated separately fails: `'1'` → 2 failures, `'0'` → 1, the absent-value pair → 2.

Also green now on lint. `scripts/` is under the **root** eslint config, which requires trailing commas — the opposite of `apps/core-app`'s — and my object literals had none. Verified locally after the fix with a negative control: stripping one comma back out reproduces CI's `177:51 error Missing trailing comma style/comma-dangle` exactly, and the real file is clean.

TalexDreamSoul:
Taken in the follow-up commit: added the mirror of the OFF-wording case — `describeUsernsRestriction('1').text.includes('ON (')`. Negative control run before pushing: mutating the ON branch text makes exactly this case fail (1 of 51), unmutated self-test passes 51/51.

PR #1795, opened 2026-08-28
docs(nexus): write the missing English API chapters and wire the parity gate

TalexDreamSoul:
Closes the second of the three items in #1776 — the `check:doc-parity` gate that was defined but invoked by nothing.

## Why it was unwired

Not an oversight. The script's own header said so:

> Those need someone who knows the subsystem, so this is NOT wired into ci.yml yet — it would fail the build on a gap it cannot fix. **Add the CI step once the list is empty.**

Measured at the start of this change:

```
EXIT=1
content/docs/dev/api/division-box:  zh 21 headings, en 12   (9 missing)
content/docs/dev/api/flow-transfer: zh 14 headings, en  9   (5 missing)
content/docs/dev/api/intelligence:  zh 20 headings, en 19   (1 missing)
```

The list is empty now, so the step is added.

## These were written against source, not translated — and two Chinese claims did not survive that

This is the part worth reviewing carefully. Checking each statement against the implementation before writing the English turned up two fabrications in the existing Chinese copies:

**1. `division-box.zh` invented an error code.** The prepare-phase section listed `DivisionBoxError(INVALID_URL)` and `DivisionBoxError(RESOURCE_LIMIT_EXCEEDED)`.

- `INVALID_URL` is `DownloadErrorType.INVALID_URL` — a different module.
- `RESOURCE_LIMIT_EXCEEDED` appears **nowhere in the codebase** except that sentence.

`DivisionBoxErrorCode` has exactly eight members, and the session-cap path throws `LIMIT_EXCEEDED` (`division-box/manager.ts:208`). Corrected in both copies.

**2. `flow-transfer.zh`'s IPC table — the one headed "实现对齐" (implementation-aligned) — listed twelve channel names, none of which exist.** Event names are composed as `namespace:module:action` (`transport/event/builder.ts`), so every real Flow channel carries a module segment.

Positive control, so this is a real absence and not a broken search:

```
'flow:dispatch'      0 matches under apps/core-app/src
flow:bus:dispatch    apps/core-app/src/main/modules/flow-bus/ipc.ts:159
```

Corrected in both copies, plus the five channels the table omitted entirely (`flow:session:deliver`, `flow:consent:check`, `flow:consent:grant`, `flow:ui:trigger-transfer`, `flow:ui:trigger-detach`).

## What the English chapters contain

| Page | Added |
| --- | --- |
| `division-box.en` | **Resource Limits** — the three real `RESOURCE_LIMITS` constants with the error each breach raises. **Lifecycle in Detail** — seven sections covering `prepare → attach → active → inactive → detach → destroy` plus the transition diagram |
| `flow-transfer.en` | **Flow Session State**, **Error Handling** (both enums verbatim from `types/flow.ts`), **Native System Sharing** (promoted from a bold label inside SDK Usage), **Target Ordering Rules**, **IPC Channels** |
| `intelligence.en` | **Complete Examples** — the translation and OCR plugin samples, with payload and result shapes checked against `IntelligenceTranslatePayload`, `IntelligenceVisionOcrResult`, `IntelligenceInvokeResult` and `IntelligenceCapabilityStatus` |

`## Technical Notes` moved in `division-box.en` to match the Chinese ordering — the check compares the sequence of heading depths, so position matters, not just count.

## Also: six corrupted lines

`## void intelligence`, `## void translateAndPaste` and four siblings sat inside `code: |` blocks in both intelligence copies. The idiom everywhere else in those files is a bare `void X`; the `## ` prefix rendered as literal text in the code sample. They were never counted as headings — block-scalar content is indented, so the line-anchored pattern never matched — so this is a rendering fix, not a parity one.

## Verification

Controls, not just a green run:

| check | result |
| --- | --- |
| `check:doc-parity` | **exit 0** across all pairs |
| negative control: demote one `###` to bold | fails — "zh has 21 headings, en has 20" |
| negative control: change a `##` to `###` | fails — "heading 10 is h2 in zh but h3 in en" |
| `check:mdc-fences` | 0 |
| `check:api-routes` | 0 |
| `check:icon-collections` | 0 |
| `actionlint -shellcheck` on the edited `ci.yml` | 0 |

Wired as **blocking**. An advisory parity check is the same defect in a quieter form — which is exactly what #1776 was filed about.

## Not covered here

#1776's third item, `build:analyze-worker`, is untouched. It needs the gzip budget question settled and the `check-worker-bundle.mjs` / `.test.ts` contradiction over `i-carbon-fingerprint-recognition` resolved before it can be wired at all. So this PR closes item 2, and #1776 stays open on item 3.

## Summary by CodeRabbit

* **Documentation**
  * Expanded API documentation for Division Box lifecycle states and resource limits.
  * Added Flow Transfer session states, error handling, target ordering, and IPC channel references.
  * Added complete Intelligence SDK plugin examples and corrected code formatting.
  * Updated Chinese documentation to match current error codes, channel names, and examples.
* **Chores**
  * Added an automated CI check to verify Chinese and English documentation heading parity.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -401,6 +401,19 @@ jobs:
       - name: Verify icon collections are installed
         run: pnpm -C apps/nexus check:icon-collections
 
+      # The third gate #1776 found defined-but-uninvoked. It stayed unwired on purpose: it would
+      # have failed the build on a gap it could not fix, since the English copies of division-box,
+      # flow-transfer and intelligence were short 15 headings between them. Those chapters are
+      # written now and the check exits 0 across all 162 pairs, so the reason to leave it out is
+      # gone. Blocking, not continue-on-error — an advisory parity check is the same defect in a
+      # quieter form.
+      #
+      # Verified it can fail before wiring, twice: demoting one `###` to bold reports
+      # "zh has 21 headings, en has 20", and changing a `##` to `###` reports
+      # "heading 10 is h2 in zh but h3 in en".
+      - name: Verify zh/en doc parity
+        run: pnpm -C apps/nexus check:doc-parity
+
       # Blocking. It was briefly continue-on-error because wiring it up is what discovered
       # the build was broken — nitro could not resolve @panva/hkdf from next-auth against a
       # root node_modules path that only existed under shamefully-hoist (#1151, #1099
--- a/apps/nexus/build/check-doc-translation-parity.mjs
+++ b/apps/nexus/build/check-doc-translation-parity.mjs
@@ -11,12 +11,15 @@
 // The check compares the *shape* — the sequence of heading depths — not the
 // text, because the headings themselves are translated.
 //
-// All 159 component doc pairs are clean. The remaining offenders are under
-// dev/api, where the English copies are missing whole chapters: division-box is
-// short an entire lifecycle chapter and flow-transfer is short five sections.
-// Those need someone who knows the subsystem, so this is NOT wired into ci.yml
-// yet — it would fail the build on a gap it cannot fix. Add the CI step once the
-// list is empty.
+// Every pair is clean and this now runs in ci.yml's nexus job. It was unwired
+// until 2026-08-27 because three dev/api pages were short 15 English headings
+// between them — division-box an entire lifecycle chapter, flow-transfer five
+// sections, intelligence one — and a gate that fails on a gap it cannot fix is
+// worse than no gate. Those chapters were written against the implementation
+// rather than translated: two claims in the Chinese copies did not survive that
+// check and were corrected rather than carried over (a `RESOURCE_LIMIT_EXCEEDED`
+// error code that exists nowhere in the codebase, and a table of `flow:*` channel
+// names missing the module segment every real event name carries).
 //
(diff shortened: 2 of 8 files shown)


TalexDreamSoul:
All five threads were correct and are fixed in #1796 — this PR had already merged when the review landed, so they go in a follow-up rather than here.

Three were substantive errors in what I wrote, verified against `division-box/manager.ts:204-262`:

- The `MAX_ACTIVE_SESSIONS` check runs before `generateSessionId()`, so `prepare --> destroy: limit exceeded` describes an edge that cannot occur.
- `keepAlive` installs a state-change handler; `lruCache.add()` fires on `INACTIVE`, not during prepare.
- Both missing-owner branches throw `CONFIG_ERROR`. `PERMISSION_DENIED` comes from `ipc.ts:229,241` at a different boundary.

The third is worth recording: I inferred that code from the phrase "missing or unauthorized plugin owner" rather than reading the `throw`, which is precisely the shortcut this PR corrected elsewhere. Same defect class, one file over.

The other two — the missing unknown-code fallback rule in `flow-transfer.zh`, and the OCR examples skipping the discovery gate the same page mandates — are also fixed there.

PR #1796, opened 2026-08-28
docs(nexus): correct four lifecycle claims I got wrong in #1795

TalexDreamSoul:
Post-merge review on #1795 caught four statements in the new DivisionBox lifecycle chapter that do not match the implementation. Three were mine. All four are fixed in both languages.

## What was wrong

Read from `apps/core-app/src/main/modules/division-box/manager.ts:204-262`:

| Claim as written | What `createSession` actually does |
| --- | --- |
| `prepare --> destroy: load failure / limit exceeded` | The `MAX_ACTIVE_SESSIONS` check runs **first** — before `generateSessionId()` and before `new DivisionBoxSession(...)`. A rejected open never produces a `prepare` session, so that edge cannot occur. Removed the limit-exceeded half. |
| "With `keepAlive`, the session is registered with the LRU cache" (in **prepare**) | `keepAlive` installs a *state-change handler*. `lruCache.add(session)` fires when the session first reaches `INACTIVE`; `updateAccess` on return to `ACTIVE`; `remove` on `DESTROY`. Nothing enters the cache during prepare. |
| "Missing or unauthorized plugin owner → `PERMISSION_DENIED`" | Both missing-owner branches throw **`CONFIG_ERROR`** — no `pluginId`, and a `pluginId` resolving to no loaded plugin. `PERMISSION_DENIED` lives at a different boundary entirely (`ipc.ts:229,241`): a non-authoritative caller requesting a UI view, or a `pluginId` that does not match the caller. |

The third is the one worth naming. I inferred the error code from the phrase "missing or unauthorized plugin owner" instead of reading the `throw` statements — the exact shortcut the rest of #1795 was written to avoid, and the same class of defect that PR corrected in the Chinese copies (a `RESOURCE_LIMIT_EXCEEDED` that existed nowhere, and a table of `flow:*` channels missing their module segment).

Both prepare sections now also state explicitly that all three failures reject the `open()` call rather than producing a session that later transitions, since that distinction is what made the state diagram wrong.

## Two consistency gaps, also fixed

- **`flow-transfer.zh`** did not carry the "treat an unrecognised `FlowErrorCode` as `INTERNAL_ERROR`" rule that the English copy states. A reader of the Chinese page would not know the enum is open. Added.
- **Both OCR examples** called `intelligence.vision.ocr` directly, while the same page (line 84) requires discovery before invocation and the translation example beside them does exactly that. Both now gate on `getCapabilityStatus({ capabilityId: 'vision.ocr' })`.

## Verification

| gate | exit |
| --- | ---: |
| `check:doc-parity` | 0 |
| `check:mdc-fences` | 0 |
| `check:api-routes` | 0 |
| `check:icon-collections` | 0 |

Parity is preserved because every change here is body text — no headings added, removed, or re-nested.

## Summary by CodeRabbit

* **Documentation**
  * Clarified DivisionBox session limits, configuration validation, permission checks, lifecycle behavior, and failure outcomes.
  * Documented standardized flow error handling, including fallback treatment for unknown error codes.
  * Updated the OCR plugin example to check capability availability before making recognition requests.

### DIFF
--- a/apps/nexus/content/docs/dev/api/division-box.en.mdc
+++ b/apps/nexus/content/docs/dev/api/division-box.en.mdc
@@ -240,15 +240,18 @@ The initial state of a session, between the `open()` call and window creation.
 - The main process validates the config, assigns a session id, and constructs the session.
 
 **Resources**
-- A session id is allocated.
+- The global session cap is checked *before* anything is allocated.
+- A session id is generated and the session registered in the session map.
 - The config is validated and defaults applied; the plugin owner is resolved before any window is constructed.
-- Session state storage is initialized.
-- With `keepAlive`, the session is registered with the LRU cache.
+- With `keepAlive`, a state-change handler is installed. It does not put the session in the LRU cache yet — the cache entry is added when the session first reaches `inactive`, refreshed on return to `active`, and removed on `destroy`.
 
 **Failures**
-- Global session cap reached → `DivisionBoxError(LIMIT_EXCEEDED)`.
-- Missing or unauthorized plugin owner → `DivisionBoxError(PERMISSION_DENIED)`.
-- Malformed configuration → `DivisionBoxError(CONFIG_ERROR)`.
+
+All three reject the `open()` call; none of them produce a session that then transitions.
+
+- Global session cap reached → `DivisionBoxError(LIMIT_EXCEEDED)`, thrown before the session id is generated.
+- A UI view requested without an owning `pluginId`, or with a `pluginId` that resolves to no loaded plugin → `DivisionBoxError(CONFIG_ERROR)`.
+- A non-authoritative caller requesting a UI view, or a `pluginId` that does not match the calling plugin → `DivisionBoxError(PERMISSION_DENIED)`, raised at the IPC boundary before the manager is reached.
 
 ### attach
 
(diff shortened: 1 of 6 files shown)


TalexDreamSoul:
All three taken in 8bd245bbe, and the third turned out to be older and larger than my edit.

**`ctx.reportError()` — the reviewer is right, and the fabrication predates this PR.** `FlowTransferHandler` (`packages/utils/plugin/sdk/flow.ts:27-31`) is:

```ts
(payload: FlowPayload, sessionId: string, senderInfo: { senderId, senderName? }) => Promise<void> | void
```

There is no `ctx`. But the English **"Receive flow (target)" example, which predates this work**, calls `ctx.acknowledge({...})` and `ctx.reportError({...})` — neither of which exists. The real methods are `flow.acknowledge(sessionId, ackPayload?)` and `flow.reportError(sessionId, message)`, and the **Chinese copy already had this right**, so the English example was the outlier. Rewritten to the real signature; my `ctx.reportError()` sentence, which propagated it one step further, corrected in both languages.

That is the third phantom API found in this doc set, after `RESOURCE_LIMIT_EXCEEDED` and the `flow:*` channel table.

**The two consistency points were both mine**, introduced by the previous commit:

- prepare's "Entered when" still said every `open()` assigns a session id and constructs a session, contradicting the Failures block immediately below it. It now states the preflight rejection first.
- detach still said a `keepAlive` session "enters the LRU cache" there, after prepare was corrected to say the entry is added on first `inactive`. It now says it is already cached and is restored from there.

Gates re-run: `check:doc-parity` 0, `check:mdc-fences` 0.

Issue #1794, opened 2026-08-27
Protocol is not a required check, so a PR that breaks the Rust build reads as mergeable — and master stayed red for nine days


TalexDreamSoul:
## Two failures from the same cause, both observed today

Master's seven required status checks are:

```
Check Not Allowed File Changes · Documentation Quality · PR Quality
Typecheck (workspace) · App suites (core-app) · App suites (nexus)
Integration suite (packages/test)
```

`Native Protocol` — the workflow that lints, tests and builds the entire Rust workspace on three runners — is not one of them. Two concrete consequences turned up within an hour of each other:

**1. Master was red for nine days and nothing said so.**

```
08-27T12:54  failure  @da95d7d99   <- master
08-18T10:09  success  @c5e6660d9   <- master, last green
```

Rust 1.98 added `clippy::chunks_exact_to_as_chunks`, `native-protocol.yml` installs the toolchain unpinned from `dtolnay/rust-toolchain`'s stable branch and lints with `-D warnings`, so master went red on the toolchain's release schedule with no commit of ours. Fixed in #1793; master's Protocol is green again as of run `33078139171`. Nothing surfaced it in the nine days between, because a red non-required run blocks no merge and notifies nobody.

A path-filtered workflow is also invisible to a plain `gh run list --branch master` — it only appears under `--workflow=native-protocol.yml` — so the redness is easy to miss even when looking.

**2. A PR that genuinely breaks the build currently reads as mergeable.**

#1756 (`napi` 3.8.6 → 3.12.2) and #1758 (`napi-derive` 3.5.5 → 3.6.3) each have **7/7 required checks green**, while `Protocol (ubuntu-latest)` fails at the test step:

```
rust-lld: error: undefined symbol: napi_call_threadsafe_function
collect2: error: ld returned 1 exit status
```

Master's ubuntu job passes, so this is introduced by the bump. Nothing would have stopped either from merging. I converted both to draft as a stopgap; that is a manual guard where a gate belongs.

## Decisions this needs

[...]

PR #1799, opened 2026-08-28
feat(nexus): fail the build when a prerendered docs page has no body

TalexDreamSoul:
Closes the gap found while triaging #1754 and recorded on #1794.

## The gap

The nexus build prerenders docs routes, and `docs-prerender-routes.test.ts` verifies the route **list** is complete. Nothing read the emitted HTML. Searching `apps/nexus/build/` for `innerText`, `textContent`, `<article>` or any body-length assertion returned nothing, and `check:runtime-evidence` is unwired on purpose (`ci.yml:361` — it "exits 1, wants an output/playwright directory from a prior run").

So a page that prerenders to an empty shell produces a green build, a green `check:mdc-fences`, a green route-list test, and a site with no prose on it. **This site has already shipped exactly that, sitewide**, when the body mode was baked into the payload key and the server render and hydration disagreed.

## The check

`check:prerender-bodies` reads the emitted HTML for the five routes `docsPrerenderEvidenceRoutes` already curated — a list that existed for this purpose and had no consumer — across both locales, and fails when visible text falls below a floor.

**A text floor rather than an element check, deliberately.** The shell renders regardless, so "an `<article>` exists" is precisely the assertion the empty-shell bug satisfied. Script and comment content is stripped first, because payload scripts are large enough to carry a body-less page over any raw-byte threshold — the same false green in a different costume.

**`MIN_BODY_CHARS = 250`, measured not picked.** Against a real build:

```
/zh/docs                                    423   <- shortest, a landing page
/en/docs                                   1002
/zh/docs/guide/start                       1250
/zh/docs/dev                               1398
/en/docs/guide/start                       2097
/zh/docs/dev/getting-started/quickstart    2657
/en/docs/dev                               2794
/en/docs/dev/getting-started/quickstart    3088
/zh/docs/dev/components                    7960
/en/docs/dev/components                   10090
```

A page whose `<main>` is emptied renders **91**. The floor sits in that gap with room for a landing page to lose a paragraph. The full table is in the source comment so the next person can re-measure rather than guess.

## One structural change

`docsPrerenderEvidenceRoutes` moves from `nexus-prerender-routes.ts` to `nexus-static-routes.mjs`, where `publicPrerenderRoutes` and `docsApiPrerenderRoutes` already live, and is re-exported so nothing else changes.

**Running the check for real is what forced that.** The self-test passed happily while the actual run died on `ERR_MODULE_NOT_FOUND` — the TS module's siblings use extensionless imports that plain node cannot resolve without a loader. Duplicating the route list in the checker would have worked and would have let the two drift, which is the failure this list exists to prevent.

## Verification

| check | result |
| --- | --- |
| self-test | 12 cases pass |
| mutation: drop the `<script>` strip | fails the script-weight case |
| mutation: treat a missing file as a pass | fails two cases |
| against a real build | 10/10, counts above |
| **end-to-end negative control**: empty `<main>` on one real page | drops to 91 chars, **fails the run and names the route** |
| `docs-prerender-routes.test.ts` after the constant moved | 8/8 |
| `actionlint -shellcheck` on the edited `ci.yml` | 0 |
| the four existing nexus gates | all 0 |

The self-test is wired **before** the build so a broken judge is reported in seconds rather than after a full nexus build. A gate whose own logic is never exercised is the thing #1794 is about.

## What this does not cover

The five evidence routes, not all ~160 docs pages — that is what the curated list scopes, and widening it is a separate decision about build-time cost. It also says nothing about *correctness* of the rendered prose, only that prose exists.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to ensure prerendered documentation pages contain visible content instead of empty page shells.
  * Checks documentation routes across supported locales after each build and blocks releases when content is missing.

* **Tests**
  * Added a self-test for the prerendered page content validation.
  * Expanded build verification across multiple documentation pages and locales.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -414,6 +414,11 @@ jobs:
       - name: Verify zh/en doc parity
         run: pnpm -C apps/nexus check:doc-parity
 
+      # The self-test runs before the build so a broken judge is reported in seconds rather than
+      # after a full nexus build. A gate whose own logic is never exercised is what #1794 is about.
+      - name: Self-test the prerender body judge
+        run: pnpm -C apps/nexus check:prerender-bodies:self-test
+
       # Blocking. It was briefly continue-on-error because wiring it up is what discovered
       # the build was broken — nitro could not resolve @panva/hkdf from next-auth against a
       # root node_modules path that only existed under shamefully-hoist (#1151, #1099
@@ -422,6 +427,16 @@ jobs:
       # break its build silently, which is the state #552 found it in.
       - name: Build
         run: pnpm -C apps/nexus build
+
+      # The build proves the pages compile; it does not prove they rendered. This site has already
+      # shipped docs pages that prerendered to an empty shell, which produced a green build, a green
+      # fence check and a green route-list test (#1794). This reads the emitted HTML for the five
+      # evidence routes in both locales and fails when a page carries no prose.
+      #
+      # Verified against a real build before wiring: the ten pages render 423-10090 visible chars,
+      # and blanking one page's <main> drops it to 91 and fails the step by name.
+      - name: Verify prerendered docs pages carry their content
+        run: pnpm -C apps/nexus check:prerender-bodies
 
   job_app_tests:
     name: App suites (${{ matrix.app }})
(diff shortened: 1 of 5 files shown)


TalexDreamSoul:
`PR Quality` caught a real miss of mine: `antfu/no-top-level-await` on the module entry. Fixed in b71c9eb6e with `main().catch(...)`, matching the sibling check scripts.

Worth naming why I missed it: I ran the four nexus gates, the self-test, the mutation controls and the real-build check — but never ran eslint on the new file itself. Re-verified with **CI's own command** this time, `node scripts/run-eslint-changed.mjs` → exit 0, rather than an ad-hoc eslint invocation, since the nexus config resolves through `.nuxt/eslint.config.mjs` and a direct call does not reproduce it.

PR #1802, opened 2026-08-28
revert(nexus): remove the prerender body check — its premise is wrong

TalexDreamSoul:
Removes the check I added in #1799. It measures correctly and it caught nothing: **this app does not have the property it asserts.** #1800 is closed as not-a-defect.

## What the ten routes actually are

| route | what it is | should it server-render prose? |
| --- | --- | --- |
| `/{locale}/docs` | a real page | **yes** — and it does (1002 / 423 chars, passed) |
| `/{locale}/docs/dev` | a **redirect stub** — `robots: 'noindex, nofollow'`, server-side `navigateTo(target, { redirectCode: 302 })`, renders `DocsRedirectLoading` | **no** — ~170 chars is correct |
| `quickstart`, `components`, `guide/start` (× 2 locales) | the catch-all `[...slug].vue` | **no** — see below |

`apps/nexus/app/pages/docs/[...slug].vue:84`:

```ts
function shouldSplitDocsPageBody(path: string) {
  const normalized = normalizeDocsPagePath(path)
  return normalized.startsWith('/docs/') && normalized !== '/docs/index'
}
```

`currentDocsPageBodyMode` is `'0'` whenever that is true, so the server asks `/api/docs/page` for metadata only and the body arrives client-side. Deliberate, landed `c039f0c0e` on 2026-06-25, and pinned by `docs-page-performance.test.ts` under "docs page performance boundaries". The `/docs/dev` redirect is older still — `2260bdcf8`, 2026-05-25.

Eight of ten routes are supposed to render without prose. Only `/{locale}/docs` was ever a valid subject, and it passed.

## How I got it wrong

Both mistakes are worth naming, because neither was subtle:

1. **I verified against a build whose provenance I had not established.** The local `apps/nexus/dist` renders those pages at 1250–10090 chars, which is what made CI look like a regression. It contains **no `DocsRedirectLoading` markers at all**, so it was built from code older than 2026-05-25 despite an Aug 27 mtime. I used a baseline I could not date.
2. **I read "CI and production agree byte-for-byte" as proof of a defect.** 170/170, 156/156, 236/236 is equally consistent with both being *correct*, and that is the reading I should have tested first. Identical numbers prove the pipeline is deterministic, not that it is wrong.

The one habit that did work was the positive control on the production fetch — it caught that my initial "0 chars everywhere" reading was an unfollowed `308`. I applied that scepticism to the measurement and not to the conclusion.

## Why remove rather than rework

A route list that excluded everything this app deliberately renders body-less would have one entry, which already passes. There is no subject left.

And leaving an unwired script plus a CI self-test would imply a gate that does not exist — precisely the state #1776 found `check:doc-parity`, `check:icon-collections` and `build:analyze-worker` in.

## What stays

`docsPrerenderEvidenceRoutes` remains in `nexus-static-routes.mjs`. It has a real consumer in `createNexusPrerenderEvidence`, and moving it out of the `.ts` (so plain-node tooling can read it) is independently correct.

The four existing nexus gates are untouched: `check:api-routes`, `check:mdc-fences`, `check:icon-collections`, `check:doc-parity`.

## Unaffected

**#1794's actual finding stands.** `Protocol` is still not a required check, master was still red for nine days on an unpinned clippy, and #1756/#1758 are still draft as the manual stopgap. That issue's prerender bullet was mine and is withdrawn; the rest of it is not.

`actionlint -shellcheck` on the edited workflow: 0.


TalexDreamSoul:
**Correcting this PR's reasoning: it was wrong, and #1800 is reopened.**

There is a contract for exactly this — `.trellis/spec/frontend/nexus-docs-rendering-contract.md`, established **today**, headed "after every docs page was found shipping an empty shell in production":

> Anything that renders docs HTML fetches the body with it. Server render and prerender must request `/api/docs/page` with `body=1`… Splitting metadata from body is allowed only for client-side navigation.

So `shouldSplitDocsPageBody` reaching the server render **is** the defect. I read it as the design because it is two months old and has a test pinning it. Age and test coverage are not intent, and I never grepped `.trellis/spec` for the subsystem before concluding.

The fix — `DOCS_PAGE_RENDER_BODY_MODE = '1'` — exists in the local working copy but is **uncommitted and absent from origin**, which is why master measured as it did.

**What stands from this PR:** the `/docs/dev` half. Those are `noindex` redirect stubs and are legitimately body-less, so the check's route list was wrong to include them. Removing the check wholesale was not the right correction — narrowing the list was. That is now tracked on #1800 rather than re-landed here, since the fix has to be pushed first for a restored check to pass.

PR #1923, opened 2026-09-15
ci(nexus): watch Cloudflare Pages deployment Secrets on a schedule

TalexDreamSoul:
Closes the gap that let `NUXT_INTELLIGENCE_ENCRYPT_KEY` stay missing from Production.

## Why a schedule, not a PR gate

`check:deployment-secrets` was deliberately left out of CI: it exits 64 without Cloudflare credentials and reads live account state. So it ran nowhere.

The drift it catches is **time-based, not commit-based** — deleting a Secret in the Cloudflare dashboard produces no commit for a gate to run on. Production also deploys through the Pages Git integration rather than Actions, so there is no deploy job to hang it off. A weekly run is the only shape that works.

Same structure as `npm-token-watch.yml`, offset 30 minutes so the two never open issues in the same minute and read as one incident.

## Alert discipline

Only **exit 78** opens the incident issue. 64 (misconfigured) and 69 (Cloudflare unreachable) fail the run without it — an alert that cries "Production is broken" on a transient 5xx trains everyone to ignore it, which is how the original gate stayed trusted while checking nothing.

## Two defects found while verifying

**Doubled error code.** The preflight validates both environments before throwing so one run reports every problem, but the aggregator joined each failure's `.message` — already carrying `[CODE]` from the constructor — and passed it to a new error, which prefixed again. Every drift message read `[CODE] [CODE] ...`; with two failing environments, three times. That string is what the alert puts in front of an operator. No test pinned it, which is why it went unnoticed.

**Undocumented drift code.** Four codes produce exit 78; the issue body explained three. An operator hitting `PREVIEW_DEPLOY_BRANCH_IS_PRODUCTION` would have read an alert that never mentioned it. Now documented, and pinned by a test that derives both sides at runtime — nothing hardcoded, so adding or renaming a 78-code fails until the alert explains it.

## Verification

Workflow shell logic executed locally with a stubbed preflight:

| Case | Result |
|---|---|
| exit 0 / 78 / 64 / 69 | `ok`/`drift`/`misconfigured`/`unavailable` |
| multiline + 900-char output | one valid `GITHUB_OUTPUT` line, capped at 500 |
| each non-ok classification | step exits 1, distinct message |
| alert body | renders; `gh issue list` → create path exercised |

The preflight itself was driven through all seven error codes with a fake `fetchImpl`, including a replay of the original incident (key present in Preview, absent from Production → exit 78). Run from a clean `git archive` checkout with no `node_modules` to confirm the no-install claim.

Gates: `workflows:lint` (actionlint + shellcheck `-S info`, shellcheck 0.10.0 present), `check:workflow-injection`, `check:action-pins`, scoped ESLint — all pass. Preflight suite 23/23; the new guard verified discriminating by deleting the bullet (red, naming the offender) and restoring byte-identically (green, sha256 match).

## Summary by CodeRabbit

* **New Features**
  * Added a weekly and on-demand deployment secret check for Preview and Production environments.
  * Automatically creates or updates an alert when deployment secrets are missing or verification cannot be completed.
  * Reports distinct outcomes for verified configurations, detected drift, configuration issues, and unavailable service data.

* **Bug Fixes**
  * Improved aggregated deployment alerts by avoiding repeated error codes, making messages easier to read.

* **Tests**
  * Added validation to keep deployment error classifications aligned with documented alerts.


TalexDreamSoul:
## Blocker before this can alert: the credentials are not in Actions scope

Checked after opening the PR — `CLOUDFLARE_ACCOUNT_ID` and `CLOUDFLARE_API_TOKEN` are **not available to Actions anywhere** in this repo:

| Scope | Result |
|---|---|
| Repo secrets | 16 present, neither Cloudflare name among them |
| Org secrets | none visible to this repo |
| Environment `production` | empty |
| Environment `github-pages` | `ACCESS_TOKEN`, `GIT_EMAIL` only |

So the first scheduled run takes the **exit 64 / `misconfigured`** path: the preflight never reaches Cloudflare, the run fails, and no drift issue opens.

That is the designed behaviour, not a bug in this PR — the workflow deliberately separates "this watcher cannot run" from "Production has drifted", and says so:

> A watcher that cannot run is a broken watcher, and silently green is how the original gate stayed trusted while checking nothing.

It fails loudly instead of passing green, which is the right call. But until the two secrets are added it is a **weekly red cron that proves nothing**, and a red cron nobody reads is exactly the failure mode this was built to end.

### To make it live

Add both secrets, then confirm with a manual run before trusting the schedule:

```bash
gh secret set CLOUDFLARE_ACCOUNT_ID
gh secret set CLOUDFLARE_API_TOKEN
gh workflow run "Nexus Deployment Secret Watch"
```

The token needs read access to Pages project metadata. The preflight reads **names and binding types only** — it never reads, logs, or writes a Secret value, so a minimal read-scoped token is sufficient.

A green manual run is also the only end-to-end proof available here: everything in the PR description was verified with a stubbed `fetchImpl`, which cannot exercise the real Cloudflare response shape.

Merging is still correct — leaving the check running nowhere is strictly worse than a red run that names its own missing input.

TalexDreamSoul:
Follow-up: dispatching this workflow immediately after merge exposed a real defect, fixed in #1924.

The first run failed with an **empty** `CLASSIFICATION` and reported *"Cloudflare metadata was unavailable"* — the one diagnosis this contract exists to keep separate from a missing credential.

Cause: Actions runs the step as `bash -e {0}`, and a non-zero preflight is the normal case here, so `-e` killed the step at the `output="$(node ...)"` assignment before the `case` that classifies it. No outputs were written, and the later steps fell through to the `*)` catch-all. `set -uo pipefail` cannot cancel the launcher's `-e`; it needed `set +e`.

My pre-merge harness ran the step body under `bash -c`, which carries no `-e`, so all four cases passed locally.

After #1924, [run 34951056069](https://github.com/talex-touch/tuff/actions/runs/34951056069) reports correctly:

```
CLASSIFICATION: misconfigured
This workflow is missing CLOUDFLARE_ACCOUNT_ID / CLOUDFLARE_API_TOKEN.
The preflight never ran, so Production is unverified — not proven healthy.
```

It still fails the run and still opens **no** incident issue, which is the intended behaviour for a watcher that cannot run.

The credential blocker above stands: until the two secrets are added, this is a weekly red run that proves nothing.

